PR #161, opened 2014-07-09
added Merge of 2 polymorphic functions

mandubian:
- added operator `|+|` as an alias to `merge`
- Due to limitations of compiler, now you cannot write:
```scala
// DOESN'T WORK : compiles but forgets implicit cases from poly1 |+| poly2
val m = poly1 |+| poly2 |+| poly3

// WORKS
val m1 = poly1 |+| poly2
val m = m1 |+| poly3
```

### DIFF
--- a/core/src/main/scala/shapeless/poly.scala
+++ b/core/src/main/scala/shapeless/poly.scala
@@ -81,6 +81,27 @@ object PolyDefns extends Cases {
   }
 
   /**
+   * Represents the merge of two polymorphic function values.
+   *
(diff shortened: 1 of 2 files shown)


milessabin:

mandubian:
Is it still useful in some way?
It's quite old to me so you can close it if you don't remember as much as me ;)

milessabin:
It would still be nice to have more general compose/merge combinators. But I guess this is pretty stale now ... I'll close it.

PR #187, opened 2014-07-19
Added Exact extractor for Sized that requires the right number of patterns

travisbrown:
This is a reworked version of #93 that uses the new `ToHList` type class from #114.

While this extractor behaves correctly (`val Sized.Exact(a, b) = Sized('foo)` doesn't compile), we can't confirm that with `illTyped` in 2.11 at the moment because of [this bug](https://issues.scala-lang.org/browse/SI-8719) (see [this Stack Overflow question](http://stackoverflow.com/q/24231855/334519) and Eugene Burmako's answer for discussion). I'll create an issue to remind us to add these tests when [Eugene's fix](https://github.com/scala/scala/pull/3876) lands.


milessabin:

travisbrown:
@milessabin Sorry, somehow I just noticed this message. I'm happy to close this, although I'd still like to have a version that works as you describe [here](https://github.com/milessabin/shapeless/pull/187#issuecomment-49523437) (when I took a stab at it last summer I didn't get very far).

Issue #301, opened 2015-01-02
LabelledGeneric and scala 2.10


alexarchambault:
`LabelledGeneric` seems not to work anymore in a scala 2.10 project (when used outside shapeless itself). For example, the following doesn't compile:
```scala
case class Foo0()
 
case class Foo1(i: Int)
 
object Lgen {
import shapeless._
 
LabelledGeneric[Foo0]
LabelledGeneric[Foo1]
} 
```

`sbt compile` gives:
```scala
> compile
[info] Compiling 1 Scala source to /home/alexandre/projects/lgen-2.10.x/target/scala-2.10/classes...
[error] /home/alexandre/projects/lgen-2.10.x/Lgen.scala:8: could not find implicit value for parameter lgen: shapeless.LabelledGeneric[Foo0]
[error] LabelledGeneric[Foo0]
[error] ^
[error] /home/alexandre/projects/lgen-2.10.x/Lgen.scala:9: could not find implicit value for parameter lgen: shapeless.LabelledGeneric[Foo1]
[error] LabelledGeneric[Foo1]
[error] ^
[error] two errors found
[error] (compile:compile) Compilation failed
[error] Total time: 2 s, completed 2 janv. 2015 01:43:08
> 
```

(All the files to reproduce this error are available here: https://gist.github.com/alexarchambault/1fd71d804f0661ad51c1).

The type parameter `R` (the one filled by the macro) may be the problem, as the following compiles:
```scala
case class Foo0()
 
case class Foo1(i: Int)
 
object Lgen2 {
import shapeless._
import language.experimental.macros
 
// Works with this one, but hard to do anything with it then...
implicit def lgen[T]: shapeless.LabelledGeneric[T] = macro shapeless.GenericMacros.materializeLabelled[T, Nothing]
 
// Does *not* work with this one
// implicit def lgen[T, R]: shapeless.LabelledGeneric.Aux[T, R] = macro shapeless.GenericMacros.materializeLabelled[T, R]
 
LabelledGeneric[Foo0]
LabelledGeneric[Foo1]
} 
```

alexarchambault:
So adding
```scala
addCompilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
```
to `build.sbt` fixes it...

Maybe we could add a warning about that in the README or the wiki...

milessabin:
[...]

PR #302, opened 2015-01-02
README line about macro paradise plugin when using scala 2.10.x

alexarchambault:
Fixes https://github.com/milessabin/shapeless/issues/301


milessabin:
Wonderful ... thanks :-)

PR #303, opened 2015-01-03
Allowed Record to be called with no arguments

alexarchambault:
Allows the following to compile:
```scala
val r = Record()
```

### DIFF
--- a/core/src/main/scala/shapeless/records.scala
+++ b/core/src/main/scala/shapeless/records.scala
@@ -56,6 +56,7 @@ object record {
    * }}}
    */
   object Record extends Dynamic {
+    def applyDynamic(method: String)(rec: Any*): HList = macro RecordMacros.mkRecordEmptyImpl
     def applyDynamicNamed(method: String)(rec: Any*): HList = macro RecordMacros.mkRecordNamedImpl
 
     def selectDynamic(tpeSelector: String): Any = macro LabelledMacros.recordTypeImpl
@@ -97,6 +98,13 @@ class RecordMacros(val c: whitebox.Context) {
   val fieldTypeTpe = typeOf[FieldType[_, _]].typeConstructor
   val SymTpe = typeOf[scala.Symbol]
   val atatTpe = typeOf[tag.@@[_,_]].typeConstructor
+
+  def mkRecordEmptyImpl(method: Tree)(rec: Tree*): Tree = {
+    if(rec.nonEmpty)
+      c.abort(c.enclosingPosition, "this method must be called with named arguments")
+    
+    q"_root_.shapeless.HNil"
+  }
 
   def mkRecordNamedImpl(method: Tree)(rec: Tree*): Tree = {
     val q"${methodString: String}" = method
--- a/core/src/test/scala/shapeless/records.scala
+++ b/core/src/test/scala/shapeless/records.scala
@@ -622,9 +622,25 @@ class RecordTests {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


milessabin:
This is basically another name for `HNil`. Could you give me a rationale for adding it?

alexarchambault:
"Makes the intent clearer" ?

core/src/main/scala/shapeless/records.scala
```diff
@@ -98,6 +99,13 @@ class RecordMacros(val c: whitebox.Context) {
   val SymTpe = typeOf[scala.Symbol]
   val atatTpe = typeOf[tag.@@[_,_]].typeConstructor
 
+  def mkRecordEmptyImpl(method: Tree)(rec: Tree*): Tree = {
+    if(rec.nonEmpty)
+      c.abort(c.enclosingPosition, "this method must be called with named arguments")
```
milessabin: s/named/no/ ?

core/src/main/scala/shapeless/records.scala
```diff
@@ -98,6 +99,13 @@ class RecordMacros(val c: whitebox.Context) {
...
+    
+    mkRecordNamedImpl(method)(rec: _*)
```
milessabin: q"_root_.shapeless.HNil"?

core/src/main/scala/shapeless/records.scala
```diff
@@ -56,6 +56,7 @@ object record {
    * }}}
    */
   object Record extends Dynamic {
+    def applyDynamic(method: String)(rec: Any*): HList = macro RecordMacros.mkRecordEmptyImpl
```
milessabin: Presumably a vanilla `apply()` would preempt the `applyDynamicNamed`?

milessabin:
Fair enough. But I'm slightly nervous that this might give the impression that shapeless records are more first-class than they are.

Happy to merge, modulo the comments I've just made above.

core/src/main/scala/shapeless/records.scala
alexarchambault: It does, indeed. Errors like:
```scala
[error] /home/alexandre/projects/shapeless/core/src/test/scala/shapeless/product.scala:132: too many arguments for method apply: ()shapeless.HNil.type in object Record
[error]       val expectedFooL = Record(i = 1, s = "b")
[error]                                ^
```

core/src/main/scala/shapeless/records.scala
alexarchambault: This error message appears when `Record` is called like
```scala
Record(2, "a")
```

So this error message ("must be called with *named* arguments") looked like the right one to me.

core/src/main/scala/shapeless/records.scala
alexarchambault: Better this way, yes, I'll update the PR.

alexarchambault:
Updated the PR. What do you mean by not giving the impression that records are more first-class than they are? Aren't they a useful ~collection (which shapeless makes all the more easy to use)?

milessabin:
Really it's just a vague worry that people might forget that they are just HLists. No big deal.

core/src/main/scala/shapeless/records.scala
milessabin: Makes sense.

milessabin:
In any case, LGTM ... many thanks ... merging :-)

PR #304, opened 2015-01-03
Generic ops, Generic withLabels method

alexarchambault:
This PR adds a method to `Generic` (via a `GenericOps` intermediate class).

This method allows to somehow override the labels set by default by `LabelledGeneric`, liked wished in https://github.com/milessabin/shapeless/pull/289, with
```scala
// case class Person(lastName: String, addressLine: String, yearsOld: Int)

implicit val personLabelsOverride = 
  Generic[Person].withLabels[HList.`'last_name, 'address_line, 'years_old`.T]
```

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -34,6 +34,10 @@ object Generic {
   def apply[T](implicit gen: Generic[T]): Aux[T, gen.Repr] = gen
 
   implicit def materialize[T, R]: Aux[T, R] = macro GenericMacros.materialize[T, R]
+
+  import syntax.GenericOps
+
+  implicit def genOps[T, R](gen: Generic.Aux[T, R]): GenericOps[T, R] = new GenericOps(gen)
 }
 
 trait LabelledGeneric[T] extends Generic[T]
--- a/core/src/main/scala/shapeless/ops/generic.scala
+++ b/core/src/main/scala/shapeless/ops/generic.scala
@@ -1,0 +1,56 @@
+package shapeless
+package ops
+
+import labelled._
+
+object generic {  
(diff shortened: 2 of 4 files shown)


alexarchambault:
The same method *could* be added to `LabelledGeneric` itself (not done here), and used like
```
implicit val personLabelsOverride =
  LabelledGeneric[Person].withLabels[HList.`'last_name, 'address_line, 'years_old`.T]
```
but this way of using it would be risky: this line requires an implicit `LabelledGeneric[Person]`, and redefines one at the same time. The implicit wouldn't find itself here, but it would if we precise its type like in
```
type R = Record.`'last_name -> String, 'address_line -> String, 'years_old -> Int`.T

implicit val personLabelsOverride: LabelledGeneric.Aux[Person, R] =
  LabelledGeneric[Person].withLabels[HList.`'last_name, 'address_line, 'years_old`.T]
```

milessabin:
I'm almost on board with this, but to be honest I'd rather go in a different direction. I'm now coming to the opinion that `LabelledGeneric[T]` _as_a_primitive_ was a mistake. I should really have gone with just `Generic[T]` (as it is now), and a `DefaultLabelling[T]` (which yields an `HList` of the default labels). `LabelledGeneric[T]` would then be trivially definable in terms of those two (without additional macros) and would provide a template for providing alternative labellings.

I'd be happy to accept a pull request along those lines, but I think it's too close to the 2.1.0 release now to make a change of that magnitude.

alexarchambault:
`DefaultLabelling` sounds like a better idea indeed. I'll wait for that.

milessabin:
It turned out that the easiest way fixing/investigating #309 and #310 was by fixing this. Now done in 11bbe2d9f045be91b806e8fea93ccc3cbf161449.

PR #305, opened 2015-01-03
prettyString method for records, unions, products

alexarchambault:
This PR proposes to add a `prettyString` method to records, unions, and products too, so that we can write
```scala
scala> Record(i=2, s="a").prettyString
res0: String = Record(i=2, s=a)
```

```scala
scala> type U = Union.`'i -> Int, 's -> String, 'b -> Boolean`.T
defined type alias U

scala> Union[U](s="a").prettyString
res0: String = Union(s=a)
```
or
```scala
scala> case class Foo(i: Int, s: String)
defined class Foo

scala> Foo(i=2, s="b").prettyString
res0: String = Foo(i=2, s=b)
```

### DIFF
--- a/core/src/main/scala/shapeless/ops/prettystring.scala
+++ b/core/src/main/scala/shapeless/ops/prettystring.scala
@@ -1,0 +1,150 @@
+package shapeless
+package ops
+
+import shapeless.labelled.FieldType
+
+object prettystring {  
+  trait PrettyString[T] {
+    def suggestedName: String
+
+    /**
+     * Prints the elements of `t` in `b`
+     * 
+     * @param sepBefore: if Left, a separator between elements, if Right, a string to prepend before each element  
+     */
+    def addElements(t: T, b: StringBuilder, sepBefore: Either[String, String]): Unit
+
+    def mkString(t: T, prefix: String, sep: String, suffix: String): String = {
+      val b = new StringBuilder      
+      b ++= prefix; addElements(t, b, Left(sep)); b ++= suffix      
+      b.result()
+    }
+    
+    def apply(t: T): String =
+      mkString(t, s"$suggestedName(", ", ", ")")
+  }
+  
+  trait LowPriorityPrettyString {
+    implicit def hconsNonHeadRecursivePrettyString[K <: Symbol, H, T <: HList](implicit
+      key: Witness.Aux[K],
+      tailPrettyPrint: Lazy[PrettyString[T]]
+    ): PrettyString[FieldType[K, H] :: T] =
+      new PrettyString[FieldType[K, H] :: T] {
+        def suggestedName = tailPrettyPrint.value.suggestedName
+        def addElements(r: FieldType[K, H] :: T, b: StringBuilder, sepBefore: Either[String, String]) = {
+          sepBefore match {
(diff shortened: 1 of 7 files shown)


alexarchambault:
Note that to avoid a name clash between products and records, like in
```scala
scala> import record._, syntax.std.product._
import record._
import syntax.std.product._

scala> Record(i=2, s="a").prettyString
<console>:17: error: type mismatch;
 found   : shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("i")],Int],shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("s")],String],shapeless.HNil]]
    (which expands to)  shapeless.::[Int with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("i")],Int],shapeless.::[String with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("s")],String],shapeless.HNil]]
 required: ?{def prettyString: ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method recordOps in object record of type [L <: shapeless.HList](l: L)shapeless.syntax.RecordOps[L]
 and method productOps in object product of type [P <: Product](p: P)shapeless.syntax.std.ProductOps[P]
 are possible conversion functions from shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("i")],Int],shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("s")],String],shapeless.HNil]] to ?{def prettyString: ?}
              Record(i=2, s="a").prettyString
                    ^
<console>:17: error: value prettyString is not a member of shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("i")],Int],shapeless.::[shapeless.labelled.FieldType[shapeless.tag.@@[Symbol,String("s")],String],shapeless.HNil]]
              Record(i=2, s="a").prettyString
                                 ^
```

it would be better to restrain `ProductOps` conversions to types having a `HasProductGeneric`, like in https://github.com/alexarchambault/shapeless/commit/fc5a5706a5a5988b9f98d1997431d423ae76c1ce

alexarchambault:
Note too that as the `PrettyString` type class is common to records, unions, and products, I put it in a specific namespace `ops.prettyprint`, for lack of a better one...

stacycurl:
What happens if a field is also a record, would it recurse ? I also think
it would be neat if the string were 'homoiconic'.

alexarchambault:
It recurses on records, unions, and products, yes.

About homoiconicity, you mean we should put some `"` around strings? I agree it would be nice to have. (I didn't for `prettyString` not to be too surprising, but I could change that.)

milessabin:
I think that this (and the `toString` methods in shapeless generally) should be provided via a `Show` typeclass ... we'll be able to do that once typelevel/cats is up and running (ie. very soon, but not soon enough for shapeless 2.1.0 I'm afraid).

alexarchambault:
Having a `Show` type class sounds like a good idea. Couldn't it also be used without modifying scalac, by giving the string conversion method another name than `toString` as a workaround? (All of that can wait after the 2.1 release...)

PR #308, opened 2015-01-05
Add CSV serialization/deserialization example

melrief:


examples/src/main/scala/shapeless/examples/csv.scala
```diff
@@ -0,0 +1,103 @@
...
+import scala.util.{Try,Success,Failure}
+
+
+/*
+ * This example shows how to createa a serializer/deserializer from CSV to
+ * products.
+ * */
+
+// The class to serialize or deserialize
+case class Person(name: String, surname: String, age: Int)
```
milessabin: Ahh ... hang on ... this is unused ... why not demo (de-)serialization of a `List[Person]`? That would be more compelling.

melrief: You are right, sorry but I submitted the code yesterday late. I changed the example, now  we deserialize into a list of `Person` by using `Generic`. What do you think?

In future I will try to make possible to create a `CSVConverter[List[Person]]` instead of `CSVConverter[List[personGen.Repr]]` + `Generic[Person]`...it shouldn't be difficult.

milessabin: I think it would be better to get `CSVConverter[List[Person]]` working before merging if you've got the time to work on that ... it shouldn't be too difficult.

melrief: Agreed but I'm not sure how to do it. The idea was to have a
`implicit def deriveClass[A](implicit gen: Generic[A], conv: CSVConverter[gen.Repr]) : CSVConverter[A] = //...`
but this doesn't work of course because of the dependent method type of the second implicit. Can I ask you what is the best way to implement this? I'm new to generic programming.

melrief: Better now? Thanks to Travis Brown I made it (I think).

milessabin:

PR #312, opened 2015-01-15
More quasiquote hygiene

alexarchambault:
Should fix https://groups.google.com/forum/#!topic/typelevel/LmgntI3LU8c, I was running into the same problem with `Lazy` uses involving labelling, like in argonaut-shapeless.


milessabin:
Many thanks ... merging :-)

PR #315, opened 2015-01-26
Type-level representation of the whole Integers set (aka rational integers, as opposed to natural integers) 

Baccata:
I wanted to extend the shapeless typelevel representation of integers to allow operations with negative integers. This started as an exercise but I figured some people might want to have it and I might as well submit a pull request ... hence the following additions / modifications  : 

- When relevant, renamed "nat" files to "numbers", as it's more generic than Nat and more understandable than "RInt" (for Rational Integers). 
- I tried to be as generic as I could be, and might have got a bit lost in the creation of my traits and classes. Any critic will be more than welcome. Anyhow, my implementation of negative numbers relies on the "Minus" class, which is parameterized by a SPos (strictly positive number). 
- Obviously added macros relative to rational integers, as well as boiler plate up to (or more exactly down to) minus 22. My choice for the type names for negative numbers was _mX, to reflect the orignal _1, _2 ... _X.   
- Duplicated/recreated the nat operations to ops/numbers.scala and kept the orignal file. Obivously I didn't want to change nat operations to keep users safe from my tweaks (5 - 6 = 0 in the Natural Integers world), but since my implementations rely on more implicit evidence than the original implementation, proposing a common implementation in some cases might impact the users' compiling time, and I'm not sure how you guys would like to tackle this. 
- I implemented (and added tests for) the operations add, diff, prod, div, less than, less than or equal, as well as some others that came in handy (divergent implicit expansion errors are my new nemesis). I still have to implement "power", and perhaps "mod". 

Anyway, kudos to the Shapeless guys for this fantastic library. I had (and still have) a hard time to get my head around it, but it's everything I've ever dream of for a strongly typed language. 

English not being my mothertongue, some of the wording might not be accurate enough . I will fix it as soon as I get feedback.


milessabin:
Apologies for the slow response on this. Pretty much everything is on hold until shapeless 2.1.0 final is released.

Baccata:
Hello, 

After seeing some other implementations, I've been re-implementing typelevel integers using [type projections](https://gist.github.com/Baccata/37dc5eaa087cbf20556c) instead of implicits in a pet project, and the benefits, readabitlity-wise and peformance-wise, seem pretty good. 

Would it be interesting to rewrite Shapeless's implementation using type projections ? For instance, it was applied to shapeless, the "apply" method in SizedOps could look like : 
    
    def apply[N <: Nat](implicit lt : (N < L) =:= True, ev: ToInt[N]): A0 = s.unsized.drop(toInt[N]).head 

The operation typeclasses in ops.nat would be rewritten in a trivial manner, and still be used to factorize type computation when relevant, but the instantiation of the implicits would cost a lot less, since their definitions would not have any implicit parameters.    
  
Is there a benefit of implicits I'm not seeing, in this use case?

milessabin:
In my experience type level computations using implicits are much more robust (in the sense of being more easily generalized, composed and being less prone to compiler bugs) than ones that use type projection. But I'm open to the possibility that at least for numerical types this might be an avenue worth exploring ... if you'd like to see what can be done in that direction then I would be very interested to see what you can come up with.

As far as this PR is concerned, I think it's very well put together but I don't really want to merge it as is. One of the goals for shapeless 3.0 is to split the library into smaller components with minimal dependencies. For that purpose `Nat` really is the fundamental type (the key thing here is the structural similarity between `Nat` and list-like things) so I would want to be able package up the `Nat` types independently from anything else.

If you're able to rejig this PR so that `Nat` is unaffected by the introduction of the new numeric types I think I would be much more comfortable with merging it.

milessabin:

Baccata:
You should definitely close it. I had a blast working on it and learnt a lot but there is no point merging it as it doesn't bring anything worth (to shapeless). 

About the integer operations based on type projections, I did try to implement it but it was messing up a couple tests. You were right, though less readable, implicits are much more flexible. 

Sorry for not notifying you, I was thinking that I would come back to it, but now that you're asking I'd rather see the PR closed.

milessabin:
No worries, and thanks again for taking a crack at this ... it's always interesting to see how different approaches pan out :-)

PR #316, opened 2015-01-30
Simplify sexp example

jedesah:
The sexp example is not using the `LabelledTypeClassCompanion` object and thusly is repeating a lot of boilerplate.

I was curious if there was a good reason for that so I set out to convert it to using the alternative method. It turns out this method would seem to work as all tests still pass.

Maybe there is a runtime performance hit or some other reason I am not thinking of in which case I would be really grateful if someone could explain what it is.

If not, then I suggest merging the PR in order to avoid misleading newcomers (like myself to a certain extent) into thinking typeclass derivation is more complex than it needs to be (for a user).

### DIFF
--- a/examples/src/main/scala/shapeless/examples/sexp.scala
+++ b/examples/src/main/scala/shapeless/examples/sexp.scala
@@ -133,75 +133,58 @@ object SexpUserConvert {
   }
 }
 
-object SexpConvert {
-  def apply[T](implicit st: Lazy[SexpConvert[T]]): SexpConvert[T] = st.value
+object SexpConvert extends LabelledTypeClassCompanion[SexpConvert] {
 
-  implicit def deriveHNil: SexpConvert[HNil] =
-    new SexpConvert[HNil] {
+  val typeClass = new LabelledTypeClass[SexpConvert] {
+    def emptyProduct = new SexpConvert[HNil] {
       def deser(s: Sexp) = if (s == SexpNil) Some(HNil) else None
       def ser(n: HNil) = SexpNil
     }
 
-  implicit def deriveHCons[K <: Symbol, V, T <: HList]
-    (implicit
-      key: Witness.Aux[K],
-      scv: Lazy[SexpConvert[V]],
-      sct: Lazy[SexpConvert[T]]
-    ): SexpConvert[FieldType[K, V] :: T] =
-      new SexpConvert[FieldType[K, V] :: T] {
-        def deser(s: Sexp): Option[FieldType[K, V] :: T] = s match {
-          case SexpProp((label, car), cdr) if label == key.value.name =>
-            for {
-              front <- scv.value.deser(car)
-              back <- sct.value.deser(cdr)
-            } yield field[K](front) :: back
-
-          case _ =>
-            println("PRODUCT MISS = " + s)
-            None
-        }
-
-        def ser(ft: FieldType[K, V] :: T): Sexp = {
-          val car = SexpProp(key.value.name, scv.value.ser(ft.head))
-          sct.value.ser(ft.tail) match {
-            case SexpNil => car
-            case cdr => SexpCons(car, cdr)
-          }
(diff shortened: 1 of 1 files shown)


milessabin:
@fommil put this example together before I had rebuilt `TypeClass` and friends on top of `Lazy` and `Generic`, so the option wasn't available then. I think that now it serves as a good example of how to use the lower-level infrastructure, as compared with the monoid derivation example which uses `TypeClass`.

Maybe it would be an idea to have both variants (maybe in the same source file) to illustrate the differences, probably with an addition to the current version which does something which goes beyond what's possible with `TypeClass` (ie. pulling in an instances for some other type class during the derivation process).

fommil:
it would be great to see this as a separate example. I'm planning on using `TypeClass` in anger within ENSIME and I'd like to see how the new approach stacks up against it.

I'd also be really interested in seeing spray-json using shapeless for sealed families... I have an extremely good performance stress test at my work that I could run it through :wink: (already been sending PRs to core `spray-json` with significant performance fixes)

fommil:
btw, an example using a simplified version of [JsValue](https://github.com/spray/spray-json/blob/master/src/main/scala/spray/json/JsValue.scala) would be awesome. In terms of "user land" serialisers/deserialisers, just int/String/List[T] would be more than enough.... the extension to full spray-json would then be fairly trivial.

jedesah:
My train of thought was that the special treatment of `Option` in the csv example was a good example of how and why to use the low-level infrastructure and so I figured since this example did not need the low-level infrastructure that it could show the more concise way. However, I was not aware that there was an example deriving instances for `Monoid` that was using the more concise approach.

I guess it can't hurt to have both for the sexp example.

@fommil I just finished upgrading to Shapeless 2.1 for this internal json lib we have so I feel pretty comfortable writing an example for spray-json if you think it would be useful.

fommil:
Oh yes please! I have so much stuff to do in ENSIME right now that I'll never get round to it :smile:

milessabin:
I think the consensus we reached was to have this as a separate example which could be easily put side by side with @fommil's original ... would be great if you have time to put that together.

jedesah:
Yes. I'll try to find the time this weekend or next week to do this! Thx
for poking me (seriously)!

fommil:
@jedesah are you doing the SExp or Spray JSON using the new shapeless API? I'll be very happy to see it because I've been meaning to bring this all into ENSIME when we move to 2.1.

fommil:
@jedesah where are you on the spray-json version of this? I will do it as part of https://github.com/ensime/ensime-server/issues/844 unless you beat me to it, so please do keep me in the loop if you have something :smile:

jedesah:
@fommil I haven't been able to find the time to do this yet, so feel free to go ahead and do it. I certainly won't mind!

@milessabin btw, the spray-json example will need to use the lower-level API in order to handle `Option`s in a nifty way (much like the CSV example) so it might make sense to simplify the SEXP example since it does not actually need the extra flexibility (at least not in it's current form). I feel like this PR should either be merged as is or I (or someone else) can submit a new PR that modifies this example to actually require the low-level API.

fommil:
I'm looking at doing the Spray JSON marshaller as port of ENSIME. If it works out well, I might release it as a separate project.

fommil:
FYI this is the first cut of my spray-json coproduct marshaller using shapeless 2.2: https://github.com/fommil/ensime-server/blob/shapeless/jerk/src/main/scala/org/ensime/jerk/FamilyFormats.scala
unfortunately, the tests don't compile https://github.com/fommil/ensime-server/blob/shapeless/jerk/src/test/scala/org/ensime/jerk/FamilyFormatsSpec.scala#L24 complaining about "forward reference extends over definition of value tokenTreeFormat"

I have no idea what is going on. this is the referenced ExampleAst https://github.com/fommil/ensime-server/blob/shapeless/sexpress/src/test/scala/org/ensime/sexp/formats/ExampleAst.scala

If I drop the "implicit" from that line, I get "unreachable code".

milessabin:

jedesah:
Yes, I think it is safe to close for now at least.

PR #319, opened 2015-02-05
Add nifty handling of the Option type to csv example

jedesah:
I find this pretty cool that it's possible to do this! And it's a good example of a case where it is necessary to avoid using `LabelledProductTypeClassCompanion`. Should we add it to the example?

### DIFF
--- a/examples/src/main/scala/shapeless/examples/csv.scala
+++ b/examples/src/main/scala/shapeless/examples/csv.scala
@@ -34,7 +34,7 @@ object CSVExample extends App {
 
   import CSVConverter._
 
-  val input = """John,Carmack,23
+  val input = """John,Carmack,23,0
 Brian,Fargo,35
 Markus,Persson,32"""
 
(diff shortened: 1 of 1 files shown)


milessabin:
Looks good ... maybe add an example exercising the new functionality?

jedesah:
It is being exercised. The PR introduces a new optional field on the case class and yet the example still compiles. It would not compile if it weren't for the new `deriveHConsOption`.

Did you have something else in mind?

jedesah:
There is something that bothers me here that I had not thought of originally. This will cause an ambiguous implicit compile error if someone provides a `CSVConvert` for `Option[T]` which for the case of csv is probably not all that useful.  The same will happen if a `deriveCNil` and `deriveCCons` are defined because Option is an ADT.

It's probably possible to play with implicit priorities to solve this.

In it's current form though it works, I just want to avoid people being surprised and frustrated if they try to extend the example.

What do you think?

milessabin:
It's just that the new field is missing in all the input rows. It'd be nice to have a present/absent mix.

milessabin:
Playing with implicit priorities is indeed the name of the game here. I think in most scenarios this will work out reasonably smoothly.

jedesah:
@milessabin Ah okay, good idea. Done!

milessabin:
Oops ... sorry, I thought I had already merged this. Thanks! :-)

PR #321, opened 2015-02-08
Refactored SMC + added SMCTests

stacycurl:


core/src/test/scala/shapeless/smc.scala
```diff
@@ -0,0 +1,166 @@
...
+
+    val ppLA = (p, p).leftAssoc
+    assertTypedEquals[(((((((String, Int), Boolean), Unit), String), Int), Boolean), Unit)](
+      ((((((("foo", 23), true), ()), "foo"), 23), true), ()),
+      ppLA
+    )
+  }
+
+  trait PolyFoldLeft extends Poly2 {
+    // Does it make sense to unify 'FoldLeftPBuilder' & 'FoldLeftCBuilder' (by using SMC), or would that be cyclic ?
```
milessabin: I think it would be worth a try.

core/src/main/scala/shapeless/smc.scala
```diff
@@ -263,67 +381,12 @@ class SMCOps[P, S <: SMC](p: P, val smc: S) {
   def reverse(implicit reverse: S#Reverse[P]): reverse.Out = reverse(p)
   def map(f: Poly)(implicit mapper: S#Mapper[f.type, P]): mapper.Out = mapper(p)
   def foldLeft[Z](z: Z)(f: Poly2)(implicit folder: S#LeftFolder[P, Z, f.type]): folder.Out = folder(p, z)
+
+  def rightAssoc(implicit rightAssoc: S#RightAssoc[P]): rightAssoc.Out = rightAssoc(p)
+  def leftAssoc(implicit leftAssoc: S#LeftAssoc[P]): leftAssoc.Out = leftAssoc(p)
 }
```
milessabin: A lot of this is to avoid calls of `swap`, yes? I guess that makes sense eventually at least from a performance PoV, but at this point it might be obscuring the essential structure.

milessabin:
A couple of comments above, but they won't stop me merging this ... good stuff :-)

PR #332, opened 2015-02-22
Include modules in top-level REPL classpath.

### LINKED ISSUE #203: SBT project has unintuitive structure ... top-level console should launch REPL with sensible classpath
I think it'd be good to have some discussion about what we want before proceeding with this.

Currently the project is structured with three subprojects: shapeless-core, shapeless-examples and shapeless-scratch. The top-level project aggregates shapeless-core and shapeless-examples, but typing console at the top-level doesn't included core (or examples) on the classpath of the resulting REPL. At one point the top-level publish used to publish shapeless-core (this was as intended) ... now it attempts to publish everything and succeeds only for shapeless-core (good) but fails noisily for itself, shapeless-examples and shapeless-scratch (not good). The shapeless-core subproject is rooted in the core directory, shapeless-examples in the examples directory and shapeless-scratch in the scratch.

TBH I think this is a bit of a mess. I think it felt motivated at the time, but I no longer remember why and right now it feels confusing, clunky and awkward.

What I think I'd like now is an arrangement which allowed,

* The project prefixes to be dropped (ie. the subprojects are core, examples, scratch to match their directory name.
* compile, test etc. at the top-level compiles, tests, etc. all subprojects.
* console at the top-level drops into a REPL with core, examples and scratch on the classpath.
* publish at the top-level publishes only core (or, alternatively, is an error) and any other future subprojects which are supposed to be published (currently core is the only one).

Anything else?

dwijnand:
Also drops "shapeless-" from the modules, locally (jars retain the prefix).

Fixes #203. Surprisingly easy (solves all 4 points).

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -34,11 +34,10 @@ import sbtrelease.Utilities._
 
 object ShapelessBuild extends Build {
 
-  lazy val shapeless = Project(
-    id = "shapeless",
-    base = file("."),
-    aggregate = Seq(shapelessCore, shapelessExamples),
-    settings = commonSettings ++ Seq(
+  lazy val shapeless = (project in file(".")
+    aggregate (core, examples)
+    settings (commonSettings: _*)
+    settings (
       moduleName := "shapeless-root",
 
       (unmanagedSourceDirectories in Compile) := Nil,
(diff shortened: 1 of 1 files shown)


milessabin:
Oh fantastic! Thanks :-)

I want to kick the tires on this locally before merging ... I should have a chance tomorrow ...

dwijnand:
Yep that's what I'd recommend, just in case I missed anything.

milessabin:
Absolutely perfect ... many thanks! :-)

It also fixes a problem I'd been seeing where `~` prefixed SBT commands against the scratch project weren't triggered by source changes ... it probably had something to do with the top-level aggregation.

Issue #313, opened 2015-01-21
toString on Typeable is not helpful


rorygraves:
In Ensime we are using shapeless to do typesafe calls to an actor:

```scala
  def callRPC[R](target: ActorRef, request: RPCRequest, maxWait: FiniteDuration = defaultMaxWait)(implicit typ: Typeable[R]): R = {
    val future = target.ask(request)(maxWait)
    val result = Await.result(future, maxWait)
    result match {
      case e: RPCError =>
        throw e
      case r =>
        val castRes = r.cast[R]
        if (castRes.isDefined)
          castRes.get
        else {
          val msg = "Error, incorrect type found in RPC call, expected: " + typ + " got " + r.getClass
          val ex = RPCError(ProtocolConst.ErrExceptionInRPC, msg)
          log.error(msg, ex)
          throw ex
        }
    }
  }
```

I'm sure there are many things that you dislike about this code, but my problem is I cannot generate a good error message: - on failure I get:

```
org.ensime.core.RPCError: 201: Error, incorrect type found in RPC call, expected: shapeless.LowPriorityTypeable$$anon$1@7383291a got class java.lang.Boolean```
```
Is there any way to improve the ```toString``` to include the expected type (```R``` in this case)?

e..g. Something like```Typeable(Boolean)``` or/and add a seperate method to get the type directly.

milessabin:
Reasonable request. It won't get in to 2.1.0 I'm afraid though (at least not unless a perfectly formed PR lands in my inbox very very soon).

stacycurl:
For some definition of 'soon' ;p

milessabin:
Oh _very_ funny ;-)

rorygraves:
:smiley:

milessabin:
Fixed in #335.

PR #335, opened 2015-02-24
Adds toString to Typeable

khronnuz:
Implements toString for Typeable instances.

### DIFF
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -27,6 +27,8 @@ import scala.reflect.macros.blackbox
  */
 trait Typeable[T] {
   def cast(t: Any): Option[T]
+  def describe: String
+  override def toString = s"Typeable[$describe]"
 }
 
 trait LowPriorityTypeable {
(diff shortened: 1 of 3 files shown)


milessabin:
Is this just waiting for support for intersection types?

khronnuz:
Yes, I submitted the PR as it is in case someone can suggest a fix to it. I am still planing on giving it a spin when have some free time.

core/src/main/scala/shapeless/typeable.scala
```diff
@@ -139,15 +157,17 @@ object Typeable extends TupleTypeableInstances with LowPriorityTypeable {
           else for(e <- o; _ <- e.cast[T]) yield t.asInstanceOf[Option[T]]
         } else None
       }
+      def describe = s"Option[${castT.describe}]"
     }
 
   /** Typeable instance for `Either`. */
   implicit def eitherTypeable[A, B]
-    (implicit castA: Typeable[Left[A, B]], castB: Typeable[Right[A, B]]): Typeable[Either[A, B]] =
+    (implicit castA: Typeable[Left[A, B]], castB: Typeable[Right[A, B]], castC: Typeable[A], castD: Typeable[B]): Typeable[Either[A, B]] =
```
milessabin: I think that here we should be able to get by with just `castC` and `castD`.

core/src/main/scala/shapeless/typeable.scala
```diff
@@ -186,11 +208,12 @@ object Typeable extends TupleTypeableInstances with LowPriorityTypeable {
           if(cc.forall(_.cast[T].isDefined)) Some(t.asInstanceOf[CC[T]])
           else None
         } else None
+      def describe = s"${mCC.runtimeClass.getSimpleName}[${castT.describe}]"
     }
 
   /** Typeable instance for `Map`. Note that the contents will be tested for conformance to the key/value types. */
   implicit def genMapTypeable[M[X, Y], K, V]
-    (implicit ev: M[K, V] <:< GenMap[K, V], mM: ClassTag[M[_, _]], castKV: Typeable[(K, V)]): Typeable[M[K, V]] =
+    (implicit ev: M[K, V] <:< GenMap[K, V], mM: ClassTag[M[_, _]], castKV: Typeable[(K, V)], castK: Typeable[K], castV: Typeable[V]): Typeable[M[K, V]] =
```
milessabin: Similarly here ... I think that we only need `castK` and `castV`.

milessabin:
This looks pretty good to me :-)

If you address my comments above and squash then I think it's good to go ...

core/src/main/scala/shapeless/typeable.scala
khronnuz: When using just castA and castB, the output for Either[Long, String] is something like "Typeable[Either[Left[Long], Right[String]]]", that is I why I added castC and castD, However, if we remove castA and castB, it will fail to require that arguments are Either in the the first place, won't it?

core/src/main/scala/shapeless/typeable.scala
khronnuz: The problem is similar here.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+  trait A
+  trait B
+  class C extends A with B
+
+  @Test
+  def testToString {
+    def typeableString[T](t: T)(implicit tp: Typeable[T]) = tp.toString
+
+    val i: Int = 7
+    assertEquals("Typeable[Integer]", typeableString(i))
```
milessabin: I think this should be mapped back to `Int`

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+
+    val u: Unit = ()
+    assertEquals("Typeable[BoxedUnit]", typeableString(u))
```
milessabin: And this to `Unit`.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+    assertEquals("Typeable[AnyVal]", typeableString(av))
+
+    val ar: AnyRef =  ""
+    assertEquals("Typeable[AnyRef]", typeableString(ar))
+
+    val f: Foo = Foo(0, "", true)
+    assertEquals("Typeable[Foo]", typeableString(f))
+
+    val bi: Bar[Int] = Bar(23)
+    assertEquals("Typeable[Bar[Integer]]", typeableString(bi))
```
milessabin: `Int` again.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+
+    val i1: A with B = new C
+    assertEquals("Typeable[A with B]", typeableString(i1))
+    assertEquals("Typeable[AnyRef with A]", typeableString(new A{}))
```
milessabin: Maybe drop the `AnyRef`?

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+
+    val o: Option[Long] = Some(4l)
+    assertEquals("Typeable[Option[Long]]", typeableString(o))
+
+    val e: Either[Long, String] = Right("")
+    assertEquals("Typeable[Either[Long, String]]", typeableString(e))
+    assertEquals("Typeable[Right[Long]]", typeableString(Right(3l)))
+
+    val l: List[Int] = List(1,2)
+    assertEquals("Typeable[List[Integer]]", typeableString(l))
```
milessabin: `Int`.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+
+    val m: Map[Int, String] = Map(1 -> "one", 2 -> "two")
+    assertEquals("Typeable[Map[Integer, String]]", typeableString(m))
```
milessabin: `Int`.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+
+    assertEquals("Typeable[HNil.type]", typeableString(HNil))
+    val hl = 1 :: "" :: HNil
+    assertEquals("Typeable[Integer :: String :: HNil]", typeableString(hl))
```
milessabin: `Int`.

core/src/test/scala/shapeless/typeable.scala
```diff
@@ -499,4 +499,75 @@ class TypeableTests {
...
+    assertEquals("Typeable[Int(1)]", typeableString(one))
+
+    object FooBar
+    val wFB = Witness(FooBar)
+    type FooBarT = wFB.T
+    val foobar: FooBarT = FooBar
+    assertEquals("Typeable[FooBar.type]", typeableString(foobar))
+
+    val tc = TypeCase[List[Int]]
+    assertEquals("TypeCase[List[Integer]]", tc.toString)
```
milessabin: `Int`.

core/src/test/scala/shapeless/typeable.scala
khronnuz: I can drop it by doing something like `def describe = parents.filter(_.describe != "AnyRef") map(_.describe) mkString" with "`. However, it will also filter cases like an explicit `new AnyRef with A{}`. What you think?

milessabin:
Good stuff! Merging ...

Many thanks ... very much appreciated :-)

milessabin:
Fixes #313.

rorygraves:
Nice - thanks @khronnuz  / @milessabin

PR #337, opened 2015-03-07
Added non emptyness constraint to hlistconstraints.scala

edmondo1984:
Follow question on the mailing list


milessabin:
Contribution very much appreciated, but this scenario is already handled by `IsHCons`.

PR #341, opened 2015-03-21
Don't rely on macro for the.apply

ceedubs:
This seems to work for all of the uses in the tests. Is there a reason a
macro is needed for this?


ceedubs:
I realized that this won't work for value classes. I'll instead submit a PR that adds a test for value classes.

PR #342, opened 2015-03-23
Add unit test for `the` with value classes

ceedubs:
This unit test wouldn't have compiled with my proposed changes in #341,
which would have alerted me that my changes were a bad idea.


milessabin:
Thanks for that! Merging :-)

PR #348, opened 2015-04-01
fix #280

### LINKED ISSUE #280: Consider the value of value classes, for instance for the syntax extension methods
It might be worth considering making the syntax extension method classes value classes, such as HListOps, to avoid unnecessary boxing.

lambdista:
Since @milessabin said PR's welcome I changed the *Ops* classes to extend `AnyVal`. Furthermore since there are many examples of @non's spire classes extending `AnyVal` such as [Ops](https://github.com/non/spire/blob/master/core/src/main/scala/spire/syntax/std/Ops.scala) and [Literals] (https://github.com/non/spire/blob/master/core/src/main/scala/spire/syntax/Literals.scala), I don't think Erik is totally against value classes. As a final example also Scala's [StringOps](https://github.com/scala/scala/blob/v2.11.6/src/library/scala/collection/immutable/StringOps.scala) extends `AnyVal` so I guess it should not be a problem following this *convention* to avoid boxing.

### DIFF
--- a/core/src/main/scala/shapeless/syntax/coproduct.scala
+++ b/core/src/main/scala/shapeless/syntax/coproduct.scala
@@ -25,7 +25,7 @@ package syntax
  *
  * @author Miles Sabin
  */
-final class CoproductOps[C <: Coproduct](c: C) {
+final class CoproductOps[C <: Coproduct](val c: C) extends AnyVal {
   import ops.coproduct._
 
   /**
--- a/core/src/main/scala/shapeless/syntax/nat.scala
+++ b/core/src/main/scala/shapeless/syntax/nat.scala
@@ -22,7 +22,7 @@ package syntax
  *
  * @author Dale Wijnand
  */
-final class NatOps[N <: Nat](n : Nat) {
+final class NatOps[N <: Nat](val n : Nat) extends AnyVal {
   import ops.nat._
 
   /**
--- a/core/src/main/scala/shapeless/syntax/records.scala
+++ b/core/src/main/scala/shapeless/syntax/records.scala
@@ -25,7 +25,7 @@ import tag.@@
  * 
  * @author Miles Sabin
  */
-final class RecordOps[L <: HList](l : L) {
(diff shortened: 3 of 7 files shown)


milessabin:
PR's very welcome! Thanks :-)

I'm knee deep in adding `with Serializable` to everything at the moment, ops traits included, so unfortunately this will cause conflicts. Would you mind holding off on this until I've merged (will be later today)?

lambdista:
Thank you Miles. :-) 

Let me know if and when I need to do anything about this PR.

milessabin:
Comparing notes with @mpilquist it looks like this will break binary compatibility with 2.1.0 for anything that uses syntax ... see [this](https://gist.github.com/mpilquist/aa7d0e4b7372d6bb97af) example.

I'm still happy to merge this PR, but not until after 2.2.0 has been released.

lambdista:
That's fine and pretty reasonable. Thank you.

milessabin:
Unfortunately I've committed to maintaining binary compatibility through the 2.2.x series, so this is going to have to wait for 2.3.0. Sorry for the hold up :-(

lambdista:
Hi Miles, that's reasonable. Thank you.

milessabin:
Thanks for your patience :-)

milessabin:
Finally about to break binary compatibility with 2.2.0, so this can be merged at long last! Would you mind rebasing against master?

lambdista:
I'm sorry Miles but, at the moment, the only technological device I have access to is my smartphone. I force myself to do this a couple weeks a year. :-)

I'll be back in front of a *real* computer next Monday.

I hope this is not a problem.

Sorry for the delay I may cause.

milessabin:
No problem at all ... it's going to take more than a week for the dust to settle on #435 :-)

lambdista:
OK, thank you. :-)

lambdista:
Done! Let me know if there's something wrong.

milessabin:

PR #362, opened 2015-04-14
Fix #354: Make PartialOrdering.tryCompare equality checks less specific

### LINKED ISSUE #354: Make PartialOrdering.tryCompare equality checks less specific
Tests [such as this](https://github.com/milessabin/shapeless/blob/d59463a0c093a21345754226702cbd807015b6ca/core/src/test/scala/shapeless/coproduct.scala#L370) fail in JS as the actual test result is `some(-1)`.

In fact, the API only specifies that `lteq` will return negative so the tests should be made less specific accordingly.

The should be applied to master as well as JS branches.

InTheNow:


milessabin:
Looks good. Will merge ASAP (still wavering on 2.2.0-RC5).

milessabin:
There's going to be an RC5, so merging this now. Thanks :-)

PR #363, opened 2015-04-14
Fix #356: TypeOperatorTests.testRejectBogus fails in JS tests

### LINKED ISSUE #356: TypeOperatorTests.testRejectBogus fails in JS tests
`assert(Try(the.Foo).isFailure)` [fails in test](https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/typeoperators.scala#L151) as shapeless macro converts to `Unit` that is represented as `undefined` in sjs. As a result, a fatal error is thrown that is not caught by the `Try`.

This can be changed in master to a regular `try...catch`

InTheNow:


japgolly:
Why does this fail in JS?

InTheNow:
As per #356, it is because `Try` does not catch fatal errors and as `Unit` maps to `undefined`, a fatal error is thrown.

milessabin:
There's going to be an RC5, so merging this now ... thanks :-)

PR #364, opened 2015-04-14
Fix #359: Replace JUnit with integral test framework

### LINKED ISSUE #359: Replace JUnit with integral test framework
See  https://github.com/milessabin/shapeless/issues/347 for rational/discusssion.

InTheNow:


InTheNow:
fyi, this patch also "makes" the typeable tests pass, removing the var assignment but putting the instantiations to a higher scope:

```
@@ -28,6 +28,9 @@ class TypeableTests extends SpecLite {
   trait A
   trait B
   class C extends A with B
+  val wOne = Witness(1)
+  type One = wOne.T
+  val one: One = 1
 
   "TypeableTests" should {
 
@@ -534,11 +537,7 @@ class TypeableTests extends SpecLite {
     val cpd: CP = Coproduct[CP](2.0)
     assertEquals("Typeable[Double :+: Boolean :+: CNil]", typeableString(cpd))
 
-    val wOne = Witness(1)
-    type One = wOne.T
-    val one: One = 1
-    val result = typeableString(one)
-    assertEquals("Typeable[Int(1)]", result )
+    assertEquals("Typeable[Int(1)]",  typeableString(one)  )
 
     object FooBar

```

milessabin:
Awesome work! Many thanks!

project/Build.scala
```diff
@@ -91,7 +93,8 @@ object ShapelessBuild extends Build {
       libraryDependencies ++= Seq(
         // needs compiler for `scala.tools.reflect.Eval`
         "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
-        "com.novocode" % "junit-interface" % "0.7" % "test"
+        "com.novocode" % "junit-interface" % "0.7" % "test",
+        "org.scala-sbt" %  "test-interface" % "1.0" % "test"
       ),
```
milessabin: JUnit still here?

project/Build.scala
```diff
@@ -108,7 +111,8 @@ object ShapelessBuild extends Build {
...
```
milessabin: And here?

project/Build.scala
InTheNow: yes, scratch and example never ported.

I'll quickly do that now

InTheNow:
Fixed 'n squashed

milessabin:
Great stuff :-)

A few comments ...

First I don't quite understand the distribution of source files under speclite between `src/main` and `src/test` and the declared `"test -> test"` dependencies. It looks to me that the sources under `speclite/src/test` are parts of the implementation of speclite rather than being tests of speclite, so shouldn't they be under `src/main`?

There's a slight problem with the package declaration of the sources under `speclite/src/main/scala` ... they're in `org.scalajs.testinterface` which I guess is a hangover. Can we put the whole lot (ie. including the stuff which is under `speclite/src/test` in package `shapeless.speclite` and move them to subdir `speclite/src/main/scala/shapeless/speclite`?

The only other thing is that the time for `test:compile` after a clean has gone from 125s with JUnit on this machine to 165s with speclite. It's possible that part of that is accounted for by the fact that `test:compile` will recompile the specilte code itself currently because it's under `src/test`, but either way it'd be nice to get a handle on what's increasing the compile times here.

InTheNow:
> ..  source files under speclite between src/main and src/test and the declared "test -> test" dependencies

There is a macro defined in `testinterface.TestUtils` that is used in the Speclite files, hence they cannot all be in `main`. Of course, there are other alternatives, one of which is to have another project, "testinterface".

> ... in org.scalajs.testinterface which I guess is a hangover

The reason for both the files is so that the main test framework crosscompiles without any changes - these files are essentially stubs. 

An alternative, should you wish, is to not have these two files local but add a build dependency to  the JVM build `libraryDependencies += "org.scala-js" % "scalajs-stubs_2.11" % "0.6.3"`. This would of course also solve the first issue, but also not fit in with your general strategy. I only mention this point so that you are aware of the option.

> get a handle on what's increasing the compile times here

I'm 99% sure its just because of the new files - let me get some concrete figures.

milessabin:
Gotcha.

I think having a separate project for the macro is the right thing to do. I'd also prefer you used the same branching vs. cross building approach for 2.10.x vs. 2.11.x ... if nothing else I think it would be very confusing to have both strategies in play in the same (overall) project.

InTheNow:
> There is a macro defined in testinterface.TestUtils 

This was only intended for scala/scalajs interop, as scalajs does not support runtime reflection. I thus intend to remove this completely, in favour of runtime reflection of test instantiation. Hence, the two sjs files, and hence the new sbt subproject, can be eliminated. 

@milessabin OK?

milessabin:
Sounds good to me :-)

milessabin:
What's the situation with this now? Has the Scala.js support for JUnit got to the point where we can close this without losing anything important?

nicolasstucki:
We are still in the process of porting JUnit to Scala.js you can follow https://github.com/scala-js/scala-js/issues/1690. But it will still take some time because we still need to port several components from java util collections.

milessabin:
I'm happy to leave this PR hanging around until you're close to 100% confident that the JUnit port will be successful :-)

InTheNow:
@nicolasstucki Are you in a position to say now what shapeless would bring in when using your port? For example, would you rely on Macro Annotations from macro paradise?

sjrd:
There won't be macro-paradise. It will be an additional compiler plugin, to be used only when compiling the test suite on JS (so it won't affect the production JS code nor the tests on the JVM).

InTheNow:
@sjrd Cool. I couldn't see any refs to macro-paradise, just trying to think of some _potential_ issues in advance - and that was my main one.

So for me, best is to have this PR dangling around until it can be replaced by a new one

milessabin:
Okey doke ... that's fine by me :-)

InTheNow:
@sjrd @nicolasstucki Just doing some house keeping - how is the Junit port coming on? Only asking so I can prioritize other tasks, no "hassle" intended

nicolasstucki:
@InTheNow JUnit porting process is advanced. Still have a couple of features to finish. I believe that we have ported all you need for your tests.

milessabin:
JUnit support has landed in Scala.js 0.6.6, so I'm going to close this now.

PR #365, opened 2015-04-14
Max[A, B] type class.

alexknvl:
I feel bad about basically just copying the `Min[A, B]` code with minimal changes and pasting my name over it, but here it is :)


stacycurl:
Don't feel bad, it's very useful that missing bits like these are added.
You forgot to add yourself to the contributors file though.

milessabin:
Nice work! Thanks :-)

Merging this might have to wait until after 2.2.0, but there might be an RC5, in which case this is sufficiently low risk that it can probably go in.

milessabin:
There is going to be an RC5 so merging now. Thanks again :-)

PR #366, opened 2015-04-14
Fields methods

alexarchambault:
So here they are. Not the most complex type classes out there, but these can still be useful (to @mpilquist in particular).


milessabin:
Wonderful! Thanks! Will merge ASAP (I'm still wavering on RC5).

milessabin:
There will be an RC5, so merging ... thanks again :-)

PR #372, opened 2015-04-25
Generic for overlapping types

### LINKED ISSUE #370: Support Generic derivation for overlapping types
Given the following ADT
```
sealed trait Root

sealed trait A extends Root
case class A1(s: String) extends A

sealed trait B extends Root
case class B1(s: String) extends B

case class AB(s: String) extends A with B
```
`Generic[Root]` fails to compile in 2.1.0 and 2.2.0-RC4.

It would be nice to have
`Generic[Root]{type Repr = A1 :+: B1 :+: AB :+: CNil}` 
as a result

notxcain:
This PR resolves #370.

### DIFF
--- a/core/src/test/scala/shapeless/generic.scala
+++ b/core/src/test/scala/shapeless/generic.scala
@@ -114,6 +114,13 @@ package GenericTestsAux {
 
   sealed trait Base[BA, BB]
   case class Swap[SA, SB](a: SA, b: SB) extends Base[SB, SA]
+
+  sealed trait Overlapping
+  sealed trait OA extends Overlapping
+  case class OAC(s: String) extends OA
+  sealed trait OB extends Overlapping
+  case class OBC(s: String) extends OB
+  case class OAB(i: Int) extends OA with OB
 }
 
 class GenericTests {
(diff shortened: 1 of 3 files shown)


milessabin:
Many thanks ... much appreciated! Merging :-)

notxcain:
:+1:

core/src/test/scala/shapeless/generic.scala
```diff
@@ -265,6 +272,17 @@ class GenericTests {
     typed[AbstractSingle](s1)
   }
 
+  @Test
+  def testOverlappingCoproducts {
+    val gen = Generic[Overlapping]
+    val o: Overlapping = OAB(1)
+    val o0 = gen.to(o)
+    typed[OAB :+: OAC :+: OBC :+: CNil](o0)
```
OlivierBlanvillain: I think it's a bit weird to auto-flatten the type hierarchy like this, wouldn't it make sense to infer `type Repr = OA :+: OB :+: CNil` instead?

milessabin: I could see a case being made for that.

PR #375, opened 2015-04-26
corrected appliedTypTree1 to manage SingleType

mandubian:


milessabin:
Thanks for that! Merged :-)

It turned out that the last two cases could be simplified a little: https://github.com/milessabin/shapeless/commit/0be012efa8b7054b126af0e672b8db56cf6de78c.

mandubian:
great!
I must say `reflect` and all those functions that do something but without doc about it is a kind of torture...

PR #377, opened 2015-04-29
Made ToTraversable for List serializable

alexarchambault:
This PR makes `ToTraversable` / `ToSized` for `List` serializable.

For this, it puts a custom `CanBuildFrom` instance for `List` into scope (under `shapeless.syntax`), which implements `Serializable`, and which is found instead of the one provided by the standard library. The caveat is that this replaces some `CanBuildFrom` from the standard library by some from shapeless, as soon as one imports `shapeless.syntax._`.

Other serializable `CanBuildFrom` like these could be implemented, like in https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/CanBuildFrom.scala

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -529,9 +529,14 @@ object hlist {
    * Type class supporting conversion of this `HList` to a `M` with elements typed
    * as the least upper bound Lub of the types of the elements of this `HList`.
    *
+   * Serializable if the `CanBuildFrom`s it implicitly finds are too.
+   * Note that the `CanBuildFrom`s from the standard library are *not*
+   * serializable. See the tests for how to make your own serializable
+   * `CanBuildFrom` available to `ToTraversable`.
+   *
    * @author Alexandre Archambault
    */
-  trait ToTraversable[L <: HList, M[_]] extends DepFn1[L] {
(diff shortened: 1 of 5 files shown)


milessabin:
This is great ... thanks! :-)

I'm not so sure about adding that `CanBuildFrom` instance directly to shapeless though, and definitely not in a package `shapeless.syntax`.

There's nothing specific to shapeless in there, and it seems likely that many other projects might benefit from a collection of Java-serializable `CanBuildFrom` instances ... so why not split this out as a completely independent project?

I think for now I'd be happy to merge this (ASAP, for RC5) but without `shapeless.syntax` or the instance defined in it. Maybe move that to an example?

alexarchambault:
The new `CanBuildFrom` can be moved elsewhere yes... (But then, there are `Serialization` everywhere in shapeless, so why not an additional `CanBuildFrom`?)

It could be put it in an external library yes. I'll update the PR.

milessabin:
Thanks :-)

alexarchambault:
Just updated the PR. I simply moved the additional `CanBuildFrom` to the tests. (I could have added an example too, but it would just have been some cut-n-paste from the tests... We can tell to look at the added definition in the tests.)

milessabin:
Many thanks! Merged :-)

PR #383, opened 2015-05-12
Better prepending to HNil

alexarchambault:
It now returns directly the prepended HList, instead of rebuilding it recursively.


alexarchambault:
It did contrary to the appearances in the current tests.

It succeeds without an implicit here:
https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/hlist.scala#L391-L392, and indeed returns the same ref (changed the tests to `assertTypedSame`, still pass). But when one prepends directly (with `l ::: HNil` say, not `prependToHNil(l)`), it finds a higher priority implicit, and returns a copy of the list.

This PR changes that.

milessabin:
Many thanks ... merging :-)

Issue #382, opened 2015-05-12
Reversing a coproduct twice not returning the original coproduct


alexarchambault:
if some types are repeated in the coproduct:

```scala
@ type C = Int :+: Boolean :+: String :+: Int :+: CNil
defined type C
@ val ci1: C = Inl(1)
ci1: C = 1
@ val ci2: C = Inr(Inr(Inr(Inl(1))))
ci2: C = 1
@ ci1.reverse
res4: Int :+: Boolean :+: String :+: Int :+: CNil = 1
@ ci2.reverse
res5: Int :+: Boolean :+: String :+: Int :+: CNil = 1
@ ci1 == ci1.reverse.reverse // true as expected
res6: Boolean = true
@ ci2 == ci2.reverse.reverse // should be true
res7: Boolean = false
```

alexknvl:
It is easier to see if you temporarily disable `toString()` in `Inl` and `Inr`:
```scala
scala> type C = Int :+: Boolean :+: String :+: Int :+: CNil
defined type alias C

scala> val ci1: C = Inl(1)
ci1: C = Inl(1)

scala> val ci2: C = Inr(Inr(Inr(Inl(1))))
ci2: C = Inr(Inr(Inr(Inl(1))))

scala> ci1.reverse.reverse
res0: shapeless.:+:[Int,shapeless.:+:[Boolean,shapeless.:+:[String,shapeless.:+:[Int,shapeless.CNil]]]] = Inl(1)

scala> ci2.reverse.reverse
res1: shapeless.:+:[Int,shapeless.:+:[Boolean,shapeless.:+:[String,shapeless.:+:[Int,shapeless.CNil]]]] = Inl(1)
```

PR #384, opened 2015-05-12
Makes reversing twice a coproduct not changing the coproduct

alexarchambault:
It used to change it for coproducts with repeated types.

Fixes https://github.com/milessabin/shapeless/issues/382


stacycurl:
Since that's basically a rewrite I've no qualms about my name being dropped
from that type class.

alexarchambault:
I wouldn't say a rewrite, more a transposition of the `Reverse` of HLists to coproducts.

(I mean, I clearly don't deserve full credit for that :-). It could be @milessabin who wrote the implementation for HLists then.)

milessabin:
Many thanks ... merging :-)

PR #385, opened 2015-05-12
Topic/rotates

alexarchambault:
This PR makes the implementation of `RotateLeft` / `RotateRight` for coproducts a simple transposition of the one for HLists.

It also makes the one of HLists safer, with no case that can return a plain wrong type class (the `noop`* ones could be called with any hlist and any nat).

The added tests for coproduct rotates were suggested by @alexknvl


milessabin:
Many thanks ... merging :-)

PR #386, opened 2015-05-15
Scalajs 2.10.x

benhutchison:

PR #387, opened 2015-05-15
Scalajs 2.11.x

benhutchison:


InTheNow:
As we have to use 0.6.3 of sjs for 2.10, can i suggest that you just change the sjs version both here and for 2.10 in https://github.com/milessabin/shapeless/pull/386 . 

Then, assuming the tests are still ok, we are done for now

InTheNow:
@benhutchison Actually, forget last request. I forgot you are in AUS tz so easier if I do myself

PR #393, opened 2015-05-25
Adds script to cross build shapeless from multiple branches.

khronnuz:
This script allows for Shapeless cross-build model, which is based on branches. Since the build definition itself can change between branches, e.g. for changing dependencies and Scala version, it is an easy task, if possible, to do using the SBT release plugins.

Note that it is still missing the actual commands that sign and release, but should be simple to add on `release ` function.

### DIFF
--- a/release.scalascript
+++ b/release.scalascript
@@ -1,0 +1,242 @@
+#!/bin/sh
+exec scala "$0" "$@"
+!#
+
+/*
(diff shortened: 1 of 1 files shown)


milessabin:
This will be a huge time saver ... many, many thanks :-)

Merging ...

PR #394, opened 2015-05-28
Extract function from materializeImpl

fthomas:
This extracts a function from `SingletonTypeMacros.materializeImpl` for extracting the value of a singleton type. This is just a refactoring that makes it easier to add `WeakWitness` (https://github.com/milessabin/shapeless/compare/master...fthomas:topic/WeakWitness) to [refined](https://github.com/fthomas/refined).


fthomas:
The motivation for this change is to make the implementation of `materializeWeakWitnessImpl` in https://github.com/fthomas/refined/blob/37a16ada5b558108a45ec1bb26d2ede34f1ade7d/src/main/scala/eu/timepit/refined/internal/WeakWitness.scala a lot simpler.

PR #400, opened 2015-05-29
Merge new integral test framework from 2.11 to 2.10

InTheNow:


milessabin:
What's the situation with this now? Has the Scala.js support for JUnit got to the point where we can close this without losing anything important?

milessabin:
JUnit support has landed in Scala.js 0.6.6, so I'm going to close this now.

PR #401, opened 2015-06-01
Added method modify to Zipper

petomat:


milessabin:
Thanks. A couple of comments ...

Could you add a standalone op type class (which can be implemented in terms of `Get` and `Put` exactly as you have now) ... the rationale is that it's helpful to need only a single type class instance where these a propagated as implicit arguments through a series of nested calls.

And would you mind adding some tests?

milessabin:
Oh, and then could you squash the resulting commit. Thanks :-)

petomat:
I will do my best :-)

petomat:
Sorry for the delay. I am still working on it...

milessabin:
No worries :-)

petomat:
@milessabin Would you mind having a look at my proposed commit? I hope it's close to what you have in mind.

milessabin:
That looks absolutely spot on :-) Many thanks ... merging ...

Issue #406, opened 2015-06-08
Compile time compatibility issue with spray 1.3


alexarchambault:
Raised by @ngbinh on gitter, can be reproduced with:
```
git clone https://github.com/spray/spray.git -b release/1.3
cd spray
# Edit project/Build.scala, add setting
#   `libraryDependencies += "com.chuusai" %% "shapeless" % "2.2.1"`
# to sprayRoutingShapeless2Tests
sbt spray-routing-shapeless2-tests/test
```
Output like https://gist.github.com/alexarchambault/32c704e52eb7884a71c2, with in particular:
```
[error] missing or invalid dependency detected while loading class file 'AnyParamDefMagnet2.class'.
[error] Could not access type Aux in trait shapeless.ops.hlist.LowPriorityPrepend,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'AnyParamDefMagnet2.class' was compiled against an incompatible version of shapeless.ops.hlist.LowPriorityPrepend.
```

PR #407, opened 2015-06-08
2.2.0 / 2.2.1 binary compatibility

alexarchambault:
This PR fixes (possibly not all?) the backward compatibility issues of 2.2.1, and adds the sbt-mima plugin (https://github.com/typesafehub/migration-manager), to check that no changes that break backward compatibility are introduced.

sbt-mima does not seem to cover all the compatibility issues, in particular the one of https://github.com/milessabin/shapeless/issues/406 (type definition that was moved) would have gone undetected apparently. https://github.com/milessabin/shapeless/issues/406 is fixed anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,9 @@ script:
           "${TRAVIS_BRANCH}" == "scalajs-2.11.x" || "${TRAVIS_BRANCH}" == "scalajs-2.10.x") &&
          $(cat version.sbt) =~ "-SNAPSHOT"
     ]]; then
-      sbt ++${TRAVIS_SCALA_VERSION} test publish ;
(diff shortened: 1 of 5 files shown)


alexarchambault:
To check for compatibility with 2.2.1, change the line with `previousArtifact` in `project/Build.scala`. (Forward compatibility issues appear - methods in new version not present in old one, but no backward compatibility ones.) But issues like https://github.com/milessabin/shapeless/issues/406 are not checked (and are likely to be present, as some types were moved for compatibility with 2.2.0).

alexarchambault:
FYI, scalacheck now seems to be using sbt-mima too: https://github.com/rickynils/scalacheck/blob/master/project/MimaSettings.scala

milessabin:
Fantastic! Merging :-)

Issue #195, opened 2014-07-28
Add sbt-typelevel plugin to the build


milessabin:
See documentation [here](https://github.com/typelevel/sbt-typelevel).

milessabin:
Closing this now ... it's been superseded by the release script and direct inclusion of MiMa in #407. My apologies to @andrelfpinto :-(

PR #409, opened 2015-06-12
Added sbt-typelevel plugin to the build.

andrelfpinto:
Tried to get the lowest hanging fruit (at least in my point of view): issue #195. Let's see if I reached it.


non:
@milessabin you might want to test this a bit. I think there are outstanding issues with sbt-typelevel and I'm not sure how active the project is. I'm currently using the plugin on only one project and it is a bit idiosyncratic.

milessabin:
I think the main thing I was looking to gain from the sbt-typelevel plugin was the release plugin configuration and the support for MiMa. The first of those has been superseded by the new multi-branch release script, and the second has just been addressed in #407.

So I'm really, really sorry to have led up the garden path @andrelfpinto, but I'm afraid I this isn't actually needed any more :-(

I hope this doesn't put you off contributing in future, I think it was a case of unfortunate timing in this case.

My apologies again ...

andrelfpinto:
No problem, @milessabin. I will try to get the next lowest fruit as soon as I can. I'm just waiting for me to get a little taller: it's a wonderful project but it's a bit out of my league.

Issue #413, opened 2015-06-16
Consider type classes for smashing hlists and coproducts


travisbrown:
First for some motivation: in the [Finch](https://github.com/finagle/finch) project we're using Shapeless in a couple of places, including a `Router[A]` type that may match a path and return an `A`. We provide an `:+:` method that allows the user to compose routers of different types. This makes combining endpoints into something we can serve a lot cleaner, since we don't have to take every router all the way to a Finagle service—we just need something that we know how to turn into a Finagle service.

For example, we could write something like this:

```scala
val router =
  Get / "foo" / long   :+:
  Get / "foo" / string :+: 
  Get / "baz" / string /> Baz(_)
```

Now we have a `Router[Long :+: String :+: Baz :+: CNil]`, and assuming we have the appropriate type class instances for all the types involved, we can turn this into a web service with `router.toService`.

Finch also allows users to combine coproduct routers with `:+:` in such a way that coproducts are flattened. For example, we might have the following:

```scala
val userEndpoints: Router[User :+: List[User] :+: Service[HttpRequest, User] :+: CNil] =
  Get  / "user" / long /> getUser(_) :+:
  Get  / "user" /> getUsers          :+:
  Post / "user" /> createUser

val ticketEndpoints: Router[Ticket :+: HttpResponse :+: CNil] =
  Get    / "ticket" / long /> getTicket(_) :+:
  Delete / "ticket" / long /> deleteTicket(_)
```

We can serve these routers together like this:

```scala
Await.ready(Httpx.serve(":8080", (userEndpoints :+: ticketEndpoints).toService))
```

This works because `:+:` just smashes the coproducts together into a big coproduct where all the elements are types we know how to serve. This would be a lot more complicated if `userEndpoints :+: ticketEndpoints` were a coproduct of coproducts.

We _could_ introduce a separate operator for this kind of concatenation, but the idea is that we want to keep the API simple. In general Finch users aren't writing Shapeless types in their code—they just put together alternatives with our `:+:` and then turn the whole thing into a service.

This is all supported by a [`CoproductSmash[A, B]` type class](https://github.com/travisbrown/finch/blob/topic/coproduct-smash/core/src/main/scala/io/finch/route/CoproductSmash.scala) that concatenates `A` and `B` if they're both coproducts, prepends `A` to `B` if only `B` is a coproduct, appends `B` to `A` if only `A` is a coproduct, and returns `A :+: B :+: CNil` if neither is a coproduct.

[...]

PR #414, opened 2015-06-18
Add new adjoin type class

travisbrown:
This is an initial sketch of an implementation of the type class described in #413 (unified for `HList` and `Coproduct`). There was [some discussion of the name on Gitter](https://gitter.im/finagle/finch?at=5582eb321c3ba5ef5bff10db) this morning. I don't think "adjoin" is perfect, but IMO it's better than the alternatives (with "concat", "abut", "smash", and "conjoin" as the runners up).

I can add some documentation, examples, etc. this evening—I just wanted to get something up sooner rather than later.


milessabin:
Given the implementation style, it would seem fairly natural to generalize this beyond the binary case ... thoughts?

I'm also thinking that this looks like something that belongs in the `ops` package, possibly with methods defined on the `HList` and `Coproduct` syntax traits?

travisbrown:
It works beyond the binary case now—that's just not represented in the tests (but should / will be). Where are you imagining it'd live under `ops`? I agree that it would make sense for it to be there, but it cuts across the current organization, in that it doesn't fit in either `hlist` or `coproduct`.

milessabin:
Oh! Sorry :-) 

I thing it can live in its own ops.align object for now.

vkostyukov:
What do you think about `Flatten`instead of `Adjoin`? Looking at the implementation it reminds me Scala's `.flatten`. Eg: `List(List(1), List(2), 3, 4).flatten = List(1, 2, 3,4)`.

milessabin:
That's a reasonable suggestion, but I think we're too used to the idea that `flatten` is equivalent to `flatMap(identity)` which is already defined for `HList` with subtly different semantics.

travisbrown:
I've started a simplified version of Finch's `Router` that I'm going to add to the examples here, but it's not quite done. If this type class doesn't make it into 2.2.3 that's not a big deal—it'd be nice to be able to use this in Finch 0.8.0, but not necessary if the timing doesn't work out.

milessabin:
I think there's only one thing I want to change here now: I'd rather the examples had asserts than printlns: partly so regressions can be caught, but mainly so that a casual reader can see what the expected result is without have to actually run the example.

It's not a blocker for merging, but I have a few things to do this morning so I'll leave it here on the offchance that you'll be able to deal with that before I come back to it. No worries if not.

milessabin:
Awesome ... thanks :-) Merging.

PR #416, opened 2015-06-21
Fix Remove and add reinsert and reinsertAll methods

travisbrown:
This fixes two small bugs (or at least inconsistencies and annoyances) in `Remove` and adds new `reinsert` operations that "undo" `removeElem` and `removeAll`.

The first bug is just in the error message for `Remove` when the requested type to remove appears more than once:

```scala
scala> Remove[String :: Int :: Int :: HNil, Int]
<console>:15: error: Implicit not found: shapeless.Ops.Remove[shapeless.::[String,shapeless.::[Int,shapeless.::[Int,shapeless.HNil]]], Int]. You requested to remove an element of type Int, but there is none in the HList shapeless.::[String,shapeless.::[Int,shapeless.::[Int,shapeless.HNil]]].
              Remove[String :: Int :: Int :: HNil, Int]
                    ^
```

There definitely is "an element of type Int" in the list.

The second bug is an inconsistency. This compiles:

```scala
implicitly[Remove.Aux[Int :: String :: Int :: HNil, Int, (Int, String :: Int :: HNil)]]
```

But this does not:

```scala
implicitly[Remove.Aux[String :: Int :: Int :: HNil, Int, (Int, String :: Int :: HNil)]]
```

And there's no good grounds for distinguishing between the two. Moving `recurse` into a lower-priority trait fixes the issue. I've also added a test case for the version that previously didn't compile.

Finally I've added `reinsert` methods to `Remove` and `RemoveAll` that make it possible to put an item (or items) back in the removed places. This is needed for e.g. deriving "partial" type class instances for case classes (such as `DecodeJson[Id => User]` where `User` has a `Id` field).

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -913,19 +913,13 @@ object hlist {
    *
    * @author Stacy Curl
    */
-  @implicitNotFound("Implicit not found: shapeless.Ops.Remove[${L}, ${E}]. You requested to remove an element of type ${E}, but there is none in the HList ${L}.")
-  trait Remove[L <: HList, E] extends DepFn1[L] with Serializable
+  @implicitNotFound("Implicit not found: shapeless.Ops.Remove[${L}, ${E}]. You requested to remove an element of type ${E}, but there is no unique candidate in the HList ${L}.")
+  trait Remove[L <: HList, E] extends DepFn1[L] with Serializable {
+    def reinsert(out: Out): L
+  }
 
-  object Remove {
-    def apply[L <: HList, E](implicit remove: Remove[L, E]): Aux[L, E, remove.Out] = remove
-
+  trait LowPriorityRemove {
     type Aux[L <: HList, E, Out0] = Remove[L, E] { type Out = Out0 }
-
-    implicit def remove[H, T <: HList]: Aux[H :: T, H, (H, T)] =
-      new Remove[H :: T, H] {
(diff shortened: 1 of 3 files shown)


milessabin:
MiMa is complaining ...

milessabin:
Bump ...

travisbrown:
Hmm, somehow I keep missing GitHub notification for Shapeless…

The new `reinsert` method is necessary to get the current functionality of this PR, so it's not going to pass MiMa as is. I ended up needing some additional changes to remove for my [incompletes](https://github.com/travisbrown/incompletes) project, though, so it might make sense to split this PR into two parts—one that could be merged now that would fix the inconsistencies in `Remove`, and another (at some point later on) that would add the reinsertion functionality.

milessabin:
shapeless-2.3.0 is coming fairly soon (quite possibly next) and we can break binary compatibility then.

milessabin:
I've just merged #435 which breaks binary compatibility, so we're on the 2.3.0-SNAPSHOT and this can go in ... can you rebase/unbreak this so I can merge?

travisbrown:
@milessabin Thanks for the update—I've rebased. As I mention above, my `reinsert` here doesn't actually work for the incomplete instances derivation in circe (since I need to support referring to fields in the record either with or without their labels). So I don't have a definite use case for this `reinsert` at the moment. I can pull it if you'd like, but I think it's probably useful enough to include here.

milessabin:
Yeah, I agree ... merging as is. Many thanks :-)

PR #425, opened 2015-07-08
Add left fold for Coproducts

AndreasKostler:
Left fold for Coproducts.

### DIFF
--- a/core/src/main/scala/shapeless/ops/coproduct.scala
+++ b/core/src/main/scala/shapeless/ops/coproduct.scala
@@ -373,6 +373,32 @@ object coproduct {
         }
   }
 
+  trait LeftFolder[C <: Coproduct, In, F] extends DepFn2[C,In] with Serializable {
+    type Out = In
+  }
+
+  object LeftFolder {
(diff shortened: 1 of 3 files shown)


ceedubs:
The constraint that the output of the `foldLeft`is `In` seems unnecessary to me.

It seems like as long as there is a `Case2[F, In, A]` for all A in the Codproduct, you should be able to return the output type of the matching `Case`. I'm guessing the current constraint is because of [this line](https://github.com/milessabin/shapeless/pull/425/files#diff-7f03e74196d53d11cd9f05dd6b87ac7cR387), but `case Inr` when the input is `H :+: CNil` is impossible, so we shouldn't need to support that case.

I don't think I articulated this particularly well. Feel free to ask me to clarify.

AndreasKostler:
Not sure I understand you fully but I pushed a change to reflect what I *think* you mean ;)

ceedubs:
Yes, this is exactly what I meant! :+1:

AndreasKostler:
Awesome!

milessabin:
Sorry it's taken me so long to get to this ... I've been bogged down with 2.12.0-M2 compatibility issues.

Anyhow, looks good :-) Many thanks ... merging!

PR #427, opened 2015-07-09
Add Priority, and Implicit type class

alexarchambault:
As discussed on [gitter](https://gitter.im/milessabin/shapeless?at=559e6afb26de773708e6798a) (for `Priority`).

Types with cycles are now fine (added to the tests).

It indeed addresses issues with the current `Orphan`, see the tests marked with "Fails with the current Orphan". Don't know about `WrappedOrphan`, I don't really know how to set it up here.

I renamed the auxiliary ADT to `Attempt` and its cases `Success` and `Fallback`, and added some inline doc too.

### DIFF
--- a/core/src/main/scala/shapeless/implicit.scala
+++ b/core/src/main/scala/shapeless/implicit.scala
@@ -1,0 +1,63 @@
+package shapeless
+
+/**
+ * Represent a `T` made of implicits.
+ *
+ * E.g. Looking for an implicit `Implicit[A :: B :: HNil]` will
+ * trigger implicit lookups for `A` and `B`, and return these
+ * wrapped in an `Implicit[A :: B :: HNil]`.
+ *
+ * Looking for an implicit `Implicit[A :+: B :+: CNil]` will
+ * trigger first an implicit lookup for a `A`. If it fails,
+ * an implicit `B` will be looked up. The result of these lookup
+ * will be wrapped in an `Implicit[A :+: B :+: CNil]`, whose
+ * value will be either a `A` (if an implicit `A` was found),
+ * or a `B` (if no implicit `A` was found, but an implicit `B` was).
+ *
+ * `T` can also be a product or an ADT. Its generic counterpart will
+ * looked upon as described above, then converted to a `T`.
+ */
+trait Implicit[T] {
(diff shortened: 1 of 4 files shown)


alexarchambault:
I changed the name to `Attempt`, but I know think that something with `Priority` in it should be better... Maybe just `Priority`, with constructors `High` / `Low`.

alexarchambault:
Also added an `Implicit` type class, that illustrates how I intend to use `Priority`.

alexarchambault:
I'm testing this on the [`develop` branch of scalacheck-shapeless](https://github.com/alexarchambault/scalacheck-shapeless/tree/develop).

I'm sometimes running into `Lazy` generating wrong code, when an error was expected anyway ([1](https://github.com/alexarchambault/scalacheck-shapeless/tree/prioritycrash/1), [2](https://github.com/alexarchambault/scalacheck-shapeless/tree/prioritycrash/2), [3](https://github.com/alexarchambault/scalacheck-shapeless/tree/prioritycrash/3)), ... or not ([4](https://github.com/alexarchambault/scalacheck-shapeless/tree/prioritycrash/4)). In the output code, some lazy implicits in the generated `Lazy` instance reference non-existing lazy implicits (by "lazy implicits", I mean the entries from `dict`, that ultimately land in the output). I think the problem originates in implicit instances being removed (https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/lazy.scala#L180), although the implicit lookup just above (https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/lazy.scala#L165) added lazy implicits referencing to the removed one.
I'm only running into this with my additions, although it seems this could have occurred before... (maybe I just wasn't pushing `Lazy` as far as that before.)

alexarchambault:
Just a note on `Implicit`: it generalizes to coproducts and types having a `Generic` what `Values` does for `HList`s, that is filling a type with implicit values. But `Values` doesn't sound right for coproducts, which only have one value in their results.

alexarchambault:
Something with a reworked version of this (among others) will pop up any time soon [here](https://github.com/alexarchambault/lookup). Watch for it!

PR #428, opened 2015-07-11
Added documentation for some user-visible classes in generic.scala

xoltar:
I have some docs for other files as well, submitting PRs on file-by-file basis to keep reviews manageable.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -23,27 +23,191 @@ import scala.reflect.macros.{ blackbox, whitebox }
 
 import ops.{ hlist, coproduct }
 
+ /** Represents the ability to convert from [[T]], a concrete type (e.g. a case class)
+  * to a generic ( [[HList]] / [[Coproduct]]} based) representation of the type.
+  *
+  * For example:
+  * {{{
+  * scala> sealed trait Animal
+
+  * defined trait Animal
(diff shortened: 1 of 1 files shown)


milessabin:
Many thanks for that! Merging ... :-)

milessabin:
Be sure to add yourself to the contributors file in the next PR ...

milessabin:
One thing cropped up on the [PR build](https://travis-ci.org/milessabin/shapeless/jobs/70712690) ... various names in the doc were accidentally picked up as substitutions because they contained '$'s and there were a couple of other places where names were used in contexts which scaladoc expected to be able to find definitions but couldn't, typically because of lack of prefix information. Running the doc task from the SBT REPL is the quickest way to catch these things.

I've fixed the problems [here](https://github.com/milessabin/shapeless/commit/8a984c7941fe5f0c05fbc645c68597084ed69115) and also simplified the various `Repr` types ... these aren't the way they would be rendered on the REPL, but they are a lot easier to read.

xoltar:
Good to know, thanks! Will run the doc task next time.

fthomas:
Ideally the Travis build should catch such errors by running both the `test` and `doc` tasks. In Cats we added a command alias for validating PRs (https://github.com/non/cats/blob/aec71b97dfbe5b3e1c8e909805fb45b4451504b3/build.sbt#L219) which is used for Travis builds and which can be used by contributors before they send PRs.

milessabin:
Agreed. PR? ;-)

Issue #374, opened 2015-04-26
Macro generated classes should be prefixed "anon$" to get on with value classes


milessabin:
See discussion with @retronym [here](https://gitter.im/scala/scala?at=553439218c86646906dd773d). Probably too risky to do for 2.2.0 from a binary compatibility point of view.

This is a fairly simple thing for someone who wants to start digging in to macros to take a look at.

milessabin:
Fixed in d8c43fbdc8fb559115ede111e3877c89229069ee.

PR #429, opened 2015-07-12
Weaken type of implicit Generic macro to Generic.Aux[T, Repr]

retronym:
The result of the generic macro is currently:

   { class anon$1 extends Generic[T]; new anon$2} : anon$2

The type of the block has an ill-scoped reference to the anonymous
class, which ends up causing problems down the line if we summon
`implicitly[Lazy[Generic[T]]]`: the lazy macro uses the exact type
of the nested materialized `Generic`, but then retypechecks the
block that defines the `anon$1` symbol. This means that the type
doesn't point to the new symbol for the anonymous class, which is
enough to crash the new `GenBCode` of the compiler from Scala 2.11.3
onwards.

    % scalac-hash v2.11.3 -Xprint:typer -nc -Ybackend:GenBCode \
                        -classpath core/target/scala-2.12/classes \
                        -e 'import shapeless._; implicitly[Lazy[Generic[Nil.type]]]
    assertion failed: ClassBType.info not yet assigned: LMain$$anon$1$fresh$macro$3$1;

I've noted a few more details in the comments of SI-9392.

My experience with macros would lead me to rewrite `Lazy` to to avoid
the retypechecking, although this is pretty tedious to get right in
its own right. (I used this apporach in scala-async.)

This commit doesn't attempt this change, but instead weakens the type
of the `Generic` macro to hide the local anonymous class.

We will also try to tweak the compiler to be tolerant of these
incoherent trees in the backend.

After this change, the test case above compiles successfully.
It also no longer incurs warnings under the diagnostic mode of
the compiler, `-Ycheck:typer`.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -616,7 +616,7 @@ class GenericMacros(val c: whitebox.Context) extends CaseClassMacros {
         def to(p: $tpe): Repr = (p match { case ..$toCases }).asInstanceOf[Repr]
         def from(p: Repr): $tpe = p match { case ..$fromCases }
       }
-      new $clsName()
+      new $clsName(): _root_.shapeless.Generic.Aux[$tpe, ${reprTypTree(tpe)}]
     """
   }
 


retronym:
@milessabin I didn't look for other instances of the `{class anon$; new anon$}` pattern in macros within shapeless. If there are others, I'd suggest to uniformly ascribe the types in the same way.

milessabin:
Thanks for that! Making that change and similar in a few other places fixes all the problem with the 2.12.0-M2 build.

I'm wondering there's a relationship with #374 though. Would the leaking references go away if instead these were all instantiated as anonymous classes, ie.,

```scala
new Generic[T] {
  ...
}
```

rather than,

```scala
final class $clsName extends Generic[T] {
  ...
}
new $clsName
```

retronym:
I don't believe those variations would make a difference here. You can experiment with this directly using the toolbox compiler: https://issues.scala-lang.org/browse/SI-9392?focusedCommentId=72837&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-72837

milessabin:
Thanks again ... can you point me in the direction of this logic in the compiler sources?

retronym:
Look at callers of packedType in Typers.scala.

milessabin:
Incorporated in https://github.com/milessabin/shapeless/commit/891fd076f3b262d0debc0c17f0205ea4a4aba931 and https://github.com/milessabin/shapeless/commit/78def231b91b45786aa9676fc1398b868056ac04.

PR #434, opened 2015-07-23
add Map toRecord method

tbfenet:
Add toRecord operation that fills a record with values from a map. Supports nested records

### DIFF
--- a/core/src/main/scala/shapeless/ops/maps.scala
+++ b/core/src/main/scala/shapeless/ops/maps.scala
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2011-15 Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shapeless.ops
+
+import shapeless._
+import shapeless.labelled.FieldType
+
+object maps {
+
+  /**
+   * Type class supporting type safe conversion of Map to Records.
+   *
+   *
+   */
+  trait FromMap[Out <: HList] extends Serializable {
+    def apply(l: Map[Any, Any]): Option[Out]
+  }
+
+  /**
+   * `FromMap` type class instances.
+   *
+   *
+   */
+  object FromMap {
+
+    def apply[Out <: HList](implicit from: FromMap[Out]) = from
+
+    implicit def hnilFromMap[T] = new FromMap[HNil] {
+
+      override def apply(l: Map[Any, Any]) = Some(HNil)
+    }
+
+    import labelled._
+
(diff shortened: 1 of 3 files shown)


milessabin:
Sorry it's taken me so long to give this proper attention.

It looks really interesting and useful ... I can some interesting directions it could go in.

One observation is that shapeless record types aren't the most convenient things to work with directly ... typically they would arise from the use of a shapeless `Generic` for an ADT. That suggests that it might also be worthwhile to consider handling shapeless unions, ie. labelled `Coproduct`s as well as records so that you can handle not just case classes, but sealed families of case classes.

If you went down that route, then I think what you have would be a form of serialization library where we're serializing to/from very loosely typed Scala (or even Java) data structures, but in a boilerplate-free and type safe way. I think that something like that would potentially be very useful and widely applicable.

I also think that similarly to other shapeless-based serialization libraries, like Scodec, argonaut-shapeless and spray-json-shapeless, it would be better as a standalone project than included directly in shapeless itself. I'm currently trying to work out how to extract out an essential core from shapeless and will be moving things like the lenses, scrap your boilerplate and zippers to separate projects and I would recommend the same here.

Let me know what you think ...

tbfenet:
I just found no convenient way to populate records from loosely typed scala. Unlike plain HList where we  have toHList.  I am using Records to define the mapping of case classes in spark to data stored in Hive tables and this main issue I faced

milessabin:
The analogy with `toHList` is quite persuasive. I've taken another look at the PR and I'm still not completely comfortable with it though. The nesting of `Map[Any, Any]` and use of `Either` rather than `Option` makes it feel like this is more of an application-specific thing rather than a low level primitive.

I'd be happy with something along these lines as an example, but I think that to include it in the main library I'd like to see if some simpler primitive can be extracted out of it. For instance do you think you could pull out a one-layer part of this (ie. convert a map to an option of record, but without recursing through nested maps) and then reconstruct the full nested conversion in terms of that?

tbfenet:
The use of Either comes from experience when using toHList where I have had to resort to using a debugger to work out which field did not match. Here we are at the interface between runtime and compile time so I think the need to be some sort of hook for runtime diagnostics.  We could populate record from function instead of a Map? This might be a better primitive for different serialisation libraries.

milessabin:
> We could populate record from function instead of a Map? This might be a better primitive for different serialisation libraries.

Could you elaborate on that a little? I'm used to the idea of a `Map[K, V]` corresponding to a function `K => V` but in that case we lose the ability to enumerate the keys, so I'm not quite sure how the mapping to a record might work.

tbfenet:
It is driven from the record side. You use the the function to resolve all keys in record as it currently does using the Map.

milessabin:
I understand your point of view, but I'm still not convinced. I can see a few ways to move this forward,

+ You could follow my suggestion above and make this a one-layer `Map` -> record operation (ie. no recursion through embedded maps) and switch to `Option` for consistency with `toHList`. I'd be happy to merge that into shapeless core. I'd also be happy to have an example (in the examples subproject) using this as a primitive to build the recursive version.

+ You could make the whole thing an example (in the examples subproject) as a specific idiom someone using Spark might need. This way, other people who have the same use case can benefit from the example, but it won't create confusion with other shapeless code.

+ You could create a new library that is specialized to the kind of problem you needed to solve (e.g. using shapeless + Spark), or add it to another library along those lines, maybe [frameless](https://github.com/adelbertc/frameless)?.

I hope you'll go for the first option, because I think that a one-layer `toRecord` operation that's consistent with the rest of shapeless would be a very welcome addition :-)

milessabin:
Oh wow! I didn't see that you'd updated this PR with that last commit :-(

This looks good to merge now ... there are a couple of layout issues which can be fixed up later, but nothing that prevents this going in at long last.

My apologies for not noticing your update :-(

PR #435, opened 2015-07-27
Lazy extensions

alexarchambault:
Opening this to discuss the developments in the [lazyextensions branch](https://github.com/milessabin/shapeless/tree/topic/lazyextensions).

It adds:
- a `Strict` type class, that circumvents implicit divergences like `Lazy`, without the overhead of an anonymous function when instantiated (hence an extra class),
- the possibility to extend `Lazy` with extensions, able to change the current derivation and its state, and keep their own state along with it,
- a `Priority` type class and a Lazy extension, to be used wrapped in an implicit `Strict` or `Lazy`, like
```scala
implicit def mkTC[T]
 (implicit
   priority: Strict[Priority[TC[T], Other[T]]]
 ): TC[T] = priority.value.fold(identity)(_.mkTC)
```
which looks up for a `TC[T]`, else for a `Other[T]`, taking into account that a `TC[T]` is also provided through itself (and ignoring it in the first `TC[T]` lookup - thus looking for orphan `TC[T]` only),
- various inlining of implicits in the `Strict` and `Lazy` output, for less overhead (see https://github.com/milessabin/shapeless/commit/1add5ffd55a7b04aef23022a9bd5089ed694436d).

Also things that proved not to be useful, that I'd like to remove carefully, like
- some changes in `Generic`,
- using the state monad to manage the derivation state,
- keep track of implicit lookup that failed, not to look them once more,
- a lazy extension called `Scoped`, that keeps the implicit dict from a lookup to the next (-> more caching).

I'm somehow benchmarking this versus the macro-based derivation of [upickle](https://github.com/lihaoyi/upickle-pprint), with [a modified version of it](https://github.com/alexarchambault/upickle-pprint). The better I get is a bit less than a 3 times longer compile time for its tests :-( (from initially... much more than that).

### DIFF
--- a/core/src/main/scala/shapeless/lazy.scala
+++ b/core/src/main/scala/shapeless/lazy.scala
@@ -21,6 +21,91 @@ import scala.language.experimental.macros
 import scala.collection.immutable.ListMap
 import scala.reflect.macros.whitebox
 
+/**
+ * Wraps a lazily computed value. Also circumvents cycles during implicit search, or wrong implicit divergences
+ * as illustrated below, and holds the corresponding implicit value lazily.
+ *
+ * The following implicit search sometimes fails to compile, because of a wrongly reported implicit divergence,
+ * {{{
+ *   case class ListCC(list: List[CC])
+ *   case class CC(i: Int, s: String)
+ *
+ *   trait TC[T]
+ *
+ *   object TC {
+ *     implicit def intTC: TC[Int] = ???
+ *     implicit def stringTC: TC[String] = ???
+ *     implicit def listTC[T](implicit underlying: TC[T]): TC[List[T]] = ???
+ *
+ *     implicit def genericTC[F, G](implicit
+ *       gen: Generic.Aux[F, G],
+ *       underlying: TC[G]
+ *     ): TC[F] = ???
+ *
+ *     implicit def hnilTC: TC[HNil] = ???
+ *
+ *     implicit def hconsTC[H, T <: HList](implicit
+ *       headTC: TC[H],
+ *       tailTC: TC[T]
+ *     ): TC[H :: T] = ???
+ *   }
+ *
+ *   implicitly[TC[CC]] // fails with: diverging implicit expansion for type TC[CC]
+ * }}}
+ *
+ * This wrongly reported implicit divergence can be circumvented by wrapping some of the implicit values in
+ * `Lazy`,
+ * {{{
+ *   case class ListCC(list: List[CC])
+ *   case class CC(i: Int, s: String)
+ *
+ *   trait TC[T]
+ *
+ *   object TC {
+ *     implicit def listTC[T](implicit underlying: TC[T]): TC[List[T]] = ???
+ *
+ *     implicit def genericTC[F, G](implicit
+ *       gen: Generic.Aux[F, G],
+ *       underlying: Lazy[TC[G]] // wrapped in Lazy
+ *     ): TC[F] = ???
+ *
+ *     implicit def hnilTC: TC[HNil] = ???
+ *
+ *     implicit def hconsTC[H, T <: HList](implicit
+ *       headTC: Lazy[TC[H]], // wrapped in Lazy
+ *       tailTC: TC[T]
+ *     ): TC[H :: T] = ???
+ *   }
+ *
+ *   implicitly[TC[CC]]
+ * }}}
+ *
+ * When looking for an implicit `Lazy[TC[T]]`, the `Lazy.mkLazy` macro will itself trigger the implicit search
+ * for a `TC[T]`. If this search itself triggers searches for types wrapped in `Lazy`, these will be done
+ * only once, their result put in a `lazy val`, and a reference to this `lazy val` will be returned as the corresponding
+ * value. It will then wrap all the resulting values together, and return a reference to the first one.
+ *
+ * E.g. with the above example definitions, when looking up for an implicit `TC[CC]`, the returned tree roughly looks
+ * like
+ * {{{
+ *   TC.genericTC(
+ *     Generic[CC], // actually, the tree returned by Generic.materialize, not written here for the sake of brevity
+ *     Lazy {
+ *       lazy val impl1: TC[List[CC] :: HNil] = TC.hconsTC(
+ *         Lazy(impl2),
+ *         TC.hnilTC
+ *       )
+ *       lazy val impl2: TC[List[CC]] = TC.listTC(TC.genericTC(
+ *         Generic[CC], // actually, the tree returned by Generic.materialize
+ *         Lazy(impl1)  // cycles to the initial TC[List[CC] :: HNil]
+ *       ))
+ *
+ *       impl1
+ *     }
+ *   )
+ * }}}
+ *
+ */
 trait Lazy[+T] extends Serializable {
   val value: T
 
--- a/core/src/main/scala/shapeless/lazy.scala
+++ b/core/src/main/scala/shapeless/lazy.scala
@@ -131,6 +131,33 @@ object Lazy {
   def values[T <: HList](implicit lv: Lazy[Values[T]]): T = lv.value.values
 
   implicit def mkLazy[I]: Lazy[I] = macro LazyMacros.mkLazyImpl[I]
+
+
+  /**
+   * Wraps a lazily computed value, and circumvents implicit cycles / wrong implicit divergences, like `Lazy` does.
+   * Assumes that all the contexts it is called in are equivalent, and caches its calculations.
+   *
+   * Unlike `Lazy`, `Lazy.Global` assumes that all the contexts in which values wrapped in it are looked up, are all
+   * equivalent. This allows to cache the generated trees upon first call, and simply return the cached trees
+   * during subsequent calls.
+   *
+   * Shares its cache with `Strict.Cache`.
+   */
+  trait Global[+T] extends Lazy[T] {
+    override def map[U](f: T => U): Global[U] = Global { f(value) }
+    def flatMap[U](f: T => Global[U]): Global[U] = Global { f(value).value }
+  }
+
+  object Global {
+    implicit def apply[T](t: => T): Global[T] =
+      new Global[T] {
+        val value = t
+      }
+
+    def unapply[T](lt: Global[T]): Option[T] = Some(lt.value)
+
+    implicit def mkGlobal[I]: Global[I] = macro LazyMacros.mkGlobalLazyImpl[I]
+  }
 }
 
 object lazily {
@@ -161,6 +188,33 @@ object Strict {
   def unapply[T](lt: Strict[T]): Option[T] = Some(lt.value)
 
   implicit def mkStrict[I]: Strict[I] = macro LazyMacros.mkStrictImpl[I]
+
+
+  /**
+   * Wraps an eagerly computed value, and circumvents wrong implicit divergences, like `Strict`.
+   * Assumes all the contexts it is called are equivalent, and caches its calculations.
+   *
+   * Unlike `Strict`, `Strict.Global` assumes that all the contexts in which values wrapped in it are looked up, are all
+   * equivalent. This allows to cache the generated trees upon first call, and simply return the cached trees
+   * during subsequent calls.
+   *
+   * Shares its cache with `Lazy.Cache`.
+   */
+  trait Global[+T] extends Strict[T] {
+    override def map[U](f: T => U): Global[U] = Global { f(value) }
+    def flatMap[U](f: T => Global[U]): Global[U] = Global { f(value).value }
+  }
+
+  object Global {
+    implicit def apply[T](t: T): Global[T] =
+      new Global[T] {
+        val value = t
+      }
+
+    def unapply[T](lt: Global[T]): Option[T] = Some(lt.value)
+
+    implicit def mkGlobal[I]: Global[I] = macro LazyMacros.mkGlobalStrictImpl[I]
+  }
 }
 
 class LazyMacros(val c: whitebox.Context) {
@@ -168,22 +222,41 @@ class LazyMacros(val c: whitebox.Context) {
   import c.ImplicitCandidate
 
   def mkLazyImpl[I](implicit iTag: WeakTypeTag[I]): Tree =
-    mkImpl[I](strict = false)
+    mkImpl[I](
+      (tree, actualType) => q"_root_.shapeless.Lazy.apply[$actualType]($tree)",
+      q"null.asInstanceOf[_root_.shapeless.Lazy[Nothing]]",
+      None
+    )
 
   def mkStrictImpl[I](implicit iTag: WeakTypeTag[I]): Tree =
-    mkImpl[I](strict = true)
+    mkImpl[I](
+      (tree, actualType) => q"_root_.shapeless.Strict.apply[$actualType]($tree)",
+      q"null.asInstanceOf[_root_.shapeless.Strict[Nothing]]",
+      None
+    )
 
-  def mkImpl[I](strict: Boolean)(implicit iTag: WeakTypeTag[I]): Tree = {
+  def mkGlobalLazyImpl[I](implicit iTag: WeakTypeTag[I]): Tree =
+    mkImpl[I](
+      (tree, actualType) => q"_root_.shapeless.Lazy.Global.apply[$actualType]($tree)",
+      q"null.asInstanceOf[_root_.shapeless.Lazy.Global[Nothing]]",
+      Some("")
+    )
+
+  def mkGlobalStrictImpl[I](implicit iTag: WeakTypeTag[I]): Tree =
+    mkImpl[I](
+      (tree, actualType) => q"_root_.shapeless.Strict.Global.apply[$actualType]($tree)",
+      q"null.asInstanceOf[_root_.shapeless.Strict.Global[Nothing]]",
+      Some("")
+    )
+
+  def mkImpl[I](mkInst: (c.Tree, c.Type) => c.Tree, nullInst: => c.Tree, cacheOpt: Option[String])(implicit iTag: WeakTypeTag[I]): Tree = {
     (c.openImplicits.headOption, iTag.tpe.dealias) match {
       case (Some(ImplicitCandidate(_, _, TypeRef(_, _, List(tpe)), _)), _) =>
-        LazyMacros.deriveInstance(c)(tpe.map(_.dealias), strict)
+        LazyMacros.deriveInstance(c)(tpe.map(_.dealias), mkInst, cacheOpt)
       case (None, tpe) if tpe.typeSymbol.isParameter =>       // Workaround for presentation compiler
-        if (strict)
-          q"null.asInstanceOf[_root_.shapeless.Strict[Nothing]]"
-        else
-          q"null.asInstanceOf[_root_.shapeless.Lazy[Nothing]]"
+        nullInst
       case (None, tpe) =>                                     // Non-implicit invocation
-        LazyMacros.deriveInstance(c)(tpe, strict)
+        LazyMacros.deriveInstance(c)(tpe, mkInst, cacheOpt)
       case _ =>
         c.abort(c.enclosingPosition, s"Bad Lazy materialization ${c.openImplicits.head}")
     }
@@ -191,27 +264,50 @@ class LazyMacros(val c: whitebox.Context) {
 }
 
 object LazyMacros {
+  var caches = Map.empty[String, List[(Any, Any)]]
+
   var dcRef: Option[DerivationContext] = None
 
-  def deriveInstance(c: whitebox.Context)(tpe: c.Type, strict: Boolean): c.Tree = {
-    val (dc, root) =
-      dcRef match {
-        case None =>
-          val dc = DerivationContext(c)
-          dcRef = Some(dc)
-          (dc, true)
-        case Some(dc) =>
-          (DerivationContext.establish(dc, c), false)
+  def deriveInstance(c: whitebox.Context)(tpe: c.Type, mkInst: (c.Tree, c.Type) => c.Tree, cacheOpt: Option[String]): c.Tree = {
+    val cached =
+      if (dcRef.isEmpty)
+        cacheOpt.flatMap(caches.get).flatMap { cache =>
+          cache.collectFirst {
+            case (tpe0, v) if tpe0.asInstanceOf[c.Type] =:= tpe =>
+              v.asInstanceOf[c.Tree]
+          }
+        }
+      else
+        None
+
+    def derive: c.Tree = {
+      val (dc, root) =
+        dcRef match {
+          case None =>
+            val dc = DerivationContext(c)
+            dcRef = Some(dc)
+            (dc, true)
+          case Some(dc) =>
+            (DerivationContext.establish(dc, c), false)
+        }
+
+      if (root)
+        // Sometimes corrupted, and slows things too
+        c.universe.asInstanceOf[scala.tools.nsc.Global].analyzer.resetImplicits()
+
+      try {
+        dc.State.deriveInstance(tpe, root, mkInst)
+      } finally {
+        if(root) dcRef = None
       }
+    }
 
-    if (root)
-      // Sometimes corrupted, and slows things too
-      c.universe.asInstanceOf[scala.tools.nsc.Global].analyzer.resetImplicits()
-
-    try {
-      dc.State.deriveInstance(tpe, root, strict)
-    } finally {
-      if(root) dcRef = None
+    cached.getOrElse{
+      val res = derive
+      if (dcRef.isEmpty)
+        for (cache <- cacheOpt)
+          LazyMacros.caches += cache -> ((tpe, res) :: LazyMacros.caches.getOrElse(cache, Nil))
+      res
     }
   }
 }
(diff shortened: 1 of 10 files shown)


alexarchambault:
I'd like to remove what turned out unnecessary above, then this would be fine for a merge for me.

alexarchambault:
Almost done or ready for review. What's missing:
- scaladoc
- merge some commits with previous ones
- scala 2.10 port (just a matter of merging lazy.scala in the scala-2.10.x branch)
- tests of the cached versions of Lazy/Strict (maybe with a stateful implicit macro, counting the number of times it gets called?)

milessabin:
Awesome!

You've been hinting at compile time speedups ... do you have any benchmarks of that?

alexarchambault:
Yes, with the tests of https://github.com/alexarchambault/upickle-pprint. I'm trying to disentangle the effects of the various changes I made (both here and [there](https://github.com/alexarchambault/upickle-pprint)).

alexarchambault:
Quick update: while assessing the effects of each change differently than my initial attempts, I found that one of them (inlining) sometimes slows things. I may remove it, I'm investigating further...

Also, the biggest speed ups seem to come from some "best practices" on the deriving library side, more than from changes here. I'll come up with clear benchmarks soon.

milessabin:
Would you say this is ready for review now?

alexarchambault:
@milessabin Not yet, I'm just ensuring that all its changes are really necessary, with various benchmarks (that I'll put online right after). It's almost done, but for inlining in Lazy results. I hope to have them, and clean the PR, later today.

alexarchambault:
I'd say its ready for review now :-)

Some benchmarks about the various additions of this PR are available [here](https://github.com/alexarchambault/auto-type-class-benchmark).

Only thing that does not speed up things is `Implicit` (https://github.com/milessabin/shapeless/commit/b9d46fcd1e6108c351ef9552e79031fe65d6c29d). I still find it nicer (and with a broader scope) than the current `Values` though, so I left it here for now.

Some minor points that I would like to change (but which don't prevent reviewing):
- remove unnecessary `0` suffixes in some variable / method argument names,
- simplify some unnecessarily complex variable / method argument names (e.g. `instTpe` -> `tpe`).

The last commit would have to be stripped before a merge (and either update the MIMA config, or bump version to 2.3.0-SNAPSHOT and temporarily disable MIMA).

.travis.yml
```diff
@@ -2,7 +2,6 @@ language: scala
 
 scala:
 - 2.11.7
-- 2.12.0-M2
```
milessabin: What was the problem with 2.12.0-M2?

core/src/test/scala/shapeless/serialization.scala
```diff
@@ -1065,7 +1065,9 @@ class SerializationTests {
 
   @Test
   def testFunctor {
-    assertSerializable(Functor[Some])
+    val f = Functor[Some]
+    f.map(Some(2))(_.toString)
+    assertSerializable(f)
     assertSerializable(Functor[Option])
```
milessabin: What's going on here?

core/src/main/scala/shapeless/lazy.scala
```diff
@@ -46,48 +47,187 @@ object Lazy {
...
+    implicit def mkCached[I]: Cached[I] = macro LazyMacros.mkCachedLazyImpl[I]
+  }
+
 }
 
 object lazily {
   def apply[T](implicit lv: Lazy[T]): T = lv.value
 }
 
+trait Strict[+T] extends Serializable {
```
milessabin: I could do with a bit of explanation of what's going on here ... you seem to be using `Strict` in situations potentially involving cycles where I would expect laziness to be essential ... how have you avoid that?

core/src/main/scala/shapeless/implicit.scala
```diff
@@ -0,0 +1,76 @@
...
+ * look up for implicits `A` and `B`. If both are found, it succeeds and the implicit
+ * `A` and `B` are available via `value`. Same thing for longer HLists.
+ *
+ * Looking for an `Implicit[A :+: B :+: CNil]` will look up for an implicit `A` first,
+ * else for an implicit `B` if it fails. If either one of them is found, it is
+ * available wrapped in a `A :+: B :+: CNil`, via `value`. Same thing with longer coproducts.
+ *
+ * One can also look for types having a `Generic` instance. Their generic representation
+ * will be looked up as above, then converted to the right type with `Generic`.
+ *
```
milessabin: Is there an example of looking for types with a `Generic` instance? I'm wondering if this overlaps with things like `WitnessWith` or the "typelevel-CPS" pattern in `Generic1` and `IsH/CCons1`?

milessabin:
This is looking pretty awesome :-)

.travis.yml
alexarchambault: Some tests didn't pass anymore in `SerializationTests`, I just pushed some workarounds (and reverted this particular change).

core/src/test/scala/shapeless/serialization.scala
alexarchambault: A safety check, I changed it a bit along with the workarounds for 2.12. I don't remember if it still has to be here, or if it's independent of the other devs of this PR.

The problem is that the `Functor[Some]` instance found implicitly has some lazily initialized fields, in particular some `Generic1` / `IsHCons1` / `IsCCons1` (the first commit of this PR makes them serializable). Thus, serializability can change once these are made concrete. The call to `map` makes these fields concrete, and tests that.

core/src/main/scala/shapeless/lazy.scala
alexarchambault: I added `Strict` and tend to use it when possible usually compiles a bit faster than with `Lazy`. I tend to incriminate the anonymous function (hence an additional `.class` file) that has to be created when calling the `apply` method of the `Lazy` companion. (In the v0 of my benchmark, https://github.com/alexarchambault/upickle-pprint/commit/2a63bef3a822dba4debf9726d7c515172698254f with https://github.com/alexarchambault/shapeless/commit/0fa6a9d4de45e656624668be3a8a8993df38b281, approx. ~12000 classes were generated, partly because of that, compared to less than 3000 when using macros.)

The thing with cycles is that they only need to be broken once! E.g., these typically appear in cases like
```scala
case class CC(l: List[CC])
```
or
```scala
sealed trait Tree
case class Leaf(value: Int) extends Tree
case class Node(children: List[Tree]) extends Tree
```

If one looks for a `TC[CC]`, the returned `Lazy[TC[CC]]` should look like
```scala
class anon {
  lazy val impl4: TC[HNil] = hnilTC
  lazy val impl3: TC[List[CC]] = listTC(impl1) // cycles on impl1
  lazy val impl2: TC[List[CC] :: HNil] = hconsTC(Lazy(impl3), Lazy(impl4))
  lazy val impl1: TC[CC] = genericTC(
    Generic[CC], // actually output of the Generic.materialize macro
    Lazy(impl2)
  )
}
new anon .impl1
```

So the cycle goes like
* `TC[CC]`, returned by
* `genericTC`, needs a
* `TC[List[CC] :: HNil]`, returned by
* `hconsTC`, needs a
* `TC[List[CC]]`, returned by
* `listTC`, needs a
* `TC[CC]`, hence a cycle.
Only one of the implicit instances of the involved types needs to be wrapped in a `Lazy` to break the cycle, the others can just be wrapped in a `Strict` (or not wrapped at all if implicit search doesn't diverge).

core/src/main/scala/shapeless/implicit.scala
alexarchambault: I removed `Implicit` from this PR, and put it in a separate branch ([topic/implicit](https://github.com/alexarchambault/shapeless/tree/topic/implicit)). I wasn't using it anymore in the benchmark, it was only slowing things. I was using it to prioritize implicits (like `Implicit[MkTupleTC[T] :+: MkProductTC[T] :+: CNil]`), but standard prioritization is faster.

alexarchambault:
@milessabin I'd recomment inspecting at the changes of this PR through https://github.com/milessabin/shapeless/tree/topic/lazyextensions, looking at the commits one-by-one (and the individual commit messages provide some explanations too).

alexarchambault:
By the way, I ported that to 2.10 too (https://github.com/milessabin/shapeless/tree/topic/lazyextensions-2.10.x). The port is not up to date anymore because of various rebase / changes, but I'll update it once the developments here are fine.

core/src/test/scala/shapeless/serialization.scala
milessabin: Surely the attempt to serialize would force any lazy fields?

core/src/main/scala/shapeless/lazy.scala
milessabin: Are you assuming that we only enter the cycle via the root trait? Suppose we didn't?

It isn't necessarily a problem if we say that there are some additional constraints on type class derivation which uses `Strict`, but it'd be good to know if there's a performance/footprint tradeoff against generality here.

core/src/test/scala/shapeless/serialization.scala
```diff
@@ -228,7 +234,7 @@ object SerializationTestDefns {
 
 }
 
-class SerializationTests {
+class SerializationTests extends Serializable {
   import SerializationTestDefns._
```
milessabin: This isn't good ... why are we capturing an outer reference now?

core/src/test/scala/shapeless/serialization.scala
```diff
@@ -937,11 +943,11 @@ class SerializationTests {
   def testLazy {
     assertSerializable(Lazy(23))
 
-    assertSerializable(implicitly[Lazy[Generic[Wibble]]])
-    assertSerializable(implicitly[Lazy[Generic1[Box, TC1]]])
+    assertSerializableBeforeAfter(implicitly[Lazy[Generic[Wibble]]])(_.value)
+    assertSerializableBeforeAfter(implicitly[Lazy[Generic1[Box, TC1]]])(_.value)
```
milessabin: I don't mind adding new tests, but I'm uncomfortable removing existing ones here ... can you explain why this is necessary? If it relates to changes in serialization behaviour due to forcing lazy vals then I think I'd like a deeper understanding of why that's a problem.

project/Build.scala
```diff
@@ -43,7 +43,7 @@ object ShapelessBuild extends Build {
       publish := (),
       publishLocal := (),
 
-      addCommandAlias("validate", ";test;mima-report-binary-issues;doc")
+      addCommandAlias("validate", ";test;doc")
```
milessabin: This will need to be reinstated before we merge.

milessabin:
At this point I think I have only two concerns, one minor, the other hopefully minor.

I think that the introduction of `Strict` as a name here to contrast with `Lazy` is revealing that neither name is particularly good. What both of these are doing is allowing implicit resolution to continue to a fixed point. Laziness at the value level is necessary somewhere along the line, but the key thing is the type level fixed point.

I think I'd now also prefer to point people at Cats's `Later` if they want something with the value level properties that shapeless's `Lazy` currently has.

So I'd be inclined to come up with a new pair of names, alias `Lazy` to the appropriate one of them and deprecate `Lazy`.

That's the minor issue: it's "just" about names and deprecation ;-)

The hopefully minor issue is the changes in the serialization tests. If there's a change in behaviour here then we need to deal with it. If there isn't then I think we have a pretty bad bug in shapeless as it stands which needs to be fixed independently of this PR.

core/src/main/scala/shapeless/lazy.scala
alexarchambault: In my tests, I cut it in the `genericTC` implicit argument for the `TC` of an hlist, `TC[List[CC] :: HNil]` here, so not the "root". As long as the cycle is cut at one point, no matter which one, it should be safe. No need to enforce additional constraints. (Unless I'm missing something here, which I don't think I do.)

core/src/test/scala/shapeless/serialization.scala
alexarchambault: This is only necessary in 2.12, not in 2.11... It seems that some materialized `Lazy` instances capture an outer reference... I have no idea why.

core/src/test/scala/shapeless/serialization.scala
alexarchambault: These are only additional tests. `assertSerializableBeforeAfter` tests that the first argument it is given is serializable upfront (like the former `assertSerializable` did), then it calls its second argument on the first one, and ensure the it is still serializable after that. What was tested before is still tested here.

core/src/test/scala/shapeless/serialization.scala
milessabin: I fixed similar things in 2.2.5 here: https://github.com/milessabin/shapeless/commit/78def231b91b45786aa9676fc1398b868056ac04. Is it possible that you've lost this in the update?

core/src/test/scala/shapeless/serialization.scala
milessabin: I'm still puzzled ... this is testing serializability before and after forcing lazy vals? But surely the first  serializing is itself forcing any lazy vals? I guess I'm suspicious because I don't fully understand how there could be a difference here.

core/src/main/scala/shapeless/lazy.scala
milessabin: Maybe add a test for that? Ie. create a cycle with a single `Lazy` and multiple `Strict`'s and verify that there are no issues wherever you start from?

core/src/test/scala/shapeless/serialization.scala
alexarchambault: Serialization doesn't force lazy vals. It just serializes the non-null fields. So if the lazy vals aren't materialized yet, their value won't be tested for serializability. I don't remember the exact details of how lazy vals are implemented, but I guess a closure get serialized in this case. A closure able to calculate the value if asked so.

milessabin: Ouch. Yes, I guess that makes sense. I still think that we shouldn't be capturing that outer reference though.

alexarchambault: Agreed. I'll try to understand what's going on.

core/src/test/scala/shapeless/serialization.scala
alexarchambault: Some of its changes got lost, yes (around [this line](https://github.com/milessabin/shapeless/commit/78def231b91b45786aa9676fc1398b868056ac04#diff-ffd1855210ab80b59329091ae5193e48R194)), but don't fix all of the issues here. I still have to comment out [two tests](https://github.com/milessabin/shapeless/commit/720164d1658354b32ec7725243b642665536d900#diff-21cd9467b2b35a579fc7910defb1d163R1085) (reverting the changes I made to them doesn't solve the issue either).

core/src/main/scala/shapeless/lazy.scala
alexarchambault: Done in https://github.com/milessabin/shapeless/commit/a609a7e25ea561f494f76ecbd54e7c73a9b2f746

alexarchambault:
Just pushed new changes, adding extra tests (https://github.com/milessabin/shapeless/commit/d34fc337932fbfd31d4ac86e4ab926eed957589d, https://github.com/milessabin/shapeless/commit/a609a7e25ea561f494f76ecbd54e7c73a9b2f746), some scaladoc (https://github.com/milessabin/shapeless/commit/fd45288f468cf99fa25316bc583afd8c2780cded) in particular.

Only remaining problems (correct me if I'm wrong):
- outer reference capturing in `SerializationTests.testShow` (failing tests commented out for now) in 2.12,
- renaming?

About the outer reference capturing, the inlining in the generated trees seems to be the problem. If one comments out `.optimize` in [this line](https://github.com/milessabin/shapeless/commit/8ea3907e9a3fab50d1cb18777f406602366bedd3#diff-ffd1855210ab80b59329091ae5193e48R547), the tests pass. I don't really know why the transformation that inlining would make the result capture an outer reference (and only so in 2.12), but I haven't looked closely at the trees generated in the failing tests yet.

core/src/main/scala/shapeless/lazy.scala
milessabin: Good stuff :-)

Bit baffled by the need for that reflective access language import though ...

milessabin:
WRT `Global` ... how confident are you that the lifecycle of those caches is well behaved. Previously the lifetime of a cached was bounded by the nesting of implicit resolutions through a `Lazy` type constructor: there would always have been a well defined entry point with an empty cache and a well-defined exit point where the cache could be discarded. Presumably that's no longer the case?

So I guess that the potential trouble spots would be with incremental compilation and the presentation compiler. Do you have a feel for how this behaves with SBT `~testQuick` and ENSIME? /cc @fommil

fommil:
if you want a good testbed, try updating the `sexp.scala` example and open it up in ENSIME. Or migrate spray-json-shapeless to use `Strict` instead of `Lazy` and see how the tests behave.

alexarchambault:
I already tested that a bit with the comprehensive test suite of upickle, with an [ad-hoc port](https://github.com/alexarchambault/upickle-pprint) of it to shapeless with the changes of this PR. See [this project](https://github.com/alexarchambault/auto-type-class-benchmark) for an analysis of it (with comparison to upickle original macros - at then end, the shapeless version still makes the tests take 2.5x longer to compile).

I also ported [argonaut-shapeless](https://github.com/alexarchambault/argonaut-shapeless) and [scalacheck-shapeless](https://github.com/alexarchambault/scalacheck-shapeless) to it, without facing any particular issue, but their test suites are not as comprehensive than the ones of upickle.

I tended to avoid incremental compilation with the test suite of upickle, to compare the compile times to previous versions in particular. I don't really know if it could be risky here.

I would expect libraries using it to have two versions of their API, one with `Cached` implicits, and the other without, so that users could use the latter first, and switch to the former if compile time is an issue. That would lower the risk of irky things to happen.

alexarchambault:
@fommil  I'll write something about the kind of changes I made to lower compile times (benchmarking with https://github.com/alexarchambault/upickle-pprint). I can send you an update of spray-json-shapeless if ever it takes too long.

fommil:
:smile: yeah looks like `Lazy => Strict` and I'll play with `Strict.Cached`. I'll wait for this to be merged... I assume this will result in a big version bump in Shapeless?

milessabin:
Probably.

fommil:
I'll be testing via snapshots way ahead of that, hopefully. Just let me know when a snapshot is available please.

milessabin:
OK, I think I'd like to get this merged so that we can start getting people playing with snapshots.

We can defer naming and the remaining test issues until later.

We should bumped the version to 2.3.0-SNAPSHOT.

We should reinstate the MiMa checks ... with `previousArtifact` set to 2.3.0 this will be a NOP.

If you'd like to squash any of the commits, please do :-)

milessabin:
Ah, yes, rebase against master too ...

milessabin:
Do you have a feel for how much of an effect does the inlining has?

milessabin:
Playing around with the inlining a little ... I've been able to get the tests to pass by commenting out the `.orElse(inline)` [here](https://github.com/milessabin/shapeless/commit/8ea3907e9a3fab50d1cb18777f406602366bedd3#diff-ffd1855210ab80b59329091ae5193e48R538) which seems a bit less drastic than your suggestion of removing the whole `optimize` step.

Can you say a little more about what `prioritySubstitute` and `inline` are doing? I'm guessing that the former has important semantics, whereas the latter is (supposed to be) semantics neutral?

alexarchambault:
+1 for a RC process for this (and recommend people to give snapshots a try).

I bumped version to 2.3.0-SNAPSHOT, updated the mima previous artifact, and temporary disabled the mima check (all at once in https://github.com/milessabin/shapeless/commit/5529ba41ffce817632651f4231dffb1090108742).

I squashed some commits, but not all. Is it fine for you? I think it would be way harder to make sense of these changes if they were all squashed together.

It's rebased against the last master.

About inlining, in quick benchmarks that I just ran (based on https://github.com/alexarchambault/upickle-pprint), there's a slight gain (~10% speed up) if one removes it (all other things being equal). It mainly makes the output code simpler / more compact, which speeds up what's happening after implicit searches. The other optimization (simplifying trees with `Priority.High`) doesn't seem to bring much, I guess it could even be removed.

Inlining takes dict entries used only once (referred to by only one other entry), and substitutes their body in the latter (it replaces the `Ident(`some name`)` by the tree it is referring to). This makes the macro output simpler code.

The other optim detects implicits involving `Priority.High`. These are typically like `mkTC(Lazy(Priority.High(value)))`, and `mkTC` just unpack the value wrapped in `Priority.High` then `Lazy` here. So the macro does that directly (replacing the previous expression by just `value`). But it doesn't bring that high gains appearently, maybe I should just remove it.

milessabin:
Fantastic! Thanks :-)

I've been thinking about the tree-rewriting part of this PR a bit more ...

On the inlining, I think that if the gain is only of the order of 10% at compile time then I'd prefer _not_ to merge it, given that it's precisely that code which is causing trouble on 2.12.x. If the other transformation isn't making a significant contribution, then I'd prefer not to merge it either.

The reason I'm saying this is that these sorts of tree transformations, where we construct new trees from existing, already typechecked trees, are prone to very subtle bugs because of deficiencies in the the macro runtime. The thing that concerns me most is that even if we have a full set of passing tests, real code in the wild will use trees of different shapes and these will cause things to blow up in ways which are hard to predict, and hard for us to control given that the issues are in scalac. I've been playing whack-a-mole with prefix bugs for a while now, as you know, and I don't relish the prospect of a whole new class of hard to understand errors.

If the gains were more significant then it'd be a different story.

If you're willing to take those out I'll merge the rest of the PR immediately :-)

alexarchambault:
I removed inlining the PR. I think it can still be done through the newly added `LazyExtensions`, thanks to the (safe?) changes of https://github.com/milessabin/shapeless/commit/b2fc860f3b4cd50c24b3a337456e45fb80bf7646.

I'd like to add more tests of the global variants of `Lazy` and `Strict`, and possibly others I don't remember now, but these can be added later (and don't prevent merging).

milessabin:
Merged! Awesome job ... thank you :-)

milessabin:
Do you have bandwidth for the backport to 2.10.x?

alexarchambault:
I should in a few hours.

There's a not up-to-date one here: https://github.com/milessabin/shapeless/tree/topic/lazyextensions-2.10.x.

It should be too hard to port if you feel like trying. Else I'll do it later today, I remember the tricky points.

milessabin:
I'll leave it to you ... there's no huge rush :-)

fommil:
yeah, I'm happy knowing that it'll come one day. I suppose I should go try this out with spray-json-shapeless and ensime then :smile:

milessabin:
@fommil yes, please do :-)

fommil:
I get a stack overflow :cry: It also takes a lot longer to compile that before, taking about 70s for the tests.

I've pushed my changes to https://github.com/fommil/spray-json-shapeless if you could please have a look.

alexarchambault:
I didn't look at your code yet, but you know you must not only Stricts,
right? There still has to be some Lazys to break possible cycles.

fommil:
Nope didn't know that. Where are the docs of what to use when?

alexarchambault:
There are some scaladocs, but I have to write something about the big picture. It should be loosely based on https://github.com/alexarchambault/auto-type-class-benchmark (that needs to be updated, and is only for the braves in its current state :-)

fommil:
if you could please have a look at `sexp.scala` in the shapeless examples and advise where you believe a `Strict` would hely, that would be good.

alexarchambault:
I updated a bit spray-json-shapeless, in [this branch](https://github.com/alexarchambault/spray-json-shapeless/commits/topic/strict) (not starting from the last commit from master, but from the previous one). I get a ~20% speed up (~66 s before, ~53 s after), mainly thanks to `Global` it seems.

Beware if looking at the times here, to:
- checkout the last shapeless,
- change its version to a non snapshot one (like `2.3.0-87697b0`) in `version.sbt`,
- publish it locally,
- and use this version from spray-json-shapeless during the test.

This way, SBT won't lose time checking for new versions each time it is run (as it does for snapshots). Comparisons will be more reliable.

I get bigger speed ups (like 50% in the most extreme cases) in my [benchmark](https://github.com/alexarchambault/auto-type-class-benchmark) (link not up-to-date, better results since), but there are some differences with it (it is using `Priority` in particular).

To speed up compilation a bit more, you could try:
- not using `LabelledGeneric` (just `Generic`), relying on an ad-hoc type class, materialized by a macro, to supply the labels in a `List[String]` (and trusting it that this list has the right length),
- using ad-hoc type classes when deriving instances for HLists or coproducts (like `WrappedRootJsonHListFormat[Wrapped, L <: HList]` and `WrappedRootJsonCoproductFormat[Wrapped, C <: Coproduct]`, and having two cases instead of `familyFormat`, one for HLists, the other for coproducts).

These brought nice improvements to me in the benchmark above (like 10% each).

fommil:
I'm looking at this today, thanks

fommil:
@alexarchambault fantastic! Thank you. I have some questions:

1. you removed some `Lazy`s, is there an easy explanation why it was not needed here?
2. after moving to `Strict` you kept two `Lazy`s around, is there an easy explanation for that?

fommil:
btw, it's hard to say for definite, but it looks like ENSIME `jerk/test` (which is mostly compilation) has gone from 75 seconds to 60 seconds. Which is nice.

Did you say you'd expect a further performance enhancement by splitting `familyFormat` into two? I tried that and it didn't seem to make a difference.

PR #440, opened 2015-08-02
illTyped: Enable DOTALL mode so that "." matches newlines

fthomas:
This adds the `DOTALL` flag to the regex used by `illTyped` so that "." also matches newlines. This makes it easier to check errors which span over multiple lines.

I'm using `illTyped` a lot in refined to check if the compiler fails with the correct error, e.g.
```scala
illTyped("""regex("(a|b")""", "(?s)Regex predicate failed.*")
```
In this example the compile error spans over multiple lines which required to add the embedded flag expression `(?s)` that also enables the `DOTALL` mode. Without that flag, `illTyped` reports that the actual error does not match the expected error. It took me a little bit to figure out that I could enable the `DOTALL` mode from the regex itself.


non:
:+1:

milessabin:

PR #441, opened 2015-08-10
Add runtimeList method to HList

davegurnell:
This handy little method discards all desire for type-safety and converts an `HList` to a `List[Any]`. Potentially useful when using shapeless to add static typing over a dynamically typed library. Use cases include this part of slickless, where we interoperate with some underlying code in Slick:

https://github.com/underscoreio/slickless/blob/master/src/main/scala/slickless/HListShape.scala#L28-L32


milessabin:
Awesome :-)

A couple of things though: can you make it tail recursive (vars are forbidden but a `ListBuffer` is allowed). Also don't forget to add yourself to `CONTRIBUTORS` and the contributors section at the end of `README.md` :-)

davegurnell:
I hope I'm doing this the right way. `List.newBuilder` appears to create a `ListBuffer`, and `ListBuffer.+=` appears to take constant time.

milessabin:

Issue #443, opened 2015-08-12
Extract items/slices from HList/Product/Coproduct


AndreasKostler:
Akin
```scala 
"foo"::1::3.14::"bar"::42:: HNil getItems _0::_3::_4::HNil
res3: shapeless.::[String,shapeless.::[String,shapeless.::[Int,shapeless.HNil]]] = foo :: bar :: 42 :: HNil
// similarly
"foo"::1::3.14::"bar"::42:: HNil slice (_0,_3::HNil)
res4: shapeless.::[String,shapeless.::[Int,shapeless.::[Double,shapeless.HNil]]] = foo :: 1 :: 3.14 :: HNil
```

milessabin:
Fixed in #444.

PR #444, opened 2015-08-12
Slice and select items

AndreasKostler:
Implements #443
Implements #445
For your review

### DIFF
--- a/core/src/main/scala/shapeless/hlists.scala
+++ b/core/src/main/scala/shapeless/hlists.scala
@@ -62,7 +62,7 @@ object HList extends Dynamic {
   def apply() = HNil
 
   def apply[T](t: T) = t :: HNil
-  
+
   def apply[P <: Product, L <: HList](p : P)(implicit gen: Generic.Aux[P, L]) : L = gen.to(p)
 
   /**
@@ -74,7 +74,7 @@ object HList extends Dynamic {
    * Produces a `N1`-length HList made of `N2`-length HLists filled with `elem`.
    */
   def fill[A](n1: Nat, n2: Nat)(elem: A)(implicit fill: Fill[(n1.N, n2.N), A]) : fill.Out = fill(elem)
-  
+
   implicit def hlistOps[L <: HList](l : L) : HListOps[L] = new HListOps(l)
 
   /**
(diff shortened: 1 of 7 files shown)


milessabin:
Good stuff! Thanks :-)

I have a couple of naming quibbles. The term "item" isn't used anywhere else in shapeless, so it sticks out a little. I suggest renaming to `selectMany` (and renaming the op type class to `SelectMany` as well). And I think it would be better to rename `slice` to `selectRange`.

I'd also prefer it if you also defined a trivial `SelectRange` type class (which can be implemented entirely in terms of `Range` and `SelectMany`) so that `selectRange` has just a single implicit argument ... this makes it easier for someone using `selectRange` in the body of a method to propagate the required evidence from the outside.

AndreasKostler:
Done

milessabin:
Awesome :-)

One last thing ... if you define and use the `Nat` arguments in [this style](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/syntax/hlists.scala#L101),

```scala
def at(n : Nat)(implicit at : At[L, n.N]) : at.Out = at(l)
```

rather than specifying them as an argument of a type variable bounded by `Nat` then it's possible to use ordinary `Int` literals (ie 0, 1, 2) rather than `Nat` literals (_0, _1, _2). Also if you look a couple of lines above where I linked you'll see that the convention is to have two variants: one which takes a `Nat` value argument as above, and one which take a type argument (but no corresponding value argument). It would be quite nice to keep to that convention.

AndreasKostler:
I have added support for `Int` literals to the syntax functions for `selectRange` 
Unfortunately, I can't see how supporting a `List[Int]` for `selectMany` could be done easily. Is there a type class supporting the conversion from a `List[Int]` to a `HList <: Nat`?

milessabin:
Sorry about the confusion over `selectMany` ... you're right that there isn't anything currently in shapeless that will lift an argument list of `Int` literals to an `HList` of `Nat`.

We can replace the value argument with an explicit type argument however, so instead of writing,
```scala
selectMany(_2 :: _3 :: _5 :: HNil)
```
you could write,
```scala
selectMany[_2 :: _3 :: _5 :: HNil]
```

This eliminates the need for the value at runtime altogether, and is (a little) more consistent with `at`.

At some point we might add support for lifting an argument list of `Int` literals to an `HList` of `Nat` in which case `selectMany` could be made completely consistent with `at` and `selectRange` (in fact, if you wanted to take a look at that that would be awesome :-) ).

AndreasKostler:
Ok, will do. I will open a new ticket.

AndreasKostler:
All the pieces are in place now. However, I am a bit stuck making `selectMany` consistent with `at` and `selectRange`...any help is appreciated.

milessabin:
This looks good to me ... many thanks ... merging :-)

PR #446, opened 2015-08-17
Rework NatProductArgs macro, add unit tests, and update selectRange s…

AndreasKostler:
Rework NatProductArgs macro, add unit tests, and update selectRange syntax methods


milessabin:
I think that the closest analog that we have to the `selectMany` overloading issue is `replace` vs. `replaceType`, which suggests `selectManyType` rather than `selectManyL`. I'm not say that `selectManyType` is a good name, nor even that it's better than what you've chosen, but it is at least consistent with what's already there.

Aside from that ... LGTM :-)

AndreasKostler:
Thanks, done :)

milessabin:
Many thanks ... merging :-)

PR #449, opened 2015-08-20
fix toString for nested HList

nevillelyh:


milessabin:
Sorry it's taken me so long to get to this! Yes, this is a good idea. Many thanks ... merging :-)

Issue #190, opened 2014-07-19
Add Nat ranges and zipWithIndex to HList and tuples


travisbrown:
As discussed [here](http://stackoverflow.com/a/24814753/334519). Three questions:

Do we want to allow the user to specify the step for ranges? It adds a lot of complexity for something that's questionably useful.

Any thoughts on whether `(1 :: 'a :: HNil).zipWithIndex` should be `(1 :: _0 :: HNil) :: ('a :: _1 :: HNil) :: HNil` or `(1, _0) :: ('a, _1) :: HNil`? My vote's for the latter but the former might be more consistent.

Is there really any reason to provide this for `Coproduct`?

milessabin:
Fixed by @AndreasKostler in #450.

PR #450, opened 2015-08-27
ZipWithIndex

AndreasKostler:
Implements #190

### DIFF
--- a/core/src/test/scala/shapeless/nat.scala
+++ b/core/src/test/scala/shapeless/nat.scala
@@ -131,10 +131,10 @@ class NatTests {
 
     import shapeless.testutil._
 
-    assertTypedEquals[r1.Out](HNil, r1())
-    assertTypedEquals[r2.Out](_0::HNil, r2())
-    assertTypedEquals[r3.Out](HNil, r3())
-    assertTypedEquals[r4.Out](_1::_2::_3::_4::HNil, r4())
+    assertTypedEquals[HNil](HNil, r1())
+    assertTypedEquals[_0::HNil](_0::HNil, r2())
+    assertTypedEquals[HNil](HNil, r3())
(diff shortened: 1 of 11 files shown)


core/src/test/scala/shapeless/nat.scala
```diff
@@ -124,6 +124,16 @@ class NatTests {
...
+    val r2 = implicitly[Range[_0,_1]]
+    val r3 = implicitly[Range[_1,_1]]
+    val r4 = implicitly[Range[_1,_5]]
+
+    assertEquals(HNil, r1())
+    assertEquals(_0::HNil, r2())
+    assertEquals(HNil, r3())
+    assertEquals(_1::_2::_3::_4::HNil, r4())
+
     // Type level
```
milessabin: This would be better with `the` instead of `implicitly`, then you could use `assertTypedEquals` to verify that the inferred types are correct, as well as the values.

milessabin:
Awesome stuff ... thanks!

Good to merge once the `Range` test has been tweaked :-)

AndreasKostler:
Done, thanks for the tip :)

milessabin:
Almost there, but I meant use the expected type (ie. `HNil`, `_0 :: HNil` etc.) as the type argument to `assertTypedEquals` rather than the actual type (ie. `r1.Out`, `r2.Out` etc.). Otherwise the computed types could be, eg., `Any` and we'd be none the wiser.

milessabin:
Arggh ... sorry I missed that last commit until just now :-(

This is perfect now ... merging ... many thanks :-)

jarreds:
```zipWithIndex``` appears to require exponential compile time. Details here https://github.com/milessabin/shapeless/issues/490.

PR #451, opened 2015-08-27
Fix implicit priority issues with UnaryTCConstraint, allowing better inference of type constructor

mpilquist:
I noticed this issue when writing `isequence` here: https://gist.github.com/mpilquist/ba9a9fc522a2ed0a0a7b#file-inv-scala-L26-L32

In order for `F` param to be inferred, I had to define the implicit constraint *after* the the `InvariantMonoidal[F]` parameter. With this change, the order should be irrelevant (though I haven't tested that).


milessabin:
Many thanks ... merging :-)

Issue #145, opened 2014-06-26
Lazy pack factors out (co-)product recursion.


milessabin:

milessabin:
This probably needs to be rethought as part of the SMC/CPS reworking for shapeless 3.0.0.

alexarchambault:
I'm currently trying various ways of doing this... Using representations of products and sums as nested `Tuple2` and `Either`s, yes.

alexarchambault:
FYI, the -1 (minus one) version is on [this branch](https://github.com/alexarchambault/shapeless/tree/topic/derive).

Illustrated [here](https://github.com/alexarchambault/shapeless/blob/topic/derive/core/src/main/scala/shapeless/derive.scala#L362-L368), along with [this](https://github.com/alexarchambault/shapeless/blob/topic/derive/core/src/main/scala/shapeless/derive.scala#L373-L374), and used in [this notebook](https://github.com/alexarchambault/shapeless/blob/topic/derive/shapeless%20derive.ipynb) (that I'm using to debug this).

alexarchambault:
In particular, I'm trying to have a loose enough implementation, in particular:
- `product` and `sum` methods with optional arguments like `name` (name of the field / constructor), `default` (default value of the field, if any - half done), maybe `someNameAnnotation` that would look for annotation `@someName(value1, value2)`, and supply `Some((value1, value2))` (not done yet),
- all methods accept optional implicit arguments,
- all arguments can optionally be by-name,
- ...

The only constraints in the current implementation are:
- the methods must be put in a member called `typeClass` of the companion of the type class,
- method names (must be `point`, `product`, `sum`, `map`),
- their first type arguments must be the ones of the example above (one type parameter for `point`, 2 for `product`, `sum`, `map`), in the exact same order,
- if `map` needs `from` and/or `to` functions, it must accept argument(s) with these exact name and signature,
- the first parameter list of `point` must consist in only one argument with the type of the example,
- ...

alexarchambault:
Also, note that:
- it's implemented as a `Lazy` extension (those of https://github.com/milessabin/shapeless/pull/435), so it should deal well with `Lazy` / `Strict`, and has to be used through them,
- it uses a shapeless-3 like (?) `Generic`, with representations made of nested `Tuple2` / `Either`, terminated by `Unit` or `Nothing`.

milessabin:
This is sounding more ambitious than the title (and original intention behind) this ticket ;-)

[...]

Issue #453, opened 2015-09-02
Generic instance exists for AnyRef


rhendric:
`implicitly[shapeless.Generic[AnyRef]]` should fail, doesn't.

This also seems to affect other empty-constructor non-case classes (of which `java.lang.Object` is a special case). If I may editorialize on a fix, I'd propose adding a check to [`CaseClassMacros.isCaseClassLike.checkCtor`](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/generic.scala#L575) requiring `params` to be non-empty, rather than just adding a guard for `AnyRef` specifically.

milessabin:
I think your proposal is too drastic ... we definitely want to support case classes with empty constructors, case objects, `Unit` etc. The problem isn't so much the absence of contents, it's the availability of subtypes.

rhendric:
I think actual case classes/objects/Unit would be fine; there's a check for actual case-class-ness on line 586, and a check for Unit in `isProduct`. This would only exclude non-case classes with empty constructors from being considered as empty products—and even those would have a second chance to be coproducts if they do have available subtypes, right?

milessabin:
I still want to allow non-case classes with empty constructors to support [case classes a la carte](https://github.com/milessabin/shapeless/blob/master/examples/src/main/scala/shapeless/examples/alacarte.scala).

I'm rapidly coming to the conclusion that `AnyRef` really _is_ special. Other empty product types lack `AnyRef`'s near-universal supertype aspect, and I think that's just about enough to justify making a special exception.

rhendric:
I agree that `AnyRef` is pretty special, and I don't think special-casing it is an awful idea. But I also think that having Generic support for empty-constructor classes can be a little dangerous.

A class with only an empty constructor is much more likely to have object identity semantics than value equality semantics. Take, as just an example, a custom Exception class that doesn't expose a message constructor. This CustomException and that CustomException are different things—they have different stack traces and so on. But if CustomException is used in a class that gets put through some scrap-your-boilerplate machinery, as far as Generic is concerned, that CustomException can be crushed into HNil and then rematerialized as a new CustomException with the wrong stack trace and so on. I'd much rather have a compiler error.

[...]

Issue #371, opened 2015-04-24
Unable to derive type class for Scrooge generated product class


philwills:
[Scrooge](https://github.com/twitter/scrooge) generates classes representing thrift structs which if you squint hard look a bit like case classes, but are different enough that Shapeless doesn't recognise them as such, so is unable to derive type classes.

I created an example repo (https://github.com/philwills/scrooge-shapeless-example) which if you fire up the console and run the following shows the issue:

```scala
scala> import scrooge.shapeless.example._
scala> import scrooge.shapeless.example.Show._
scala> def s[T](t: T)(implicit s: Show[T]) = s.show(t)
scala> s(Prod(1, "f"))
res0: String = a = 1.0, b = f
scala> s(thrift.Product(1, "f"))
<console>:15: error: could not find implicit value for parameter s: scrooge.shapeless.example.Show[thrift.Product]

```

The generated thrift if we cut out a lot of guff looks like
```scala
object Product extends ThriftStructCodec3[Product] {
   def apply(a: Double, b: String): Product =
    new Immutable(a, b)

  class Immutable(val a: Double, val b: String,override val _passthroughFields: immutable$Map[Short, TFieldBlob]) extends Product {
    def this(a: Double, b: String) = this(a, b, Map.empty)
  }
  
  def unapply(_item: Product): Option[scala.Product2[Double, String]] = Some(_item)
}

trait Product
  extends ThriftStruct
  with scala.Product2[Double, String]
  with java.io.Serializable
{
  import Product._

  def a: Double
  def b: String

  def _passthroughFields: immutable$Map[Short, TFieldBlob] = immutable$Map.empty

  def _1 = a
  def _2 = b

  override def productArity: Int = 2

  override def productElement(n: Int): Any = n match {
    case 0 => this.a
    case 1 => this.b
    case _ => throw new IndexOutOfBoundsException(n.toString)
  }

  override def productPrefix: String = "Product"
}
```

i.e. it is a non-sealed trait, rather than a class which puts shapeless off the scent.

milessabin:
Just looking at this again ... are you able to summon a `Generic` instance for `Product.Immutable`?

philwills:
```scala
 implicitly[shapeless.Generic[thrift.Product.Immutable]]
```
fails to summon anything.

Incidentally, I did have a stab at trying to address this by broadening the `isCaseClassLike` check, but then ran into issues in `mkProductCases` that were sufficiently deep in the AST that I decided to retreat.

[...]

PR #459, opened 2015-09-16
Add Derive Lazy extension - derive type classes for products / sums...

alexarchambault:
...without the overhead of recursive implicit lookup on HList-s or Coproduct-s

If you think this fixes https://github.com/milessabin/shapeless/issues/145 and are fine with the way it's done, it could be merged.

I'm about to test and benchmark it outside of shapeless, to ensure these changes are worth it. So if this goes beyond your initial idea in https://github.com/milessabin/shapeless/issues/145, maybe wait for this benchmark.

I'd add some scaladoc too, if this PR is fine for you.

### DIFF
--- a/core/src/main/scala/shapeless/derived.scala
+++ b/core/src/main/scala/shapeless/derived.scala
@@ -1,0 +1,172 @@
+package shapeless
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+trait Derived[+T] extends Serializable {
+  def value: T
+}
+
+object Derived extends LazyExtensionCompanion {
+  def apply[T](t: T): Derived[T] =
+    new Derived[T] {
+      def value = t
+    }
+
+  implicit def init[T]: Derived[T] = macro initImpl
+
+  def instantiate(ctx0: DerivationContext): LazyExtension { type Ctx = ctx0.type } =
+    new DerivedLazyExtension {
+      type Ctx = ctx0.type
+      val ctx: ctx0.type = ctx0
+    }
+}
+
+trait DerivedTypes {
+  type C <: whitebox.Context
+  val c: C
+
+  import c.universe._
+
+  def deriveTpe: Type = typeOf[Derived[_]].typeConstructor
+
+  object DeriveTpe {
+    def unapply(tpe: Type): Option[(Type, Type)] =
+      tpe.dealias match {
+        case TypeRef(_, cpdTpe, List(derivedTpe))
+          if cpdTpe.asType.toType.typeConstructor =:= deriveTpe =>
+          derivedTpe.dealias match {
+            case TypeRef(_, tcTpe, List(dTpe)) => Some((tcTpe.asType.toType.typeConstructor, dTpe))
+            case _ => None
+          }
+
+        case _ => None
+      }
+  }
+
+}
+
+trait DerivedLazyExtension extends DerivedTypes with LazyExtension with CaseClassMacros {
+  type C = ctx.c.type
+  lazy val c: C = ctx.c
+
+  import ctx._
+  import c.universe._
+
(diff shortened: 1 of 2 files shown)


milessabin:
Annoyingly this is directly colliding with the work in `Generic` and `Generic1` that I'm doing for #453 and #371 ... can you elaborate on why you need to touch that part of the code for this ticket?

alexarchambault:
I do only very little changes in it. The diff just doesn't look nice from Github... (it's nicer from Intellij).

I just extract the `mkProductCases` and `mkCoproductCases` methods from `mkProductGeneric` and `mkCoproductGeneric`, and have `mkProductCases` accept `nil: Tree` and `cons: Tree` arguments, so that I can provide `q"()"` and `q"Tuple2"` from `derive.scala` (instead of `q"HNil"` and `q"shapeless.::"` from `generic.scala`).

milessabin:
Gotcha ... I'm right in the middle of a significant refactor of that area though. If it's not essential to this PR could you revert that part?

alexarchambault:
The changes in `generic.scala` allow for some re-use of things from it. Either what I re-used is still around after your refactor, and and I should use the new form, or the refactor is broader, and things from `derive.scala` have to be rethought...

milessabin:
OK, but so long as you know that I'm going to merge my changes first ;-)

alexarchambault:
Your changes are planned for 2.3.0, right?

milessabin:
Yes, they are. It's looking like I've run out of time now until after Scala World unfortunately, so this is going to be pushed back a little :-(

travisbrown:
I was just glancing over this change, and the idea of traversing companion objects looking for an appropriately named `typeClass` object with appropriately named methods with appropriately named parameters scares me a little. :smile: Is there a reason not to use something a little more like the current `TypeClass` type classes?

alexarchambault:
@travisbrown The main advantage of this is that it allows for more optional feature combinations. The current `TypeClass` allows to handle either just products or both products and coproducts, and get the labels of fields/constructors or not, and that's it. That's 2 x 2 possibilities. When using `Lazy`, one often does other things that would not have been possible with just these, like looking for extra type classes for the field/constructor types or the type a type class is derived for, or even getting default values of fields ;-), or we could want to pull some annotations on them, etc. I didn't find a way to keep `TypeClass`-like types, and at the same time handling all these extra possibilities. That's why I went for "brute compile-time reflection" :-).

On the overall, the compile-time speed up I hope this could bring (I'll check that soon) are the main motivation for these attempts...

milessabin:
@alexarchambault's observation about flexibility is what led me to prefer simply using `Lazy`, `Generic` and vanilla implicit resolution over `TypeClass` in the end ... once all the possible feature combinations are added to the mix the relative complexity of an extended `TypeClass` vs. hand rolling seemed to tip in favour of hand rolling.

On balance I think I would be most in favour of hand rolling for the general case with one or more restricted forms (like `TypeClass` is now) which are specialized to and simplify particular applications.

milessabin:
The point about compile-time speed up is good. Is there some way that the speedups can be made available in the hand rolled case, or is a `TypeClass`-like framework essential? If it is, then why?

alexarchambault:
Updated so that this uses the implicit instances of `TypeClass` and the like - and thus only work in the simple cases of `TypeClass` et al.

Just a (working) draft for now.

Is it blocking for shapeless 2.3.0? I could manage to clean this very soon then.

milessabin:
I've done as much on `Generic` as I'm going to do for 2.3.0 now, so if you haven't completely given up on this by now, please have at it! :-)

alexarchambault:
@milessabin I'll have a look! Apart from tidying up the code, there's a conflict with orphan derivation to address too.

Issue #461, opened 2015-09-18
Add typeclass giving access to constructor default arguments


milessabin:
/cc @travisbrown

alexarchambault:
I did something similar while benchmarking type class derivation some time ago: https://github.com/alexarchambault/upickle-pprint/blob/topic/shapeless/upickle/shared/src/main/scala/upickle/experimental/Default.scala

Could be backported here.

milessabin:
Go for it! :-)

PR #462, opened 2015-09-18
Add Default type class

alexarchambault:
Fixes https://github.com/milessabin/shapeless/issues/461

### DIFF
--- a/core/src/main/scala/shapeless/default.scala
+++ b/core/src/main/scala/shapeless/default.scala
@@ -1,0 +1,86 @@
+package shapeless
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+/**
+ * Provides default values of case class-like types.
+ *
+ * The `Out` type parameter is an HList type whose length is the number of fields of `T`. Its elements correspond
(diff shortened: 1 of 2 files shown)


alexarchambault:
Beware that a test that I would expect to fail ([`Default[AnyRef]`](https://github.com/alexarchambault/shapeless/commit/08dd3c102326acbff2745094bcc8d4b8e093db84#diff-e62c779b4e79399cb728676a9c58e196R40)), actually passes. This should be related to https://github.com/milessabin/shapeless/issues/453.

milessabin:
+1 for creative use of `None.type` and `Some[t]`.

I wonder, though, if it wouldn't be better to return a record containing only fields for the default values which are present? Then it would be possible to use the record `Merge` operation to combine defaults with, eg., the records produced by `RecordArgs`.

travisbrown:
+1, this is great.

I don't have strong opinions on the representation. In general it seems likely that returning a record would be more convenient, but in the case of circe in particular I don't think there'd be much of a difference in ease of implementation, and this representation might actually be a little nicer.

milessabin:
@travisbrown that's good enough for me ... merging :-)

fommil:
:+1:

PR #463, opened 2015-09-20
Various enhancements to Default

alexarchambault:
See the commit messages and content for more details.

In particular, this adds extra `Default`-like type classes, providing the default values as a record or an HList of `Option[...]` (instead of `None.type` / `Some[...]`).


alexarchambault:
As a side node, as soon as your refactoring of the generic macros is out, type classes providing annotations of fields / constructors, or of just a given type, could be added.

milessabin:

PR #464, opened 2015-09-21
Add Annotation(s) type classes

alexarchambault:
This PR adds type classes `Annotation` and `Annotations` (latter ends with `s`).

`Annotation[A, T]` provides the annotation of type `A` on type `T`, if any, like
```scala
case class First(i: Int) // annotation type

@First(3) trait Something

val somethingFirst = Annotation[First, Something].apply()
assert(somethingFirst == First(3))
```

`Annotations[A, T]` provides annotations on the fields or constructors of `T`, a bit like `Default` provides default values. Example usage:
```scala
case class First(s: String) // annotation type

case class CC(i: Int, @First("a") s: String)

sealed trait Base
@First("b") case class BaseI(i: Int) extends Base
case class BaseS(s: String) extends Base

val ccFirsts = Annotations[First, CC]
val baseFirsts = Annotations[First, Base]

// ccFirsts.Out is  None.type :: Some[First] :: HNil
// ccFirsts.apply() is
//   None :: Some(First("a")) :: HNil

// baseFirsts.Out is  Some[First] :: None.type :: HNil
// baseFirsts.apply() is
//   Some(First("b")) :: None :: HNil
```

### DIFF
--- a/core/src/main/scala/shapeless/annotation.scala
+++ b/core/src/main/scala/shapeless/annotation.scala
@@ -1,0 +1,137 @@
+package shapeless
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+/**
+ * Provides the annotations of type `A` of the fields or constructors of case class-like or sum type `T`.
+ *
+ * If type `T` is case class-like, this type class inspects its fields and provides their annotations of type `A`. If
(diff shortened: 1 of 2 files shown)


milessabin:
This looks very useful :-)

Merging ...

PR #466, opened 2015-09-23
Grouper type class, ToCoproduct for Tuples, ToSum type class

AndreasKostler:
* ToSum implementation as discussed
* ToCoproduct was missing for tuples
* Grouper implementation (I would really like to call this `partition` but that one's taken already)
* PaddedGrouper implementation
* Greatest Common Divisor and Least Common Multiple for nat


milessabin:
Awesome stuff! Thanks :-)

PR #467, opened 2015-09-25
Additional implicit resolutions for LT and LTEq

TomasMikula:
This pull request modifies implicits for `LT` and `LTEq` so that the following cases are covered without the need for any additional knowledge about N (other than `N <: Nat`):

 - `_0 <= N`
 - `N <= N`
 - `N < Succ[N]`
 - `N <= Succ[N]`

See the added test cases.

### DIFF
--- a/core/src/test/scala/shapeless/nat.scala
+++ b/core/src/test/scala/shapeless/nat.scala
@@ -22,6 +22,8 @@ import org.junit.Assert._
 class NatTests {
   import nat._
   import ops.nat._
+
+  type N <: Nat
   
   trait Check[N <: Nat]
   def check(expected: Nat)(actually : => Check[expected.N]) {}
@@ -85,6 +87,8 @@ class NatTests {
     implicitly[LT[_10, _15]]
     implicitly[LTEq[_2, _2]]
     implicitly[LTEq[_2, _3]]
+    implicitly[LTEq[_0, N]]
+    implicitly[LTEq[N, N]]
 
     implicitly[Min.Aux[_0, _0, _0]]
     implicitly[Min.Aux[_5, _2, _2]]
--- a/core/src/test/scala/shapeless/nat.scala
+++ b/core/src/test/scala/shapeless/nat.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-13 Miles Sabin 
+ * Copyright (c) 2011-15 Miles Sabin
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package shapeless
 
 import org.junit.Test
 import org.junit.Assert._
+import shapeless.test.illTyped
 
 class NatTests {
   import nat._
@@ -86,12 +87,20 @@ class NatTests {
     implicitly[LTEq[_2, _2]]
     implicitly[LTEq[_2, _3]]
 
+    illTyped(""" implicitly[LT[_5, _5]] """)
+    illTyped(""" implicitly[LTEq[_6, _5]] """)
+
     def relativeToN[N <: Nat]: Unit = {
       implicitly[LT[_0, Succ[N]]]
       implicitly[LT[N, Succ[N]]]
       implicitly[LTEq[_0, N]]
       implicitly[LTEq[N, N]]
       implicitly[LTEq[N, Succ[N]]]
+
+      illTyped(""" implicitly[LT[_0, N]] """)
+      illTyped(""" implicitly[LT[N, N]] """)
+      illTyped(""" implicitly[LTEq[_1, N]] """)
+      illTyped(""" implicitly[LTEq[Succ[N], N]] """)
     }
 
     implicitly[Min.Aux[_0, _0, _0]]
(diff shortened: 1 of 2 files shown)


TomasMikula:
I also wanted to have transitivity, i.e.

```scala
implicit def transitivity[A <: Nat, B <: Nat, C <: Nat](implicit
    lteq1: A <= B,
    lteq2: B <= C) =
  new <=[A, C] {}
```

but that could cause the type inference to diverge. Would having a `compose` method be a good idea?

```scala
trait LTEq[A <: Nat, B <: Nat] {
  def compose[C <: Nat](lteq: B <= C) = new <=[A, C] {}
}
```

milessabin:
Great stuff! :-)

Wrapping the types of the implicit arguments of transitivity in `Lazy` will probably prevent divergence. Do you want to give that a try before I merge?

TomasMikula:
No luck with `Lazy` so far. It's likely that I'm doing something wrong. I isolated the problem, to be sure nothing is interfering. Here is what I'm struggling with:

**Version 1**

```scala
import shapeless.Lazy

trait Rel[A, B]

object Rel {
  implicit def transitivity[A, B, C](implicit
      r1: Lazy[Rel[A, B]],
      r2: Lazy[Rel[B, C]]): Rel[A, C] =
    new Rel[A, C] {}

  def transitivityTest[A, B, C](implicit r1: Rel[A, B], r2: Rel[B, C]): Unit = {
    implicitly[Rel[A, C]]
  }
}
```

This causes the compilation to go into an infinite loop (no error reported, it just does not terminate).

**Version 2**

The only difference here is the `transitivityTest` method moved to a separate module.

```scala
import shapeless.Lazy

trait Rel[A, B]

object Rel {
  implicit def transitivity[A, B, C](implicit
      r1: Lazy[Rel[A, B]],
      r2: Lazy[Rel[B, C]]): Rel[A, C] =
    new Rel[A, C] {}
}

object TransitivityTest {
  def transitivityTest[A, B, C](implicit r1: Rel[A, B], r2: Rel[B, C]): Unit = {
    implicitly[Rel[A, C]]
  }
}
```

This results in

```
could not find implicit value for parameter e: Rel[A,C]
    implicitly[Rel[A, C]]
              ^
```

If I add `import Rel._` to `TransitivityTest`, I get an infinite loop again.

milessabin:
There's something not quite right with those two examples ... neither of them mention `LTEq` (or have an inductive base case of some other sort).

TomasMikula:
Not mentioning `LTEq` is on purpose, to make it clear that the examples are self-contained and that the implicit resolution does not interfere with anything else in `LTEq`. Think of `Rel` in the examples as `LTEq`, or any other transitive relation.

One would need a base case to make `Rel` any useful, but the code above (most importantly the `transitivityTest` method) should still compile, right? I wanted to keep the examples minimal.

milessabin:
Well, the use of `Lazy` has the effect of suppressing the divergence checker, so your version 1 will chase its tail indefinitely, as you've observed. You really do need a base case of some sort there.

TomasMikula:
Inside the `transitivityTest` method, don't the arguments `r1: Rel[A, B]` and `r2: Rel[B, C]` serve as a base case?

milessabin:
No, I think that `transitivity` is just going to loop on its first implicit argument.

BTW, I'm happy to merge without `transitivity` if you're happy with the rest.

TomasMikula:
Yeah, I might try to work out transitivity at some later point. Thanks.

TomasMikula:
Interestingly enough, if I omit `Lazy` from my constrained examples, they compile:

```scala
trait Rel[A, B]

object Rel {
  implicit def transitivity[A, B, C](implicit
      r1: /* not Lazy */ Rel[A, B],
      r2: /* not Lazy */ Rel[B, C]): Rel[A, C] =
    new Rel[A, C] {}

  def transitivityTest[A, B, C](implicit r1: Rel[A, B], r2: Rel[B, C]): Unit = {
    implicitly[Rel[A, C]]
  }
}
```

TomasMikula:
> Yeah, I might try to work out transitivity at some later point. Thanks.

By which I meant that I'm happy if you merge this PR and I might try to work out transitivity in the future as a separate PR.

milessabin:
Sorry it's taken me so long to get back to this ... I've been preoccupied with family issues.

core/src/test/scala/shapeless/nat.scala
```diff
@@ -86,6 +86,14 @@ class NatTests {
...
 
+    def relativeToN[N <: Nat]: Unit = {
+      implicitly[LT[_0, Succ[N]]]
+      implicitly[LT[N, Succ[N]]]
+      implicitly[LTEq[_0, N]]
+      implicitly[LTEq[N, N]]
+      implicitly[LTEq[N, Succ[N]]]
+    }
+
     implicitly[Min.Aux[_0, _0, _0]]
```
milessabin: It might be an idea to add a few negative tests using `illTyped`.

milessabin:
With a couple of negative tests this LGTM ... thanks! :-)

TomasMikula:
No worries, I hope everything is OK. I added some negative tests.

milessabin:
Perfect ... many thanks. Merging.

PR #471, opened 2015-09-27
Add ModifierAt typeclass and tests

AndreasKostler:


milessabin:
This is great! Thanks :-)

A couple of things though.

First, I think we can probably get better type inference for the function argument. See `updateWith` in record [here](https://github.com/AndreasKostler/shapeless/blob/master/core/src/main/scala/shapeless/syntax/records.scala#L63). The `WitnessWith` conversion target for the index argument (a key in the record case, a `Nat` index in yours) resolves the element type early enough to determine the argument type for the update function.

The other thing is that I think that `updateAt` is too similar to `updatedAt` ... we need to find another name ;-)

AndreasKostler:
Oh nice one! I didn't know about `WitnessWith` - thanks a bunch.
Any suggestions for a name? I guess `modifyAt` breaks with `updateWith` but maybe it's the way to go?

milessabin:
It's a shame that `updateWith` is already taken for something less useful. It's fairly barbarous, but how about `updateAtWith`? At least it's reasonably clear what it means.

AndreasKostler:
Just thinking about this, how would I go about using the `WitnessWith` approach for `updateAt[N]`? I shouldn't really use a second implicit, should I?
If I understood you correctly, (something like) this is what you are talking about 
```scala
def updateAt[V](n: WitnessWith[LSL])(f : n.instance.Out => V)(implicit
                                                                upd: ModifierAt[L, n.T, n.instance.Out, V]
    ): upd.Out = upd(l, f)

  type LSL[N <: Nat] = At[L, N]
```
So `ModifierAt` now takes extra `U` and `V` type parameters instead of `F`. I still need to carry `N` along otherwise `HList`s with more than one `U` would be handled incorrectly. 

With this approach I get a type error since `LSL`s type parameter bounds are tighter than `WitnessWith`s `TC` bounds.  

Maybe I misunderstood where you were getting at. Any pointers?

milessabin:
Does `WitnessWith.Lt` get you any further?

AndreasKostler:
Not really: 
```
[error] HListOps.this.LSL's type parameters do not match type TC's expected parameters:
[error] type N's bounds <: shapeless.Nat are stricter than type _'s declared bounds >: Nothing <: Any
[error]   def updateAt[V](n: WitnessWith.Lt[LSL, Nat])(f : n.instance.Out => V)(implicit
```

milessabin:
Right. I think what we need here is actually a `NatWith` along similar lines to `WitnessWith` ... would you like to take a swing at that?

AndreasKostler:
I can try...

milessabin:
Awesome :-)

AndreasKostler:
Hmm, I fail right at the start...the simplest, specced down, do nothing case I can think of is this
```scala
trait NatWith[TC[_]] extends Witness {
  type T <: Nat
  val instance: TC[T]
}

object NatWith {

  implicit def apply1[TC[_]](i: Int) = {}

}
```

which fails quite miserable with the same error
```
error: kinds of the type arguments (LAT) do not conform to the expected kinds of the type parameters (type TC) in trait NatWith.
LAT's type parameters do not match type TC's expected parameters:
type N's bounds <: shapeless.Nat are stricter than type _'s declared bounds >: Nothing <: Any
       def foo(i: NatWith[LAT]) = i
```

AndreasKostler:
Ok, so I got a little further:
```scala
trait NatWith[TC[_ <: Nat]] extends Witness {

  val instance: TC[Nat]
}

object NatWith {

  implicit def apply1[TC[_ <: Nat]](i: Int): NatWith[TC] = macro SingletonTypeMacros.convertInstanceImplNat[TC]

}
```
and
```scala
def convertInstanceImplNat[TC[_ <: Nat]](i: Expr[Int])(implicit
                                                         tcTag: WeakTypeTag[TC[_]]
    ): Tree = {
    val nat = NatMacros.materializeSingleton(this.c)(i)
    val natTpd = c.typecheck(nat)
    val natExpr = c.Expr[Nat](natTpd)
    extractResult(natExpr) { (sTpe, value) =>
      val tc = tcTag.tpe.typeConstructor
      val wwTC = typeOf[NatWith[Nothing]].typeConstructor
      val parent = appliedType(wwTC, List(tc))
      val tci = appliedType(tc, List(sTpe))
      val i = inferInstance(tci)
      mkWitnessWith(parent, sTpe, value, i)
    }
  }
```
`inferInstance` barfs at me though:
```
type LAT[N<:Nat] = At[L, N]
def foo(i: NatWith[LAT]) = {}
foo(1)
error: Unable to resolve implicit value of type LAT[nat_$macro$2.type]
```
Any ideas? Am I missing something obvious? The problem is that `c.inferImplicitValue(tci)` returns an `emptyTree` in `inferInstance`

AndreasKostler:
@milessabin I got quite a bit further still. The macro seems to be doing the right thing which the following REPL session makes me believe:
```scala
type L = Int::Int::Int::HNil
val l: L = 1::2::3::HNil
def updateAtWith[F, V](n: NatWith[LAT])(f : n.instance.Out => V)(implicit upd: ModifierAt[L, n.N,n.instance.Out, V]): upd.Out = upd(l, f)
updateAtWith(1)( _ => 42)
res3: (Int, shapeless.::[Int,shapeless.::[Int,shapeless.::[Int,shapeless.HNil]]]) = (2,1 :: 42 :: 3 :: HNil)
```

The `updateAtWith` function is a copy-and-paste clone of what's in `syntax.HListOps`. However, there it doesn't work:
```scala
1::2::3::HNil updateAtWith(1)( _ => 42 )
<console>:20: error: Int(1) does not take parameters
       1::2::3::HNil updateAtWith(1)( _ => 42 )
```
I am quite puzzled by this. I've pushed all the code, maybe you could have a look. I suspect it's something minor but right now I'm hitting a wall.

milessabin:
At a wild guess I would say that it's a hygiene bug in your macro ... make sure that all reference to types or terms are fully qualified, starting with `_root_`.

AndreasKostler:
Maybe my eyes are not attuned to it but I can't make out any hygiene issues...

core/src/main/scala/shapeless/syntax/hlists.scala
```diff
@@ -221,6 +221,23 @@ final class HListOps[L <: HList](l : L) extends Serializable {
...
+   * Available only if there is evidence that this `HList` has `N` elements.
+   *
+   * @author Andreas Koestler
+   */
+  //class UpdateAtWithAux[N <: Nat] {
+  //  def apply[F](f : F)(implicit upd: ModifierAt[L, N, F]): upd.Out = upd(l, f)
+  //}
+  //def updateAtWith[N <: Nat] = new UpdateAtAux[N]
+
+  def updateAtWith[F, V](n: NatWith[LAT])(f : n.instance.Out => V)(implicit
```
milessabin: I think `F` is a stray.

milessabin:
OK, I think that the implicit conversion from `Int` to `NatWith[LAT]` isn't being applied from some reason.

milessabin:
Or alternatively, it is being attempted, but the expansion is being rejected. I suggesting type checking your result tree to see if there's an error there that's being masked ... ie. `c.typecheck(res, silent = false)`. FWIW, it think you should take a close look at `extends $parent` ... I think that might be unhygenic.

AndreasKostler:
Hmm, I think the macro expansion isn't even happening in the `1::2::3::HNil updateAtWith(1)( _ => 42 )` case: Any debugging information I'm adding doesn't get printed at all. 
Fully qualifying `parent` doesn't hep:
```scala
{
  final class anon$macro$1 extends _root_.shapeless.NatWith[LAT] {
    def <init>() = {
      super.<init>();
      ()
    };
    val instance: shapeless.ops.hlist.At[shapeless.::[Int,shapeless.::[Int,shapeless.::[Int,shapeless.HNil]]],shapeless.Succ[shapeless._0]]{type Out = Int} = hlist.this.At.hlistAtN[Int, shapeless.::[Int,shapeless.::[Int,shapeless.HNil]], shapeless._0](hlist.this.At.hlistAtZero[Int, shapeless.::[Int,shapeless.HNil]]);
    type N = shapeless.Succ[shapeless._0];
    val value: shapeless.Succ[shapeless._0] = new shapeless.Succ[shapeless._0]()
  };
  new anon$macro$1()
}
```

AndreasKostler:
@milessabin any idea who I could bug with this? I had the guys at work have a look but no-one found any obvious issues. My macro-fu isn't nearly strong enough to work this out ;)

milessabin:
Maybe ask for help on the gitter channel? I'll try and find time to take a look before 2.3.0, but I can't make any promises.

AndreasKostler:
@milessabin Any chance you had another look at it? I have a couple of more unrelated pull requests that I don't want to conflate with this one.

milessabin:
I'm afraid I've been tied up with the single branch build ... did you try taking this to the gitter channel?

AndreasKostler:
Yes I have - I didn't get any feedback. Any other channels I might try? The
macro isn't special in any way - I really don't see how it works in the
REPL and not in the unit tests...

milessabin:
Oh, I don't remember seeing that go past. Try asking again, and if you get no joy I'll try and find time to take a look next week.

AndreasKostler:
@milessabin I posted it again - to no avail...It doesn't look too hard, I am really puzzled by this one.

milessabin:
OK, I'll take a look next week ...

milessabin:
The main issue seems to have been the bounds on the type parameters of type classes we're trying to instantiate with `NatWith`. I've made this work by brute force [here](https://github.com/milessabin/shapeless/commit/9ae0dcfeb56b032d780d3706c30747153b1db13c). I'll see if I can come up with something a little more general and then merge.

milessabin:
Closed via 9ae0dcfeb56b032d780d3706c30747153b1db13c and d6dd17528b545a9738e891b73cfc7128f56140f1.

AndreasKostler:
@milessabin Thanks with your help on this! :)

PR #472, opened 2015-09-29
Tiny documentation fix: use Record instead of Union in the Record example

liff:
I suppose the scaladoc may have been partly copied from `Union`?

### DIFF
--- a/core/src/main/scala/shapeless/records.scala
+++ b/core/src/main/scala/shapeless/records.scala
@@ -51,8 +51,8 @@ object record {
    * named argument syntax. Values of the type just defined can be created as follows,
    *
    * {{{
-   * val xyz = Union(x = 23, y = "foo", z = true)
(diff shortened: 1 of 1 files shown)


ceedubs:
:+1: thanks!

alexarchambault:
:+1:

alexarchambault:
Wait :-)
Next line needs fixing too (returns just `"foo"`, not `Some("foo")`)

alexarchambault:
@liff We should just have the doc *not* use `get`, it should be directly `xyz('y)`.

Maybe squash the commits too :-)

liff:
Ok, no problem

liff:
Sorry about the possible spam. Squashed now :)

alexarchambault:
Thanks :-)

:+1: for the current version

ceedubs:
:+1:

PR #473, opened 2015-10-04
Replace global variants of Lazy / Strict by Cached

alexarchambault:
This PR replaces `Strict.Global` and `Lazy.Global` from https://github.com/milessabin/shapeless/pull/435 by a more explicit type class, `Cached`. It is not necessarily summoned by `Lazy` / `Strict`, so it's put it at the root of the `shapeless` namespace.


milessabin:
LGTM ... merging ... many thanks :-)

Issue #474, opened 2015-10-09
Default type class broken for top-level classes


cvogt:
@alexarchambault playing around with the Default type class. Great job. Only it doesn't work for top-level classes for me :). If you move `CC` in the test out of DefaultTestDefinitions into the top-level it produces HNil instead of the record.

alexarchambault:
I used this type class quite a bit, and somehow managed not stumble upon this! Good catch :-)

https://github.com/milessabin/shapeless/pull/475 should fix it.

cvogt:
Sweet, thx :)!

PR #475, opened 2015-10-09
Fix Default missing some default values

alexarchambault:
Fixes https://github.com/milessabin/shapeless/issues/474

### DIFF
--- a/core/src/test/scala/shapeless/default.scala
+++ b/core/src/test/scala/shapeless/default.scala
@@ -5,6 +5,11 @@ import shapeless.record.Record
 import org.junit.Test
 import shapeless.test.illTyped
 
+// Intentionally defined as a top-level class - (compile time) reflection API not behaving
+// the same way compared to definitions in a singleton, like CC below.
+// See https://github.com/milessabin/shapeless/issues/474
+case class DefaultCC(i: Int, s: String = "b", flagOpt: Option[Boolean] = Some(true))
+
 object DefaultTestDefinitions {
 
   case class CC(i: Int, s: String = "b", flagOpt: Option[Boolean] = Some(true))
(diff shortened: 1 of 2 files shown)


ceedubs:
@alexarchambault should these tests be using `assertTypedEquals` instead of ascribing the type (which could influence inference)?

I assume that the return types can be inferred.

alexarchambault:
Good point, it'd be better to ensure we get the right type without influencing inference.

I'm not too keen about using `assertTypedEquals[A]` for that though, as it only checks that the type under scrutiny is `<: A`, not that it is `=:= A`. I'd like to be sure that the `Some` parameters aren't literal ones, e.g. I want `Some[String]`, not ``Some[Witness.`"b"`.T]`` (note that this point is not tested currently, just saying it would be better like this).

There's a [`equalInferredType` function](https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/hlist.scala#L450) that does just that. So here, I think I'd prefer to devise a `assertSameTypeEquals`, ensuring that the inferred types of both of its arguments are the same (and not that they are just `<:` to a common super type).

ceedubs:
I just recalled that I think we are reliving a past conversation :) https://github.com/milessabin/shapeless/pull/97#issuecomment-45450057

This PR seems to fix the much bigger problem of default arguments not working for top-level classes, so I'd hate for it to get hung up on being finicky about testing the exact inferred types. I say :+1: for either merging it now or with `assertTypedEquals` and/or `equalInferredType`, and we can always revisit our typed equal test functions.

alexarchambault:
@ceedubs I remember https://github.com/milessabin/shapeless/pull/97 :-) I just used `assertTypedEquals` in the tests of `Default`, so that it would easier to revisit them later.

ceedubs:
:+1:

alexarchambault:
Who else could :+1: that... :smile: @travisbrown ?

core/src/test/scala/shapeless/default.scala
```diff
@@ -4,6 +4,12 @@ import shapeless.record.Record
 
 import org.junit.Test
 import shapeless.test.illTyped
+import shapeless.testutil.assertTypedEquals
+
+// Willingly defined in the shapeless scope - (compile time) reflection API not behaving the same way
```
travisbrown: This is nitpicky, but the important thing is that this is a top-level class, not that it's in the `shapeless` package.

Even more nitpicky: "intentionally" would be more idiomatic than "willingly".

travisbrown:
:+1:, looks good to me!

alexarchambault:
Changed the above comment following @travisbrown advice.

As the last change is rather minor, I'll consider this good to merge in a few hours unless anyone objects.

alexarchambault:
(Forced push a change in a test method name (now `topLevel`), that should have been done along with the previous change.)

milessabin:
Thanks folks :-)

PR #477, opened 2015-10-20
Emit a proper warning when preventing caching of partial trees from Lazy/Strict

alexarchambault:
Just one small issue: the warning seems not get emitted in 2.11! (It is on the 2.10 branch.)
I actually can't get a warning to be emitted (with `c.warning(c.enclosingPosition, "...")`) in 2.11.


milessabin:
I think in 2.11.x the behaviour was changed to hide errors and warnings from implicit macros other than at the top level.

milessabin:
Thanks :-)

PR #478, opened 2015-10-20
Port Cached to 2.10.x

alexarchambault:
Features the same attachement / structural typing stuff as `Lazy` (that dates back before https://github.com/milessabin/shapeless/pull/435 / https://github.com/milessabin/shapeless/pull/447).


milessabin:
Many thanks :-)

Issue #476, opened 2015-10-11
Type classes supporting incomplete decoder derivation


travisbrown:
Last week @mpilquist asked about adding the [`Complement` type class](https://github.com/travisbrown/circe/blob/baca9f6f1d32e40157eb9b2a023218b723f1fde2/generic/shared/src/main/scala/io/circe/generic/util/Complement.scala#L58) from [circe](https://github.com/travisbrown/circe) to Shapeless. I'd love to see this happen (and the idea has @milessabin's stamp of approval), but I have a couple of questions about how it should happen.

The motivation for the type class is to support "partial" decoding instances. I originally described the idea [in this blog post](https://meta.plasm.us/posts/2015/06/21/deriving-incomplete-type-class-instances/), and we're currently using it in [circe](https://meta.plasm.us/slides/finagle/finch-and-shapeless/#9), some [Finch](https://github.com/vkostyukov/finch-101/blob/master/src/main/scala/i/f/workshop/finch/Todo.scala#L38) examples, etc., and there are reports that it's being used [in the wild](https://gitter.im/milessabin/shapeless?at=55a861f8b83437005acb5a26) outside of the context of circe as well.

The implementation uses two type classes, `Insert` and `Complement`. `Insert` is essentially just a variation of `ops.hlist.Remove` that supports the unlabeled selection of an element in a record:

```scala
import io.circe.generic.util._
import shapeless._, record.Record, labelled.FieldType

type Rec = Record.`'foo -> Int, 'bar -> String, 'baz -> Double`.T

val labelled = Insert[Rec, FieldType[Witness.`'foo`.T, Int]]
val unlabelled = Insert[Rec, Int]
```

This is necessary to support the desired usage provided by `Complement`.

`Complement` is similarly similar to `ops.hlist.RemoveAll` but with unlabeled selection.

I have two questions:

1. Where should these go? I'm inclined to say `ops.record`, and for my needs they only ever need to work on records, but the implementation is more generic and works just fine on non-record `HList`s.
2. What should they be named? If they go in `ops.record` they could just be called `Remove` and `RemoveAll`, but they don't really work the same as the other record type classes (in that they're not about selection by keys), and those names would make the relationship with `ops.record.Remover` kind of unclear. I like `LabelAgnosticRemove(All)`, but that's a little unwieldy.

Any opinions? If nobody feels strongly I'll submit a PR with `ops.record.Remove(All)`.

milessabin:
[...]

PR #480, opened 2015-10-22
Add new Remove and RemoveAll type classes for records

travisbrown:
As discussed in #476. I've not added any syntax for these type classes, since `rec.remove(...)` would collide with the method for the `Remover` type class. I could add syntax if people think it would be useful and if we can come up with acceptable names.

I have [a branch of circe](https://github.com/travisbrown/circe/commit/564528fd05c8b9b0617ef89dde5d2207c94cf4d7) that uses this change instead of its own `Complement` type class.

### DIFF
--- a/core/src/main/scala/shapeless/ops/records.scala
+++ b/core/src/main/scala/shapeless/ops/records.scala
@@ -254,6 +254,94 @@ package record {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


milessabin:
Many thanks ... merging :-)

Issue #432, opened 2015-07-15
Provide wide types of singleton ones


alexarchambault:
Opening this issue to keep track of and discuss https://gitter.im/milessabin/shapeless?at=55a2ab1b4962d17932cda3c2.

@milessabin Are you sure you want to add a type member to `Witness`, like
```scala
trait Witness extends Serializable {
  /** Wide type of `value`, e.g. just `String` for a singleton string, or `Int` for a singleton integer */
  type W

  type T <: W
  val value: T {}
}
```

That would make the current `Witness.Aux` missing `W`, unless it is added to it like
```scala
type Aux[W0, T0 <: W0] = Witness { type W = W0; type T = T0 }
```
which would in turn break compatibility with previous versions. (Plus it would just be ignored most of the time, for the current uses of `Witness`.)

Couldn't an additional type class like
```scala
trait Widen[T] {
  type W >: T
}
```
materialized by a macro, be better, with regard to compatibility at least?

alexarchambault:
Another possibility could be to keep the wide type as a type parameter, like
```scala
trait Witness[W] extends Serializable {
  type T <: W
  val value: T {}
}

object Witness {
  type Aux[W, T0 <: W]  = Witness[W] { type T = T0 }
  // ...
}
```
This would break binary compatibility too.

PR #481, opened 2015-10-22
Add Widen type class

alexarchambault:
Fixes https://github.com/milessabin/shapeless/issues/432.

Note that, as a peculiarity, there's no `Widen` instance available for types of singleton objects (like `case object O`), as these can't really be widen to another super type. This might be changed, but I'm fearing this would be inconvenient when using this to derive type classes for them (as type classes both as a singleton and as an empty case class through `Generic` would be available).


milessabin:
Many thanks ... merging :-)

PR #485, opened 2015-10-26
Make Cached behave more like Strict.Global it replaced

alexarchambault:
Fixes https://github.com/fommil/spray-json-shapeless/issues/14

I wasn't able to reproduce the issue as is in the tests here, although it blows up quite easily from external projects.

I have no idea *why* this fixes it. It just makes `Cached` look even more similar to `Lazy.Global` and `Strict.Global` it replaced, and that were working perfectly fine.


milessabin:
I would guess that the key change here is the explicit type argument on the `Cached` constructor. With it, the singleton typed-ness of `actualTpe` will be perserved; without it the type would be widened.

Anyhow LGTM ... many thanks ... merging :-)

PR #488, opened 2015-10-31
Introduced FieldType constructor on Witness

pchlupacek:
Just added simple addition to Witness syntax to allow arbitrary Union and Record Types. Modified Union tests to take benefits of this syntax. 
```scala
//before
Union.`'a -> Int, 'b -> String, 'c–>Double`.T
// now
Witness.`'a`.->>[Int] :+:  Witness.`'a`.->>[Int]  :+: Witness.`'a`.->>[Int]  :+: CNil
// -- or --
Witness.`'a`.Field[Int] :+:  Witness.`'a`.Field[Int]  :+: Witness.`'a`.Field[Int]  :+: CNil 

```

### DIFF
--- a/core/src/main/scala/shapeless/singletons.scala
+++ b/core/src/main/scala/shapeless/singletons.scala
@@ -181,8 +181,14 @@ trait SingletonTypeUtils extends ReprTypes {
   def parseType(typeStr: String): Option[c.Type] =
     parseStandardType(typeStr) orElse parseLiteralType(typeStr)
 
-  def typeCarrier(tpe: c.Type) = {
-    val carrier = c.typecheck(tq"{ type T = $tpe }", mode = c.TYPEmode).tpe
+  def typeCarrier(tpe: c.Type) =
+    mkTypeCarrier(tq"{ type T = $tpe }")
+
+  def fieldTypeCarrier(tpe: c.Type) =
+    mkTypeCarrier(tq"{ type T = $tpe ; type ->>[V] = Field[V] ; type Field[V] = shapeless.labelled.FieldType[$tpe,V] }")
+
+  def mkTypeCarrier(tree:c.Tree) = {
+    val carrier = c.typecheck(tree, mode = c.TYPEmode).tpe
 
     // We can't yield a useful value here, so return Unit instead which is at least guaranteed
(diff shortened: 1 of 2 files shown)


alexarchambault:
About the changes in the tests, wouldn't it be better to keep the former tests (of the ``Union.`...`.T`` syntax), and only add tests of the new syntax along with it?

E.g. for the first test, I'm thinking of having something like
```scala
type U = Union.`'i -> Int, 's -> String, 'b -> Boolean`.T // former type, kept as is

// only additions
type U0 = Witness.`'i`.Field[Int] :+: Witness.`'s`.Field[String] :+: Witness.`'b`.Field[Boolean] :+: CNil
// ensures the new syntax gives the same result as the former one
implicitly[U =:= U0]
```

pchlupacek:
@alexarchambault yeah perhaps you right for shorter types the previous syntax is more concise. I'll update this PR.

milessabin:
Sorry if I'm being a bit dense, I can't quite see the benefits of this change ... as far as the examples of the tests are concerned, they offer the same functionality but are more verbose. So I agree with @alexarchambault that the existing tests should stay, but I'd also like to see something that demonstrates something new.

pchlupacek:
@milessabin I'll just write a simple test that proves U =:= U0. As discussed on gitter, particular benefit of this is that we don't re-implement parser of the compiler (in backticks) and this allows us more complex contstruct and multi-line record/union types. It is not only beneficial in tests but anywhere where you need explicitly specify the type, for example in serialization/deserialization

Do I understand correctly that PR is ok, you just want to see different tests ?

milessabin:
I really want to see a test of at least one compelling example that this PR enables that wasn't possible previously.

pchlupacek:
@milessabin ok understood

pchlupacek:
@milessabin @alexarchambault Just pushed change. LMK if the demonstration is ok or I need to put more examples....

pchlupacek:
@milessabin not sure if you had chance to review this? If there is alternative to what this tries to achieve you like better, can you please share? Thanks.

milessabin:
Yes, sorry about the hold up ... LGTM ... many thanks ... merging :-)

Issue #490, opened 2015-11-07
`HList.zipWithIndex` requires exponential compile time


jarreds:
Running the following in a repl against 2.3.0-SNAPSHOT you will see that compile appears to be exponential for ```zipWithIndex``` as the size of the ```HList``` increases:

```scala
import shapeless._
import scala.language.postfixOps

(0::1::2::3::4::5::6::7::8::9::10::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::16::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::16::17::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::16::17::18::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::16::17::18::19::HNil) zipWithIndex
(0::1::2::3::4::5::6::7::8::9::10::11::12::13::14::15::16::17::18::19::20::HNil) zipWithIndex
```

cc @AndreasKostler

PR #491, opened 2015-11-09
ZipWithIndex with O(n) implicit search complexity

alexarchambault:
Looks like no one is taking a cut at https://github.com/milessabin/shapeless/issues/490, so here I go :-)

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -2026,31 +2026,52 @@ object hlist {
    * ({element from input tuple}, {element index})
    *
    * @author Andreas Koestler
+   * @author Alexandre Archambault
    */
   trait ZipWithIndex[L <: HList] extends Serializable with DepFn1[L] {
     type Out <: HList
   }
 
   object ZipWithIndex {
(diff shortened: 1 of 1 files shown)


AndreasKostler:
LGTM :+1:
However, could you explain to me as to why the original version exhibited the slow behaviour while this version doesn't. We can take this offline if you like, just find me on gitter.

alexarchambault:
Maybe not a problem of some kind of "implicit search complexity" after all, but this implementation just more simple, not resorting to `Zip`, itself resorting to `Transposer` that I find hard to get one's head around to. I can easily call it on HLists of length ~70 (before seeing other unrelated scalac crashes).

alexarchambault:
Argh, merged by mistake :-(

alexarchambault:
Ping me to re-open :-|

jarreds:
@alexarchambault +1 on reopening. I'd like to use `zipWithIndex` but can't until something like this in place.

alexarchambault:
@AndreasKostler My changes make it look (almost) like the dual of what *you* did for coproducts.

AndreasKostler:
@alexarchambault I still have to get my head around as to why the exponential compile time happens. Neither `Range` nor `Zip` exhibit this behaviour and I can't see why `zipWithIndex` would...any light on this is highly appreciatetd

PR #494, opened 2015-11-10
Add ForEach to HList for stateful effect

ntnj:
Applies the given Poly function to each element of HList in sequence


ntnj:
I didn't see any other stateful functions like this in HList class, so I don't know if it's a good candidate to include in shapeless-core.
I needed this function for an HList, so I wrote my own.

pomadchin:
And what about Mapper, what is the difference except the Unit return type?

milessabin:
There's no blanket ban per se on statefulness in shapeless, however its use would need an extremely compelling justification over the alternatives. As @pomadchin observes, this is equivalent to  a `map` with a stateful `Poly`. An even better choice would be a stateless `foldLeft`. Can you use either of those in your context?

milessabin:
@ntnj any thoughts on my questions? Or should I close this?

PR #499, opened 2015-11-15
Have Default work with method local classes

alexarchambault:
See the added test, which doesn't pass without the other changes.

I had to cut-n-paste code from [macro-paradise](http://i2.kym-cdn.com/photos/images/original/000/234/739/fa5.jpg) to have this work.

### DIFF
--- a/core/src/main/scala/shapeless/default.scala
+++ b/core/src/main/scala/shapeless/default.scala
@@ -225,12 +225,23 @@ class DefaultMacros(val c: whitebox.Context) extends shapeless.CaseClassMacros {
     // Fixes https://github.com/milessabin/shapeless/issues/474
     tpe.typeSymbol.companion.info.members
 
-    def wrapTpeTree(idx: Int, argTpe: Type) = {
-      val method = TermName(s"apply$$default$$${idx + 1}")
+    def methodFrom(tpe: Type, name: String): Option[Symbol] = {
+      val m = tpe.member(TermName(name))
+      if (m == NoSymbol)
+        None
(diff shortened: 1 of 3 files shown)


milessabin:
A couple of thoughts on this ... is there any reason not to _always_ do this ... ie. remove the `patched` flag and make this the default behaviour of `companionRef`? I'm sure I've seen (and maybe even worked around) similar issues with method-local classes in other contexts.

The other thing is that I'm not exactly relishing the prospect of backporting this to 2.10.x ... would you mind taking that on this time? I'm guessing that you'll be able to find corresponding code on the 2.10.x branch of macro paradise.

milessabin:
Oh, could you also include the links to the [two macro-paradise issues](https://github.com/scalamacros/paradise/blob/c14c634923313dd03f4f483be3d7782a9b56de0e/plugin/src/main/scala/org/scalamacros/paradise/typechecker/Namers.scala#L566-L567).

alexarchambault:
@milessabin I enabled the "patched" companion ref by default, which doesn't seem to break anything. I also added the links to the two macro-paradise issues.

I couldn't find the equivalent method in the 2.10 branch of macro-paradise, but the one here can be easily ported to scala 2.10. Did on this branch: https://github.com/alexarchambault/shapeless/tree/topic/default-2.10.x

milessabin:
Awesome ... thanks :-)

I'll merge as soon as Travis goes green ...

milessabin:
Many thanks ... merging :-)

PR #501, opened 2015-11-17
Have Default work with method local classes (2.10.x)

alexarchambault:
Port of https://github.com/milessabin/shapeless/pull/499


milessabin:
That looks better! :-)

milessabin:
Many thanks ... merging :-)

PR #502, opened 2015-11-18
Type classes for HList Union, Intersection and Diff

OlivierBlanvillain:


core/src/main/scala/shapeless/syntax/records.scala
```diff
@@ -81,9 +80,17 @@ final class RecordOps[L <: HList](val l : L) extends AnyVal with Serializable {
    */
   def -[V, Out <: HList](k: Witness)(implicit remover : Remover.Aux[L, k.T, (V, Out)]): Out = remover(l)._2
 
+  /**
+   * Computes the union between this record and another record.
+   */
   def merge[M <: HList](m: M)(implicit merger: Merger[L, M]): merger.Out = merger(l, m)
```
OlivierBlanvillain: Both union and intersection are combining records, it might be better be rename this `def merge` to a `def union`, and `shapeless.ops.record.Merger` to `shapeless.ops.record.Unifier`?

OlivierBlanvillain:
I've just realised this could also be written directly on HList, the implementation does not uses the fact that it's record. It's also the case for `shapeless.ops.record.Merger`. What do you think about the following:?

- Move `shapeless.ops.record.Merger` to `shapeless.ops.hlist.Union`
- Add `shapeless.ops.hlist.Intersection`
- Add `shapeless.ops.hlist.Subtraction` (or `Except`? naming is hard...)

milessabin:
Yes, I think you're right that both of these operations can be defined on `HList`s ... give it a try. The existing `Merger` is used quite a bit, so it will have to stay for now.

I think `Union` and `Intersection` are good names for the ops. I think what you're calling `Subtraction` already exists as `RemoveAll`. It might be worth going through the existing `HList` ops to make absolutely sure that we don't already have something equivalent to `Intersection`.

OlivierBlanvillain:
`RemoveAll[L, M]` requires that `L` is a subset of `M`, my idea would be that `Subtraction` does not have this constraint to model what EXCEPT does in SQL. I have a use case for something like this:

    case class Foo(a: Int, b: Int)
    case class Bar(b: Int, c: Int)
    case class FooMinusBar(b: Int)

    Foo(1, 2).minus(Bar(3, 4)) == FooMinusBar(2).toRecord

It's called `diff` on `scala.collection.Set`, maybe that's a better name?

milessabin:
Yes, `Diff` would be fine.

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -1142,6 +1142,130 @@ object hlist {
...
+    implicit def hnilDiff[L <: HList]: Aux[L, HNil, L] =
+      new Diff[L, HNil] {
+        type Out = L
+        def apply(l: L): Out = l
+      }
+
+    implicit def hconsDiff2[L <: HList, LT <: HList, H, T <: HList]
+      (implicit
+        r: Remove.Aux[L, H, (H, LT)],
+        d: Diff[LT, T]
```
OlivierBlanvillain: Note: recusing on `Diff[LT, H :: T]` instead of `Diff[LT, T]` would work as well. With `Diff[LT, H :: T]`, all duplicate types `A` of `L` are removed if `A` is present in `M`, while with `Diff[LT, T]`, at most one type of `L` is removed for each type of `M`. The same consideration applies to `Intersection`.

OlivierBlanvillain:
I tried to handle duplicates the same way they are handled on `scala.collection.immutable.List.{ union, intersect, diff }`, and reused the terminology of the `List` documentation.

After going thought all the HList type classes I'm pretty sure these are not implemented already, but they are quite similar to what's already there:  `Diff` is `RemoveAll` without the sublist constraint, `Intersection` is `SelectMany` also without the sublist constraint, and `Union` is `Merger` for `HList`s.

core/src/main/scala/shapeless/syntax/records.scala
```diff
@@ -95,7 +98,7 @@ final class RecordOps[L <: HList](val l : L) extends AnyVal with Serializable {
   def keys(implicit keys: Keys[L]): keys.Out = keys()
 
   /**
-   * Returns an `HList` of the values of this record.
+   * Returns a `HList` of the values of this record.
```
milessabin: It was right the first time ;-)

milessabin:
Great stuff! Many thanks ... merging :-)

OlivierBlanvillain:
Thanks!

PR #503, opened 2015-11-20
Typeclass to verify all elements of an hlist belong to a typeclass

triggerNZ:
This was suggested by @mpilquist on gitter. I found it useful, so thought I would PR it in. It is my first shapeless pull request, so any feedback is appreciated.

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -2636,21 +2636,16 @@ object hlist {
     def apply[F[_]] = new Curried[F]
     def apply[F[_], In <: HList](implicit ev: LiftAll[F, In]) = ev
 
-
-
-
     implicit def hnil[F[_]]: LiftAll.Aux[F, HNil, HNil] = new LiftAll[F, HNil] {
       type Out = HNil
       def instances = HNil
     }
 
-    implicit def hcons[F[_], H, T <: HList](implicit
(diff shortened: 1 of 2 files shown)


milessabin:
Needs rebasing now ...

Would you mind reformatting to be more consistent with the rest of the source tree (essentially two space indent and break lines agressively).

Otherwise looks good. I'm not completely convinced by the name ... it's really something more like `Forall`, but I'm wary of using a name which is so generally useful. Also, we could have a similar definition for `Coproduct` which might also be useful.

AlecZorab:
Also very useful, for whatever name you decide to go with, would be 

```
    def apply[F[_]] = new TypeclassInstancesAux[F]
    class TypeclassInstancesAux[F[_]] {def apply[In <: HList](in: In)(implicit ev: TypeClassInstances[F, In]) = ev}

    def apply[F[_], In <: HList](implicit ev: TypeClassInstances[F, In]) = ev
```

which I'm helpfully leaving in a comment instead of PRing @triggerNZ's branch because *lazy*

triggerNZ:
@milessabin thank you for your comments. I did the rename. Are you okay with me doing the coproduct case in a seperate PR?

triggerNZ:
@AlecZorab That's a cool addition - put it in

milessabin:

milessabin:
Yes, please do submit another PR for the coproduct case.

PR #507, opened 2015-11-23
Remove WitnessWith.Out in 2.10.x.

dwijnand:
Fixed in 2.10.5 in https://github.com/scala/scala/pull/4303


milessabin:
Awesome ... thanks :-)

PR #511, opened 2015-12-11
Short-circuit HList / Coproduct implicit recursion via a macro-backed type class

alexarchambault:
Closed https://github.com/milessabin/shapeless/pull/459 by mistake, reopening it here.

### DIFF
--- a/core/src/main/scala/shapeless/derived.scala
+++ b/core/src/main/scala/shapeless/derived.scala
@@ -1,0 +1,125 @@
+package shapeless
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+abstract class Derived[+T] extends Serializable {
+  def value: T
+}
+
+object Derived {
+  def apply[T](t: T): Derived[T] =
+    new Derived[T] {
+      def value = t
+    }
+
+  implicit def derive[T]: Derived[T] = macro DerivedMacros.derive[T]
+}
+
+@macrocompat.bundle
+class DerivedMacros(val c: whitebox.Context) extends CaseClassMacros with OpenImplicitMacros {
+  import c.universe._
+
+  lazy val typeClassTpe = typeOf[TypeClass[Any]].typeConstructor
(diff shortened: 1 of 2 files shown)


alexarchambault:
Still need some work.

milessabin:
We could have a look at it during the Scala eXchange hack day if you're around.

milessabin:
Is this redundant now?

alexarchambault:
It's not really redundant. It's something that short-circuits standard (recursive) derivation for case classes / ADTs, when using `TypeClass`.

I don't have much time for it right now, I think that can wait... (I can close and re-open it later if you'd like.)

milessabin:
I'll punt this to 2.3.1 ...

milessabin:
Do you think you'll have any time for this for 2.3.2?

alexarchambault:
I'd be willing to have a look at it, yes (some type class derivation slowness bite me quite hard very recently...). I'll update the PR in a few days.

alexarchambault:
Updated the PR.

I may try to make this mechanism more general in a second time (like https://github.com/milessabin/shapeless/pull/459 was at some point). So that it can handle default values or require extra auxiliary type class instances for each field / constructor - if you're fine with it.

alexarchambault:
I haven't tried to benchmark that in its current state (although there's no reason it doesn't make things better).

alexarchambault:
Note that it uses `Console.err.println` to explicitly report which field or constructor is making the derivation fail (which helps a great deal to debug derivations). `c.error` or `c.warning` are silenced here.

alexarchambault:
This seems to break the orphan tests, for reasons that are unclear to me... I don't much about their internals. Aren't they supposed to be deprecated somehow (by export-hook, or `LowPriority` here)? If that's the case, I'd rather wait for shapeless 2.4. I'm not sure there's a real problem in `Derived` here (and in any case, that can be more thoroughly tested in the mean time).

milessabin:

milessabin:
I'm going to close this for now. I don't want to add new macros to shapeless at this point and I think that `-Yinduction-heuristics` is the better way forward.

PR #514, opened 2015-12-15
Replace `Priority` by `LowPriority`

alexarchambault:
`LowPriority[T]` allows to give a lower priority to implicits, e.g. in
```scala
implicit def derivedTC[T](implicit ev: LowPriority[TC[T]]): TC[T] = ...
```

`derivedTC` will only be available for orphaned `T`, that is it will give the priority to the already existing `TC[T]` else - effectively having a lower priority. `LowPriority[TC[T]]` acts like some kind of poison pill, refusing to materialize if a `TC[T]` already exists elsewhere.

It deals with some interplay with `Lazy`, that is, if the `TC[T]` is derived through it, `Lazy` would make the instance that is being built available to us straightaway. This has to be somehow prevented when `LowPriority` looks for instances of `TC[T]` "elsewhere" - that's why `LowPriority` is a Lazy extension.


milessabin:
Hi Alex! I want to hold off on this until we have the single branch build up and running ... hopefully not too long now :-)

alexarchambault:
Superseded by https://github.com/milessabin/shapeless/pull/536.

PR #515, opened 2015-12-18
Add macro-compat to enable Scala 2.10 in master

dwijnand:
Required the creation of a compat.scala for 2.10/2.11+ for exceptions.

---

Given this depends on macro-compat version "1.2.0-SNAPSHOT", Travis will fail. Perhaps after an initial review, we can cut some kind of release of macro-compat and I'll update this pull request.

As always, I'm happy to rework even minor things, so feel free to review aggressively.

For a bit of scary trivial, `git grep` tells me that's 33 instances of "BACKPORT" added..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 
 scala:
+- 2.10.6
 - 2.11.7
 - 2.12.0-M3
 
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ being [good candidates to take on][lowhangingfruit]. No contribution is too smal
 ## Using shapeless
 
 Binary release artefacts are published to the [Sonatype OSS Repository Hosting service][sonatype] and synced to Maven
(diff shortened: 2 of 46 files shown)


dwijnand:
Also noticed that there's some branch name logic to be removed, but I guess that can wait for after the merge:
https://github.com/milessabin/shapeless/blob/4df4eb1853f4b030369509d9fe37e6e7ed637c27/.travis.yml#L17

dwijnand:
* rebased onto current master
* recovered a bunch of tests with cross-version sources
* tweaked the "compat.scala"s a little
* split into 2 separate commits

Now there are only 16 "BACKPORT"s and 7 (of my own) "Note:"s (these last in the tests).

Ready for review and merge.

dwijnand:
Update copyright notices of touched files.

For files created off of existing files which had copyright notices, I've kept such notices but set the years to "2016", even if the origin existed before that. Not sure if that's right or not.

milessabin:
Awesome work ... many thanks!

PR #517, opened 2015-12-19
Add padTo method to HList

ryoppy:


milessabin:
Great stuff!

Can you say why you've chosen to have pad with a length shorter than the unpadded list length return the list unchanged rather than being a compile time error? My gut feeling is that it would be more useful if when you pad to length _n_ you are guaranteed that the length of the resulting list is exactly _n_.

ryoppy:
thank you for your comments!
I modified to return list has length exactly n.

milessabin:

milessabin:
Could you send me another PR adding your details to the `CONTRIBUTORS` file?

ryoppy:
thx!
I sent PR #518

milessabin:
Thank you! :-)

Issue #426, opened 2015-07-09
Add instructions for inclusion in a Maven build to README


milessabin:

milessabin:
Fixed in #521.

PR #521, opened 2015-12-28
Add Maven build details to README

hamishdickson:
This is to address issue #426 

I've tested these instructions out on Scala 2.11, but don't have Scala 2.10 installed - do 2.10 users also need to include the marco paradise plugin to their build (I can imagine they would)? If so, I'm happy to add that as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,27 @@ libraryDependencies ++= Seq(
 )
 ```
 
+shapeless is also available for projects using the Maven build tool. To include shapeless for projects using Scala
+2.11.x you should add the following dependency to your pom file,
+
+```xml
+<dependency>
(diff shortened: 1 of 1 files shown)


milessabin:
Wonderful! Thanks for digging in to this :-)

Yes, Scala 2.10.x users do need the macro paradise plugin included in their build ... if you could add that to the instructions as well that would be perfect.

hamishdickson:
My pleasure :)

I'll do some research and push an update

hamishdickson:
@milessabin I've updated this now with the paradise plugin and tested it locally :)

Let me know if you have any questions/comments

milessabin:
That's wonderful ... thanks! :-)

PR #523, opened 2015-12-30
Fully qualify types in quasiquotes

fthomas:
While browsing the sources I noticed some unqualified types in quasiquotes. They haven't caused any hygiene issues for me but I think those are potential bugs.


milessabin:
Good catch! Many thanks ... merging :-)

PR #524, opened 2016-01-02
Add Unwrapped typeclass for unwrapping in codecs

clhodapp:
This adds a new typeclass in order to support getting at the unwrapped thing inside of wrapper types (e.g. AnyVal wrappers and shapeless newtypes).

The key use case is codec generation, as most existing codec generators handle these types pretty terribly.

Check the included example for... well... an example

### DIFF
--- a/core/src/main/scala/shapeless/syntax/unwrapped.scala
+++ b/core/src/main/scala/shapeless/syntax/unwrapped.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2016 Miles Sabin 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shapeless
+package syntax
+
+object unwrapped {
+  implicit class UnwrappedSyntax[T](val t: T) extends AnyVal {
+    def unwrap[U](implicit uw: Unwrapped.Aux[T, U]): U = uw.unwrap(t)
+    def wrap[W](implicit uw: Unwrapped.Aux[W, T]): W = uw.wrap(t)
+  }
+}
--- a/core/src/main/scala/shapeless/unwrapped.scala
+++ b/core/src/main/scala/shapeless/unwrapped.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2016 Miles Sabin
+ *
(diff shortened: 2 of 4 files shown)


milessabin:
This looked really interesting ... why did you close it?

clhodapp:
It was totally an accident.

milessabin:
This is really fascinating ... it had never occurred to me that the `Generic` instance for a value class could be used to extract the underlying value. Extracting the value underlying a `NewType` is clearly evil, and none the worse for that.

I'm resisting the urge to merge this immediately, because I'd like to know if you've thought about how this might relate to GHC's [generalized newtype deriving](https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/deriving.html) (see section 7.5.5) and, if there _is_ a useful relationship, if we could make that more apparent.

clhodapp:
So... there are a few differences between this and what happens in Haskell. First, as I'm sure you know, the answer to "who controls typeclass derivation" is generally different in Scala vs Haskell. In Haskell, the data type implementor opts into derivation by adding a "deriving" clause to indicate that they want instances of some set of typeclasses to be derived for their datatype. In Scala, the answer is generally a bit different (for derived instances, at least). In Scala, generally, derivation is controlled by either the typeclass implementor (by putting the derivation machinery in the companion object) or by the end user (by importing the derivation machinery).

In both Haskell and Scala, derivation is can be a bit of a cooperation between users and the compiler. The compiler implements the ability to "materialize" instances for us to cobble to together to build our own instances. In Hasekell's case, the compiler itself has an understanding of wrapper types vis-a-vis derivation. For Scala, though, the compiler doesn't understand wrapper types in this context, so we end up having to do the footwork to add support for unwrapping explicitly in the machinery for each and every typeclass we want to support unwrapping for. There are some substantial drawbacks to this. First, because unwrapping support is baked into the derivation machinery itself, it's all-or-nothing: you either support unwrapping all wrapper types or none (or you have to build a more complex derivation system with special cases for things you don't want (un)wrapped). Second, we end up with instances that, in practical terms, look like a slightly-worse version of the ones in the `7.5.5` intro: we end up calling an effectively-identity function to "unbox" things that don't need to be unboxed. It's even worse for `AnyVal`s, because the compiler has really annoying expectations that basically cause it to generate code to box `AnyVal` wrappers whenever they get passed to or returned from functions (but unbox them inside). For these, you end up in a weird situation where the generated code always passes the Unwrapped instance the type of object it doesn't want (even if the correctly-classed object was just sitting around, theoretically waiting to be cast to the right thing).

All this thought about efficiency (and especially thinking about what's in Haskell for this) made me feel sad, so I made a pass at rewriting everything to use macros (mostly a dead end) and another where I set up the materializers to cast a single shared object to the correct type where possible (if only Scala had vals with type parameters... sigh).

milessabin:
OK, thanks for that ... are you happy for it to be merged as it is now or would you like to do some more work on it first?

clhodapp:
I'm happy for it to be merged. Someday, it might make sense to write macros to help with converting (via casting or creating new wrapping instances) instances of other typeclasses, but that could happen as its own feature down the road.

milessabin:
Okey doke ...

Anyhow this LGTM ... merging :-)

clhodapp:
Yay! Thanks

PR #525, opened 2016-01-03
Remove path-dependent types in ops.hlists.Fill

OlivierBlanvillain:
This fixes the issue I discussed today on Gitter, which is that the following code compilation time is exponential in the number of `Int`s in `HListOfInts`:

```
import shapeless.{HList, ::, HNil, Nat}
import shapeless.ops.hlist.{Length, Fill}

object Example {
  trait Exp[L]

  implicit def mkExp[L <: HList, N <: Nat]
    (implicit
      l: Length.Aux[L, N],
      f: Fill[N, Any]
    ): Exp[L] =
      new Object with Exp[L]

  type HListOfInts =
    Int :: Int :: Int :: Int :: Int :: Int :: Int :: Int :: 
    Int :: Int :: Int :: Int :: Int :: Int :: Int :: Int :: 
    Int :: Int :: Int :: Int :: HNil

  val a = implicitly[Exp[HListOfInts]]
}
```

Thank you @alexarchambault for the solution :smile:

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -2529,7 +2529,7 @@ object hlist {
   trait Fill[N, A] extends DepFn1[A] with Serializable { type Out <: HList }
 
   object Fill {
-    def apply[N, A](implicit fill: Fill[N, A]) = fill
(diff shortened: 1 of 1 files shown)


milessabin:
LGTM ... Many thanks ... merging :-)

Issue #390, opened 2015-05-20
`cachedImplicit` fails for `LabelledGeneric` when used inside companion object


notxcain:
Here is the code
```
import shapeless._

sealed trait Root
case class A(s: String) extends Root
case class B(i: Int) extends Root

object Root {
  implicit val lgen: LabelledGeneric[Root] = cachedImplicit //fails
  implicit val gen: Generic[Root] = cachedImplicit //compiles
}
```

milessabin:
I think that part of the problem here is that the type annotations are actually incorrect: in both cases they're explicitly discarding the refinement, ie. `type Repr = ...` which is essential for these values to be usable.

There's no straightforward way of adding the refinement to the annotation, so I think that for now at least the unannotated, hence inferred, form will have to do,

```scala
object Root {
  implicit val gen = the[Generic[Root]]
  implicit val lgen = the[LabelledGeneric[Root]]
}
```

milessabin:
I'm going leave this open for further investigation, but given that there is a workaround I'll put this until the next release.

notxcain:
Ok, thanks for the workaround!

milessabin:
Closing this in favour of #508.

milessabin:
I have a fairly ugly workaround for the type annotations part of this issue,

```scala
import shapeless._, test._

case class Foo(i: Int, s: String)

object Foo {
  import Generic._ // This is the workaround from #508
  val g1 = { // nb. type of g1 must be inferred
    def fooGen = ??? // Here we shadow the following defn so it isn't used in the Generic below
    Generic[Foo]
  }

  // Now we can define both the value and the type of fooGen in terms of the
  // non-implicit g1 ...
  implicit val fooGen: Generic.Aux[Foo, g1.Repr] = g1 
}

object Demo extends App {
  val gen = Generic[Foo]
  val foo = Foo(23, "foo")
  val fooRepr = gen.to(foo)
  typed[Int :: String :: HNil](fooRepr)
  assert(fooRepr.head+13 == 36)
}
```

notxcain:
I see what you did here :) And yes it's pretty ugly. I like where it goes in #527

milessabin:
Reopening because it's about to be fixed properly :-)

milessabin:
As of 49bb32ff5051b9b2650b37106d9925b7b92fcb43 the following should work as expected,

```scala
import shapeless._

sealed trait Root
case class A(s: String) extends Root
[...]

PR #527, opened 2016-01-10
Fix #508: Perform a customized implicit search

### LINKED ISSUE #508: cachedImplicit does not work when used in companion
Consider the following:
```scala
scala> :paste
import shapeless.cachedImplicit

trait Foo[A]
object Foo {
  implicit def materialize[A]: Foo[A] = new Foo[A] {}
}

case class Bar(x: Int)
object Bar {
  implicit val foo: Foo[Bar] = cachedImplicit
}

Bar.foo

import shapeless.cachedImplicit
defined trait Foo
defined object Foo
defined class Bar
defined object Bar
res0: Foo[Bar] = null
```
If instead of defining `Bar.foo` as a val, we define it as a def, we get a stack overflow exception.

This occurs because the implementation of `cachedImplicit` attempts to hide the implicit definition by shadowing it's name, however, because the definition is in the `Bar` companion, scalac finds the definition when searching the companion.

clhodapp:
This changes cachedImplicit to cause it to perform a customized implicit search that is not allowed to find the owner. Unfortunately, it makes us depend on compiler internals a bit, but that's the price you pay for doing advanced things in macros, I think.

Question, though: should this be merged with `the`?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -160,10 +160,11 @@ lazy val scalaMacroDependencies: Seq[Setting[_]] = Seq(
 lazy val crossVersionSharedSources: Seq[Setting[_]] =
   Seq(Compile, Test).map { sc =>
     (unmanagedSourceDirectories in sc) ++= {
-      (unmanagedSourceDirectories in sc ).value.map { dir: File =>
+      (unmanagedSourceDirectories in sc ).value.flatMap { dir: File =>
         CrossVersion.partialVersion(scalaVersion.value) match {
-          case Some((2, y)) if y == 10 => new File(dir.getPath + "_2.10")
-          case Some((2, y)) if y >= 11 => new File(dir.getPath + "_2.11+")
+          case Some((2, y)) if y == 10 => List(new File(dir.getPath + "_2.10"))
+          case Some((2, y)) if y >= 11 => List(new File(dir.getPath + "_2.11+"))
+          case Some((2, y)) if y >= 12 => List(new File(dir.getPath + "_2.11+"), new File(dir.getPath + "_2.12+"))
         }
       }
     }
--- a/core/shared/src/main/scala/shapeless/package.scala
+++ b/core/shared/src/main/scala/shapeless/package.scala
@@ -134,7 +134,9 @@ package shapeless {
       // the thing we are enclosed in
       val sCtx = tCtx.makeImplicit(true)
 
-      trait ImplicitSearchCompat extends analyzer.ImplicitsContextErrors { self: analyzer.ImplicitSearch =>
+      // 2.12 changed the signature on a method we need to override. In order to allow us to define
+      // both versions below with the "override" modifier, we need an interface that contains both.
+      trait CompatImplicitSearch extends analyzer.ImplicitsContextErrors { self: analyzer.ImplicitSearch =>
         import analyzer.global.{Type, Tree}
         import analyzer.{ImplicitInfo, Context}
         def AmbiguousImplicitError(
@@ -158,7 +160,7 @@ package shapeless {
         isView=false,
         context0=sCtx,
         pos0=c.enclosingPosition.asInstanceOf[global.Position]
-      ) with ImplicitSearchCompat {
+      ) with CompatImplicitSearch {
 
         import analyzer.global.{Type, Tree}
         import analyzer.{ImplicitInfo, Context}
@@ -193,11 +195,10 @@ package shapeless {
               s"""| $pre1 ${info1.sym.fullLocationString} of type ${info1.tpe}
                   | $pre2 ${info2.sym.fullLocationString} of type ${info2.tpe}
                   | $trailer""".stripMargin.trim
-            val msg = s"ambiguous implicit values:\n${coreMsg}match expected type $pt"
+            val msg = s"ambiguous implicit values:\n${coreMsg}\nmatch expected type $pt"
             c.abort(c.enclosingPosition, msg)
           }
         }
-
       }
 
       val best = is.bestImplicit
--- a/core/shared/src/main/scala/shapeless/package.scala
+++ b/core/shared/src/main/scala/shapeless/package.scala
@@ -115,27 +115,58 @@ package shapeless {
   class CachedImplicitMacros(val c: whitebox.Context) {
     import c.universe._
 
-    def dropLocal(nme: TermName): TermName = {
-      val LOCAL_SUFFIX_STRING = " "
-      val TermName(nmeString) = nme
-      if(nmeString endsWith LOCAL_SUFFIX_STRING)
-        TermName(nmeString.dropRight(LOCAL_SUFFIX_STRING.length))
-      else
-        nme
-    }
-
     def cachedImplicitImpl[T](implicit tTag: WeakTypeTag[T]): Tree = {
+      val casted = c.asInstanceOf[reflect.macros.runtime.Context]
+      val typer = casted.callsiteTyper
+      val global: casted.universe.type = casted.universe
+      val analyzer: global.analyzer.type = global.analyzer
+      val tCtx = typer.context
+      val owner = tCtx.owner
       val tTpe = weakTypeOf[T]
-      val owner = c.internal.enclosingOwner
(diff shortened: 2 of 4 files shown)


clhodapp:
Definitely don't merge this yet. I still need to check/fix some things. Feel free to discuss, though.

milessabin:
I've been experimenting with the idea of manipulating implicit search directly as you've done here, so I'm very sympathetic to the general approach.

clhodapp:
OK. I've done some work on resilience / error reporting. I think it also fixes https://github.com/milessabin/shapeless/issues/390 now, as we now return the *exact* tree we found from `cachedImplicit`, so you can pass things like `LabelledGeneric[Root]` to `cachedImplicit` through its type parameter and get back the fully refined type for the LHS to infer. The results are a bit weird for that example, though, because `LabelledGeneric[T] <:< Generic[T]`, so `gen` ends up finding `lgen` as an implicit `Generic[Root]` in `Root`'s companion object (and so it uses it).

clhodapp:
Ok. So that leaves the following two questions for you to answer:

1. Does it make sense for `the` and `cachedImplicit` to continue existing as separate entities, or should `the` just be modified to work this way?
2. What type of testing do you think this needs to go through and how much?

clhodapp:
OK. So, as far as I can tell, this is good to go. The only major problem that I know about is that it doesn't show custom ambiguous implicit messages in 2.12, which I don't think there's an easy way to do without having 2.12-specific code.

dwijnand:
Is there no way to assert a negative test? With illTyped or a richer illTyped?

clhodapp:
I didn't try! It may work! Let me find out...

clhodapp:
Basically, I reasoned that I couldn't test that it'd show my desired error messages and gave up... but there's no reason that I can't at least assert that things are e.g. ambiguous or not found.

dwijnand:
Oh that's what I mean by a richer illTyped - one that includes capturing error messages.

clhodapp:
Ah, so I do think that that is doable as a thing to make if it's wanted... maybe illTyped could be made to start returning a value to enable this...

dwijnand:
Currently illTyped matches the TypecheckException message against a given string: https://github.com/milessabin/shapeless/blob/shapeless-2.2.5/core/src/main/scala/shapeless/test/typechecking.scala#L51-L53

clhodapp:
Whoa! How did I not know that?!? Ok. I'll add tests tomorrow.

clhodapp:
As an update, I now have some issues to sort out. I don't think I'm blocked, I just need to get to it. @milessabin how urgent is this (how eager are you to release a shapeless 2.3 and how much is this a bottleneck?)

milessabin:
I'd very much like this to appear in 2.3.0 ... do you think you'll be done before the end of next week?

clhodapp:
Yes. I certainly will. But it's likely that I won't get to it until at least the weekend.

clhodapp:
OK. I'm close. I just have one more thing to do: supporting 2.12 custom ambiguous implicit messages, which I believe I know how to do without resorting to variant sources.

clhodapp:
So, after poking a bunch more (and actually getting things working), I actually think that custom ambiguous implicit methods in 2.12 are broken: https://gist.github.com/clhodapp/d74b050d7f046ccbe3b3. This doesn't meet the description in the doc, I think, as the compiler is choose to show _any_  of the defined custom ambiguous implicit messages, but it has to show _one of them_. I believe that this happens because it actually gives up as soon as it finds any ambiguous implicits and only checks those two for annotations. I'm actually pretty confused about the entire design decision of putting the annotations on the implicit values and not on the typeclass, I will say.. In any case, I am going to raise this on scala-internals, I think. Would it be OK to just not support these until these in 2.12 until issues are ironed out??

clhodapp:
For what it's worth, I actually do have them "working" exactly as they do in the compiler now if you just want that code, it's just that if the compiler behavior changes, we would need to change this code to work the same way.

milessabin:
Do you think those changes would be likely to break binary compatibility of shapeless for < 2.12.x? 2.12.x is less of a concern given that it isn't released yet.

clhodapp:
The way I'm doing it now, I don't think it will break binary compatibility to add it later, as long as the compiler itself doesn't break binary compatibility (which it is totally allowed to do: these are milestones, after all)

milessabin:
Sure, but hopefully we'll have no variant source, so changes to accommodate fixes in 2.12.x would entail changes for all versions ... if that doesn't have a effect on shapeless binary compatibility for < 2.12.x that's totally fine.

clhodapp:
To be clear, "< 2.12.x" do you mean "< 2.12.0"?

clhodapp:
Oh "< 2.12.x" sorry, I misread. I think probably I should go to bed now and pick this up in the morning. I'll push a branch of what I have for ambiguous implicit stuff right now just in case you want to take a look first, though.

clhodapp:
https://github.com/clhodapp/shapeless/compare/feature/cached-implicit-doesn't-find-self...clhodapp:custom-ambiguous-implicit-messages?expand=1

clhodapp:
By the way, it seems that my earlier gist actually shows something I didn't even _realize_ was the case: Apparently, there are issues with custom ambiguous implicit message annotations on vals. That said, the issue I was _trying_ to highlight still does exist: https://gist.github.com/clhodapp/5b85993c524670896705

Anyway, I really am going to sign off for now.

clhodapp:
https://groups.google.com/forum/#!topic/scala-internals/yDY8AWUpjSA

clhodapp:
Still no activity on my scala-internals thread.. @milessabin I propose that you decide if you'd rather have `@ambiguousImplicit` support that mirrors the imo-broken support in 2.12.0-M3 in shapeless 2.3.0 or if you'd rather just not support `@ambiguousImplicit` for `cachedImplicit` in 2.3.0. Whichever you opt to go with, I believe we will be able to add support for a fixed version of `@ambiguousImplicit` in a subsequent shapeless version without breaking binary compatibility.

milessabin:
Let's leave out for now ... there's no pressure until 2.12.0 is out and people start using it.

BTW, did you open a ticket for the 2.12 bug? If not I think you should.

clhodapp:
I will open one or more tickets, but the problem is that I think that the current implementation is not just buggy but actually misdesigned, so I wanted to discuss the design before filing the tickets.

clhodapp:
Anyway, if you are fine with leaving `@implicitAmbiguous` support out, I think this should be good to go, pending review.

core/src/main/scala/shapeless/package.scala
```diff
@@ -114,28 +114,111 @@ package shapeless {
...
+        else tTpe
+      }.asInstanceOf[global.Type]
+
+      // Run our own custom implicit search that isn't allowed to find
+      // the thing we are enclosed in
+      val sCtx = tCtx.makeImplicit(true)
+
+      // 2.12 changed the signature on a method we need to override. In order to allow us to define
+      // both versions below with the "override" modifier, we need an interface that contains both.
+      trait CompatImplicitSearch extends analyzer.ImplicitsContextErrors { self: analyzer.ImplicitSearch =>
```
milessabin: Could this be moved to macro-compat? Ideally we want shapeless to be written to just one version of the macro api/compiler internals and move all the version dependencies to macro-compat.

milessabin:
This is great stuff. Other than the comment about moving the 2.12.x specifics to macro-compat I think this is good to go :-)

milessabin:
Where are we with this now? I'm hoping for an RC1 before the end of the week and I'd like to get this in if at all possible.

milessabin:
I've tried just the first commit from this PR and it compiles and passes tests on Scala 2.10-12 ... is there any reason for not going with just that?

milessabin:
Merged as ffe50c8da741f1b0adf1436e0a17bd93ac076964, bf220958f90725074e925e3bc58633d9ffc96f56 and 49bb32ff5051b9b2650b37106d9925b7b92fcb43. Many thanks for leading the way on this :-)

clhodapp:
:smiley_cat:

PR #528, opened 2016-01-12
Drop c2 in CaseClassMacrosMixin (compat.scala)

dwijnand:
.. given that now macrocompat.CompatContext is-a scala.reflect.macros.runtime.Context


milessabin:
LGTM ... merging :-)

PR #532, opened 2016-01-14
Add ToSizedHList

tpolecat:
This adds a `ToSizedHList` type class witnessing that a stdlib collection can be turned into a uniform `HList` of a specific size.

```scala
scala> import syntax.std.traversable._
import syntax.std.traversable._

scala> List("fox", "banana", "penguin").toSizedHList(2)
res1: Option[shapeless.::[String,shapeless.::[String,shapeless.HNil]]] = None

scala> List("fox", "banana", "penguin").toSizedHList(3)
res2: Option[shapeless.::[String,shapeless.::[String,shapeless.::[String,shapeless.HNil]]]] = Some(fox :: banana :: penguin :: HNil)
```

It doesn't look like this can be serializable, because I use an `ops.sized.ToHList` which can't be made serializable because its implementation closes over a `scala.collection.generic.IsTraversableLike` which isn't serializable.

### DIFF
--- a/core/src/main/scala/shapeless/ops/sizeds.scala
+++ b/core/src/main/scala/shapeless/ops/sizeds.scala
@@ -9,7 +9,7 @@ object sized {
    *
    * @author Alexandre Archambault
    */
-  trait ToHList[-Repr, L <: Nat] {
+  trait ToHList[-Repr, L <: Nat] extends Serializable {
     type Out <: HList
     def apply(s: Sized[Repr, L]): Out
   }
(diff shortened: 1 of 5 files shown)


alexarchambault:
@tpolecat It should be possible to make it serializable, by putting in scope `IsTraversableLike` instances that are themselves serializable. This is done with `CanBuildFrom`s in https://github.com/milessabin/shapeless/commit/2ef958c65b8d0fd7d271860d294092ce8a36b608.

milessabin:
That would just be a change to the tests, right? Yes, it would be good to add that.

tpolecat:
Ok, I hacked around the `IsTraversableLike` in 69bf919.

milessabin:
In practice the same remedy is available to users, so I don't think that comment is warranted.

I'll merge anyway, but I think the comment should go.

PR #533, opened 2016-01-16
revise comment re: ToSizedHList serialization

tpolecat:
This removes unnecessary language from the serialization test for `ToSizedHList`, as suggested in the discussion for #532.


milessabin:
Thanks :-)

PR #534, opened 2016-01-17
Fix Nat.toInt

dwijnand:


milessabin:
Oh, blimey! Would those new tests have failed (to compile) previously?

dwijnand:
Yep, tested.

Seems no one used Nat.toInt (or has submitted a bug report) since I contributed it 13 months ago, #284. Broken since its inception throughout v2.1.0 - v2.2.5.

milessabin:
Ouch :-(

PR #535, opened 2016-01-19
Kill cross version sources

dwijnand:
No longer needed as of fe578ef8697dc86ccb78196543f02699ba4040bf


milessabin:
Hang on ... we still have cross version source for some tests!

dwijnand:
You're right, doh.

PR #536, opened 2016-01-24
Add LowPriority

alexarchambault:
Allows to give some implicits a lower priority than the existing ones.

Added to an implicit def like
```scala
implicit def genericThing[F, G]
 (implicit
   ev: LowPriority,
   gen: Generic.Aux[F, G],
   underlying: Thing[G]
 ): Thing[F] = ???
```
it prevents `genericThing` to provide an instance of `Thing[F]` if an implicit one is already available elsewhere. This effectively gives `genericThing` a lower priority than the already existing implicits - without having to deal with complex priority scopes.

This somehow fills the same gap as export-hook, but without the need for extra dependencies on both the type class and the deriver sides.

Things like `LowPriority.Orphan`, `LowPriority.Subclass`, `LowPriority.Algebraic`, `LowPriority.Instantiated`, etc. could even be added here, to achieve the same kind of detailed prioritization as export-hook - all externally, without external dependencies and an extra API to master.

### DIFF
--- a/core/src/main/scala/shapeless/lazy.scala
+++ b/core/src/main/scala/shapeless/lazy.scala
@@ -164,13 +164,13 @@ object Strict {
 }
 
 @macrocompat.bundle
-trait OpenImplicitMacros extends compat.CompatLite {
+trait OpenImplicitMacros {
   val c: whitebox.Context
 
   import c.universe._
 
   def openImplicitTpe: Option[Type] =
-    c.openImplicits.headOption.map(x => (x: ImplicitCandidate211).pt)
+    c.openImplicits.headOption.map(_.pt)
 
   def openImplicitTpeParam: Option[Type] =
     openImplicitTpe.map {
@@ -183,7 +183,7 @@ trait OpenImplicitMacros extends compat.CompatLite {
   def secondOpenImplicitTpe: Option[Type] =
     c.openImplicits match {
       case (List(_, second, _ @ _*)) =>
-        Some((second: ImplicitCandidate211).pt)
+        Some(second.pt)
       case _ => None
     }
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/shapeless/lazy.scala
```diff
@@ -164,9 +164,34 @@ object Strict {
...
 
 @macrocompat.bundle
-class LazyMacros(val c: whitebox.Context) extends CaseClassMacros {
+trait OpenImplicitMacros extends compat.CompatLite {
+  val c: whitebox.Context
+
+  import c.universe._
+
+  def openImplicitTpe: Option[Type] =
+    c.openImplicits.headOption.map(x => (x: ImplicitCandidate211).pt)
```
dwijnand: Given the implicit conversion https://github.com/milessabin/macro-compat/blob/7297f3152892ea9027fb45fdb2798710f52a7859/core/src/main/scala_2.10/macrocompat/compatcontext.scala#L390

you should be able to just call `x.pt` here.

milessabin: Yes, I'd like to avoid reintroducing any non-test variant source here. If there is a problem it would need to be fixed in macro-compat.

alexarchambault: Fixed that in https://github.com/milessabin/shapeless/commit/455412dd955a5a505b8727538323c17ad34360e2. I'll squash when everything's fine.

milessabin:

I'm going to merge this now, with the caveat that I think we're going to need to try and pull all of this stuff (`Lazy` included) into Typeclassic and/or Typelevel Scala.

alexarchambault:
@milessabin np to move this elsewhere!

PR #538, opened 2016-01-29
unwrap tag (tagged)

etaty:
I was trying to play with tagged type 
https://gitter.im/milessabin/shapeless?at=56ab8938dc33b33c7548a5a2

### DIFF
--- a/core/src/main/scala/shapeless/unwrapped.scala
+++ b/core/src/main/scala/shapeless/unwrapped.scala
@@ -17,6 +17,7 @@
 package shapeless
 
 import newtype._
+import shapeless.tag._
 
 trait Unwrapped[W] extends Serializable {
   type U
@@ -60,6 +61,12 @@ trait UnwrappedInstances extends LowPriorityUnwrappedInstances {
     chain: Strict[Unwrapped.Aux[UI, UF]]
   ) = chain.value.asInstanceOf[Unwrapped.Aux[Newtype[UI, Ops], UF]]
 
+  implicit def taggedUnwrapped[UI, T]: Unwrapped.Aux[UI @@ T, UI] =
+    new Unwrapped[UI @@ T] {
+      type U = UI
+      def unwrap(w: UI @@ T) = w.asInstanceOf[UI]
+      def wrap(u: UI) = tag[T](u)
+    }
 }
 
 trait LowPriorityUnwrappedInstances {
--- a/core/src/test/scala/shapeless/unwrapped.scala
+++ b/core/src/test/scala/shapeless/unwrapped.scala
@@ -110,6 +110,20 @@ class UnwrappedTests {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


clhodapp:
Can you add a test similar to https://github.com/etaty/shapeless/blob/features/tagged-unwrap/core/src/test/scala/shapeless/unwrapped.scala#L109 ? I'm concerned about whether it will pass.

etaty:
Yep good catch.
`test.illTyped("unwrapped: String @@ TestTag")` don't work
unless I write
`val unwrapped : String = wrapped.unwrap`

etaty:
that's weird, I am curious for the reason why it keeps the tagged type.

clhodapp:
It's because e.g. `String @@ TestTag` is the most-specific `T` that can be extracted from `T @@ TestTag`, so that's what the compiler chooses to infer (as strange as it is, `(String @@ TestTag) @@ TestTag` is actually the same as `String @@ TestTag` and the compiler is able to see that this is the case and reason from it). As far as I currently know, you actually need to be working in the context of a macro to actually strip tags off in the general case, such as what's needed here.

clhodapp:
Also, the version of this in your PR actually leaves off the maybe-important feature of being able to see through multiple wrappers over the same type. This is the purpose of all of those `chain` parameters in the other instance providers. Also, rather than constructing a new instance, you can take advantage of the micro-optimization I make in some of my other instances: Taking an `implicit chain` and simply casting it. This works because Unwrapped instances for e.g. tagged types don't have to actually change the object, they just need to make the types line up.

milessabin:
What's the status of this? I'm aiming for an RC1 before the end of the week and it'd be nice if we could get this in by then.

etaty:
Hi Miles,
I know nothing about macro.
And I did not realise that in my case, I did not need the `unwrap`. (I am also not sure it is a needed feature).
So I won't probably finish the pull-request by the end of the week.
(you can close/reject)

milessabin:
No worries ... if you'd like to work on this I'll punt it to 2.3.1?

milessabin:
Moved to 2.3.1 ...

clhodapp:
Just got around to checking how easy it is to define an untagger with Mile's higher order unification patch (https://github.com/scala/scala/pull/5102). It's pretty easy, I think: https://gist.github.com/clhodapp/a2f1b6fc011df9b60b107bd3034aef6d

clhodapp:
Actually, that technique seems to work even without higher order unification!

clhodapp:
https://gist.github.com/clhodapp/3862b54ddda3dc45a581b1733342a7bc

milessabin:
Do either or both of you have time to try and get this over the finish line for 2.3.1?

clhodapp:
I'll make a PR this evening or next.

milessabin:
Oh, blimey! Missed this ... sorry :-(

PR #539, opened 2016-02-02
Fix #519: Fix documentation for CoproductOps#unify

### LINKED ISSUE #519: The doc on Coproduct Unifier is wrong
[Here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/syntax/coproduct.scala#L182) the doc says that `Unifier` returns `Coproduct` made of `LUB`. In fact it returns the `LUB` itself.

vpavkin:


milessabin:
Many thanks! Merging :-)

PR #540, opened 2016-02-08
Fixed OSGi bundle name/symbolic name and removed quasiquotes import package on 2.10

mpilquist:


milessabin:
Many thanks! Merging :-)

tpolecat:
@mpilquist thanks! I'm probably your only customer so I do appreciate it. :chicken:

PR #541, opened 2016-02-09
More robust Default derivation

alexarchambault:
Fixes it failing in some cases, when an extra `apply` method (with defaults) is defined in the case class companion.


milessabin:
Could you rebase this against master? To support a Scala.js `CrossProject(CrossType.Full)` everything under `core/src` has been moved to `core/shared/src`.

milessabin:
Awesome! Thanks ... merging :-)

PR #542, opened 2016-02-09
FIx #457 : Add ToPreciseTraversable type class and tests

### LINKED ISSUE #457: Add a way to transform any HList into a Monoid of some Coproduct
Given a `Monoid` instance `F`, it is possible to convert any `HList` into a `F[C]` where `C` would be the smallest coproduct that encompasses all the input `HList` elements' types.

For example the desired function would transform 
`1 :: "Foo" :: true :: 42 :: HNil` 
into a 
`F[Int :+: String :+: Boolean :+: CNil]` 
for any given `F`.

See [this gist](https://gist.github.com/vil1/2c45599bc01c05353476) for a stub implementation (specialized for the `List` monoid)

vil1:
In the absence of an accessible `Monoid` definition, this should be ok for now.

The name is debatable (I tried to find a name close to `ToTraversable` since both type classes perform similar work).

Tests could be enhanced, especially if I were able (or aware of a way) to express "equality" of Coproducts (e.g. expressing that `A :+: B :+: CNil == B :+: A :+: CNil`).

### DIFF
--- a/core/shared/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/shared/src/main/scala/shapeless/ops/hlists.scala
@@ -630,13 +630,13 @@ object hlist {
     *
     * @author Valentin Kasas
     */
-  trait ToPreciseTraversable[L <: HList, M[_]] extends DepFn1[L] with Serializable {
+  trait ToCoproductTraversable[L <: HList, M[_]] extends DepFn1[L] with Serializable {
     type Cop <: Coproduct
     type Out = M[Cop]
   }
 
-  trait LowPriorityToPreciseTraversable {
-    type Aux[L <: HList, M[_], C <: Coproduct] = ToPreciseTraversable[L, M] {type Cop = C}
+  trait LowPriorityToCoproductTraversable {
+    type Aux[L <: HList, M[_], C <: Coproduct] = ToCoproductTraversable[L, M] {type Cop = C}
 
     /**
       * Auxiliary type class witnessing that type `T` is not a member of `Coproduct` `C`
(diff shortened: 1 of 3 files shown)


milessabin:
You should be able to assert equality of `Coproduct` types using `shapeless.ops.coproduct.Basis`.

milessabin:
Could you rebase this against master? To support a Scala.js `CrossProject(CrossType.Full)` everything under `core/src` has been moved to `core/shared/src`.

vil1:
Done (great news by the way !)

milessabin:
This looks pretty good :-)

I think we need a better name though. Given that the distinguishing feature from `to[M]` and `ToTraversable` is that the element type of the resulting collection is a `Coproduct` rather than the LUB then I think maybe `toCoproduct[M]` and `ToCoproductTraversable` would be better.

vil1:
Ok. 
I had considered that in the beginning, but feared that it would be slightly misleading (mostly due to the words order). I felt that it would be unsettling that `toCoproduct[M]` does not return a `Coproduct`, at least for newcomers.

milessabin:
I think those are all very sensible concerns, but I think this is the better option.

milessabin:
LGTM and Travis is green! Many thanks ... merging :-)

vil1:
All the pleasure we for me. 

(look Mom, I contributed to open source software ! :-) ).

milessabin:
Would you mind adding your details to CONTRIBUTORS and README.md?

dwijnand:
From the peanut gallery here, I think `toCoproduct[List]`/`toCoproduct[Vector]` looks great.

PR #548, opened 2016-02-18
NotContains and IsDistinct constraints for HList

vpavkin:
There was a [conversation](https://gitter.im/milessabin/shapeless?at=56bf14a14dfe1fa71ffc852f) on gitter regarding these, but some urgent stuff weighted in short after.
So I leave it here for better times to be either merged or rejected :)

Note:
An implementation of `IsDistinctConstraint` without use of `NotContainsConstraint` can be following (with a use of ambiguous implicits trick):

```scala
trait IsDistinctConstraint[L <: HList]

object IsDistinctConstraint {
  def apply[T <: HList](implicit ev: IsDistinctConstraint[T]): IsDistinctConstraint[T] = ev

  // we have to build chain here so that ambiguous breaker works for smth like A :: B :: B :: HNil
  implicit def unique0 = new IsDistinctConstraint[HNil] {}
  implicit def unique1[H, T <: HList](implicit ev: IsDistinctConstraint[T]) = new IsDistinctConstraint[H :: T] {}

  implicit def uniqueAmb1[H <: HList, T](implicit ev: Selector[H, T]): IsDistinctConstraint[T :: H] = 
    shapeless.unexpected
  implicit def uniqueAmb2[H <: HList, T](implicit ev: Selector[H, T]): IsDistinctConstraint[T :: H] = 
    shapeless.unexpected
}
```

Maybe it's a better choice...

### DIFF
--- a/core/src/main/scala/shapeless/hlistconstraints.scala
+++ b/core/src/main/scala/shapeless/hlistconstraints.scala
@@ -18,6 +18,8 @@ package shapeless
 
 import ops.hlist.Selector
 
+import scala.annotation.implicitNotFound
+
 /**
  * Type class witnessing that every element of `L` has `TC` as its outer type constructor. 
  */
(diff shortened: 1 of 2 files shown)


milessabin:
This looks very promising, but the RC cycle for 2.3.0 has already started so I'm afraid I'm going to have to punt it until 2.3.1.

vpavkin:
Sure, would be great! :)

milessabin:

PR #549, opened 2016-02-21
Add impossible method to CNil

m50d:
As discussed on gitter, add an `impossible` method on `CNil` to allow safer `match`/`case` expressions for `Coproduct`s.

### DIFF
--- a/core/src/main/scala/shapeless/coproduct.scala
+++ b/core/src/main/scala/shapeless/coproduct.scala
@@ -105,7 +105,9 @@ final case class Inr[+H, +T <: Coproduct](tail : T) extends :+:[H, T]
   * This makes the type `Int :+: CNil` equivalent to `Int`, because the right (`Inr`) alternative
   * of `:+:` can not be constructed properly.
   */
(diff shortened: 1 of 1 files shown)


milessabin:
Even if the compiler is happy with there being no implementation I really think there ought to be one.

m50d:
I disagree: that there isn't an implementation emphasises that there *can't* be an implementation, that signature is unimplementable (in a safe subset of scala), and the compiler is happy precisely because the trait will never be instantiated (something that would be less clear if it were implemented as `???` or similar, IMO). Also the whole point of this change is to allow programs that use coproducts to be written in safe subsets of scala; implementing it using an unsafe construct would undermine that somewhat.

Incomplete thought: `CNil` is isomorphic to `Nothing`, and could possibly be replaced by `Nothing` throughout (since `Nothing <: Coproduct`).

milessabin:
Sorry, you're right, I'm wrong. Many thanks ... merging :-)

larsrh:
You've already merged, but :+1: anyway.

PR #550, opened 2016-02-24
Added typeclass SwapRecord collecting the keys tagged by value type

kailuowang:


milessabin:

PR #552, opened 2016-02-26
adding SelectAll typeclass to shapeless.ops.hlist

eugengarkusha:
related to todays conversation in gitter:
milessabin 12:05
I think it would work as a SelectAll operation for HLists though ... I don't think there's anything there which is intrinsically tied to records.

Also deleted the "Aux" type from Selector object as it is just an alias for Selector trait.


milessabin:
This is great! Many thanks ... merging :-)

PR #559, opened 2016-03-11
fix typo in union example

joprice:


milessabin:
Good spot ... many thanks ... merging :-)

joprice:
By the by, have you considered using https://github.com/tkawachi/sbt-doctest to keep things like this in check?

milessabin:
@joprice Yes, that would be useful ... PR?

joprice:
I took a stab at it. I'm getting this in the coproduct examples:

```
could not find implicit value for parameter gen: shapeless.Generic[this.Animal]
```

The code is extracted into a test case in src_managed and run inside the constructor of an anonymous org.scalacheck.Properties instance. Perhaps defining them as inner classes trips up `knownDirectSubclasses`?

PR #561, opened 2016-03-13
Add generalized sub-typing evidence type class

alexarchambault:
Adds a more general sub-type evidence `<::<`, compatible with coproducts. Like with `<:<` from the standard library, usual sub-typing is fine, e.g.

``` scala
<::<[String, AnyRef]
```

but the first type can be a coproduct too, e.g.

``` scala
<::<[List[Double] :+: Vector[Double] :+: IndexedSeq[Double] :+: CNil, Seq[Double]]
```

This can also recurse on the coproducts elements, like in

``` scala
<::<[(String :+: CNil) :+: (String :+: CNil) :+: CNil, String]
```

### DIFF
--- a/core/src/main/scala/shapeless/coproduct.scala
+++ b/core/src/main/scala/shapeless/coproduct.scala
@@ -147,3 +147,52 @@ object Coproduct extends Dynamic {
    */
   def selectDynamic(tpeSelector: String): Any = macro LabelledMacros.coproductTypeImpl
 }
+
+/**
+  * More general sub-type evidence, compatible with coproducts.
+  *
+  * Evidence that `From` can be converted to `To`. Works if `From` is a sub-type of `To`, like
+  * with `<:<` from the standard library. `From` can also possibly be a coproduct, whose elements
+  * can all be converted to `To`.
+  *
+  * This can recurse on the elements of coproducts, so that if `From` is a coproduct, this will also
+  * work if some or all its elements are coproducts that can be converted to `To`.
+  *
+  * E.g.
+  *
+  * {{{
+  *   <::<[String, AnyRef]
+  *   <::<[String :+: Seq[Int] :+: Option[Double] :+: CNil, AnyRef]
+  *   <::<[List[Double] :+: Vector[Double] :+: IndexedSeq[Double] :+: CNil, Seq[Double]]
+  * }}}
+  *
+  * @author Alexandre Archambault
+  */
+sealed abstract class <::<[From, To] extends (From => To) with Serializable
+
+object <::< {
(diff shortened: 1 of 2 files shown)


milessabin:
I missed this earlier somehow ... sorry :-(

I'm fine with the general idea, but I'm a bit worried that making this look so much like a subtype-based operation will cause confusion given that coproducts are already covariant. I think I'd prefer something that made it seem like a generalized inject. Do you have any scenarios where you would need this operation where there could be both coproducts and non-coproducts on the left?

alexarchambault:
I have one use case. Which might sound a bit specific - I was thinking the whole thing was more general than that and could have other uses.

The use case is in [case-app](https://github.com/alexarchambault/case-app). It has a [CommandAppOf](https://github.com/alexarchambault/case-app/blob/912590824ec11a88953ba3d5967fe8f664886b70/core/shared/src/main/scala/caseapp/App.scala#L187) class with a type parameter, that it enforces to be `<: ArgsApp`. Up to now, this type parameter was the root of an ADT, itself extending `ArgsApp`. [Recently](https://github.com/alexarchambault/coursier/blob/fd4d1bd1dc011ed232d203f76c72f7ba84118234/cli/src/main/scala-2.11/coursier/cli/Coursier.scala#L31-L33), I wanted to split the constructors of this ADT among several files, for readability. Which one can't do without removing the `sealed` of the root type, which makes the whole thing not be an ADT anymore. A union type with the former constructor types as cases is a natural substitute to the former root type then - like it, an instance of it can only be one of the previous constructor types. Then, I needed to express the `<: ArgsApp` constraint, for either a standard type, or a coproduct or union type. Which `<::<` in this PR does.

I'm fine with closing this or putting it on hold if it doesn't seem general enough you though...

milessabin:
Gotcha.

I'm definitely interested in having the generalized inject. I think you should be able to use that as a building block for the operation you need.

milessabin:
Do you think you'll have time to work on this for 2.3.1?

alexarchambault:
@milessabin What did you mean by generalized inject?

The initial idea of this PR was to have something somehow approaching the [union types](https://d-d.me/talks/scalaworld2015/#/11) of dotty. That is, in dotty, it seems that we have `A <:< (A | B)` and `B <:< (A | B)`.

With `<::<` of this PR, we get `A <::< (A :+: B :+: CNil)` and `B <::< (A :+: B :+: CNil)`, that approaches the former relations.

milessabin:

milessabin:
I'm closing this for now. Happy to look at this again if you want to work on the "generalized inject" approach.

Issue #562, opened 2016-03-19
Intersection.Aux[Int :: HNil, Int :: HNil, HNil]


refried:
There seems to be a problem with Intersection:

```
scala> implicitly[Intersection.Aux[Int :: HNil, Int :: HNil, Int :: HNil]]
res27: Intersection.Aux[Int :: HNil, Int :: HNil, Int :: HNil] = shapeless.ops.hlist$Intersection$$anon$58@342165b9
```
This correctly witnesses that  { Int } ∩ { Int } = { Int }

```scala
scala> implicitly[Intersection.Aux[Int :: HNil, Int :: HNil, HNil]]
res25: Intersection.Aux[Int :: HNil, Int :: HNil, HNil] = shapeless.ops.hlist$LowPriorityIntersection$$anon$56@50fc9b74
```
This incorrectly witnesses that { Int } ∩ { Int } = ∅ 
:disappointed:

refried:
```
scala> reify { implicitly[Intersection.Aux[Int::HNil,Int::HNil,HNil]] }
res29: $r.intp.global.Expr[shapeless.ops.hlist.Intersection.Aux[shapeless.::[Int,shapeless.HNil],shapeless.::[Int,shapeless.HNil],shapeless.HNil]] = Expr[shapeless.ops.hlist.Intersection.Aux[shapeless.::[Int,shapeless.HNil],shapeless.::[Int,shapeless.HNil],shapeless.HNil]](Predef.implicitly[hlist.Intersection.Aux[::[Int, HNil], ::[Int, HNil], HNil]](hlist.this.Intersection.hlistIntersection1(hlist.this.Intersection.hnilIntersection)))
```
simplifies to 
```
Intersection.hlistIntersection1(Intersection.hnilIntersection))
```
```scala
implicit def hnilIntersection[M <: HList]: Aux[HNil, M, HNil] =
  new Intersection[HNil, M] {
    type Out = HNil
    def apply(l: HNil): Out = HNil
  }

implicit def hlistIntersection1[H, T <: HList, M <: HList]
  (implicit i: Intersection[T, M]): Aux[H :: T, M, i.Out] =
    new Intersection[H :: T, M] {
      type Out = i.Out
      def apply(l: H :: T): Out = i(l.tail)
    }
```
Yeah.  `hnilIntersection` says `∅ ∩ {Int} = ∅`, which is good.
[...]

PR #563, opened 2016-03-19
fix bug in hlist.{Intersection, Union}

refried:
Fix bugs described in #562.

Currently `Intersection` is witnessed by three cases:
`hnilIntersection` says ∅ ∩ M = ∅, which is good.
`hlistIntersection1` says (H :: T ∩ M) = T ∩ M
`hlistIntersection2` says: if H ∈ M, (H :: T ∩ M) = H :: (T ∩ (M - R))

Unfortunately, (H :: T ∩ M) = T ∩ M is false if H ∉ M.  This patch adds the H ∉ M condition to `hlistIntersection1` using a `FilterNot.Aux[M,H,M]` witness.

The patch also fixes an equivalent bug in `Union`.

---
I had a couple additional questions:

1. Should `hlistIntersection1` and `hlistIntersection2` reside in the same trait now, eliminating the need for `LowPriorityIntersection`?
2. Why is it that `Intersection extends DepFn1` but `Union extends DepFn2`?  It seems like they should be the same, but I don't know what the right answer is.

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -1251,12 +1251,10 @@ object hlist {
   trait LowPriorityUnion {
     type Aux[L <: HList, M <: HList, Out0 <: HList] = Union[L, M] { type Out = Out0 }
 
-    // let (H :: T) ∪ M  =  H :: (T ∪ M) when H ∉ M
-    implicit def hlistUnion1[H, T <: HList, M <: HList]
-      (implicit
-       u: Union[T, M],
-       f: FilterNot.Aux[M, H, M]
-      ): Aux[H :: T, M, H :: u.Out] =
+    // buggy version; let (H :: T) ∪ M  =  H :: (T ∪ M)
+    @deprecated("Incorrectly witnesses that {x} ∪ {x} = {x, x}", "2.3.1")
+    def hlistUnion1[H, T <: HList, M <: HList]
+      (implicit u: Union[T, M]): Aux[H :: T, M, H :: u.Out] =
         new Union[H :: T, M] {
           type Out = H :: u.Out
           def apply(l: H :: T, m: M): Out = l.head :: u(l.tail, m)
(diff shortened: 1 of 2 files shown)


refried:
Fails mima check — this is tricky, the patch not bincompat with existing, but existing is buggy. :-\

OlivierBlanvillain:
Hi @refried, thanks for fixing this, your changes are LGTM!

Regarding your 2nd question, I think `A ∩ B` only contains elements from `A`, while `A U B` could have elements from both `A` and `B`, which would explain why you don't need the same number of arguments to produce the output. Note that for `A - B`, the type of `B` is also sufficient, as the output will only contain values from `A`.

alexarchambault:
About MiMa, it is possible keep binary compatibility. You just have to keep the former methods as is, possibly modifying a bit their implementation, and *removing their `implicit` modifier*.

This way, code compiled against older versions of shapeless will still find the methods they're looking for at runtime, and code compiled against the newer version will only find the newest method, which are the only ones flagged as implicit.

milessabin:
+1 to @alexarchambault's advice ... I'll merge when MiMa passes ... thanks! :-)

refried:
@alexarchambault @milessabin Ok, I can make that MiMa change, but please advise/confirm:  the binary-compatible version will necessarily continue to be buggy, which is pretty scary to me.  (The fix was to add the additional witness.)  My instinct to add a @deprecation warning won't help if there's no recompilation.

Do I add `@deprecation` for now we delete the buggy method in the next minor version?  Is there a process for remembering to have that happen?  Thanks!

alexarchambault:
@refried As the methods kept for binary compatibility don't have the `implicit` modifier anymore, they won't be found when using `Intersection` or `Union`. Only code compiled against older versions of shapeless should use them.

I don't mind about adding `@deprecated`. These methods are seldomly directly called anyway...

refried:
@alexarchambault Sure, I understand that — I just wished there was a way to warn the person linking against the old version; but I guess there isn't, and they wouldn't discover a bincompat issue until runtime.  I pushed the update, so fingers crossed! :smile:

milessabin:

PR #564, opened 2016-03-22
Scope credentials to ThisBuild

dwijnand:
Something about using .sbtrc & switching projects in validate etc is confusingly
not making credentials evaluate.

As a stop-gap, set this to ThisBuild which arguably is what it should be: the
entire project, every module, should publish with these credentials.


milessabin:

PR #565, opened 2016-03-22
Bump to Scala 2.11.8

dwijnand:
There's a trailing change about shapeless-2.2.5 using Scala 2.11.7, I've
purposely left that as is.


milessabin:
LGTM. Many thanks ... merging.

PR #566, opened 2016-03-22
Document Witness

dwijnand:
(Nerd-sniping Miles into documenting these awesome features!)


notxcain:
:rocket:

milessabin:
It isn't strictly speaking an implementation of SIP-23, although it does cover the same ground (sorry if I was misleading earlier). I think it would be better to say something like "See SIP-23 for a related proposed language change".

dwijnand:
Amended.

milessabin:
Many thanks! Merged :-)

Issue #486, opened 2015-10-29
compilation time for record access depends on value types


tpolecat:
Following up on a Gitter conversation from a few days ago.

When accessing an element in a record the compilation time not only depends on the position of the element (as expected since you have to walk the list) but also on the type of the associated value.

```scala
// shapeless 2.2.5
// scala 2.11.7

import shapeless._
import shapeless.record._
import shapeless.ops.record._
import shapeless.syntax._
import shapeless.syntax.singleton._
import shapeless.labelled._

// construct a record with 15 elements
def test[A](a:A) = {
  ('k1  ->> a) :: ('k2  ->> a) :: ('k3  ->> a) :: ('k4  ->> a) :: ('k5  ->> a) :: ('k6  ->> a) ::
  ('k7  ->> a) :: ('k8  ->> a) :: ('k9  ->> a) :: ('k10 ->> a) :: ('k11 ->> a) :: ('k12 ->> a) ::
  ('k13 ->> a) :: ('k14 ->> a) :: ('k15 ->> a) :: HNil
}

val r1 = test(42)
val r2 = test(1 :: 2 :: 3 :: 4 :: 5 :: 6 :: 7 :: 8 :: 9 :: 10 :: 11 :: 12 :: 13 :: 14 :: 15 :: HNil)
val r3 = test(test(42))
 
// accessing 'k1 is acceptably fast in all cases, but when looking at 'k15 the 
// difference in cost becomes significant

// r1('k15) //  1 sec
// r2('k15) //  8 sec
// r3('k15) // 54 sec
```

milessabin:
Interesting! This ought to be the same issue as #420, but clearly isn't.

milessabin:
I'm not seeing that with the latest 2.3.0-SNAPSHOT ... SBT incremental compile time for uncommenting each of those calls individually is a constant 2s on my (modest) machine. Could you give it another try?

tpolecat:
Yep, works great in 2.3.0-SNAPSHOT. :+1:

milessabin:
Awesome ... thanks :-)

tpolecat:
This behavior persists when mapping, and becomes worse as the element types become fancier. Not sure if this is related/expected.

```scala

// continued from above
object Two extends Poly1 {
  implicit def all[A] = at[A](_ => 2)
}

r1.map(Two) // 0 sec
r2.map(Two) // 1 sec
r3.map(Two) // 5 sec
```

milessabin:
[...]

Issue #420, opened 2015-06-23
Poor compile time performance of shapeless.ops.record.{ Selector, Updater }


milessabin:
Compile time scales super-linearly with number of case class members. This seems to account for most of the poor compile time performance of shapeless lens builders.

```scala
import shapeless._, record._

case class Foo(
  /*
  z0: Int,
  y0: Int,
  x0: Int,
  w0: Int,
  v0: Int,
  u0: Int,
  t0: Int,
  s0: Int,
  r0: Int,
  q0: Int,
  p0: Int,
  o0: Int,
  n0: Int,
  m0: Int,
  l0: Int,
  k0: Int,
  j0: Int,
  i0: Int,
  h0: Int,
  g0: Int,
  f0: Int,
  e0: Int,
  d0: Int,
  c0: Int,
  b0: Int,
  a0: Int,
  z: Int,
  */
  y: Int,
  x: Int,                                                                                                
  w: Int,                                                                                                
  v: Int,                                                                                                
  u: Int,                                                                                                
  t: Int,                                                                                                
  s: Int,                                                                                                
  r: Int,                                                                                                
  q: Int,                                                                                                
  p: Int,                                                                                                
  o: Int,                                                                                                
  n: Int,                                                                                                
  m: Int,                                                                                                
  l: Int,                                                                                                
  k: Int,                                                                                                
  j: Int,                                                                                                
  i: Int,                                                                                                
  h: Int,                                                                                                
  g: Int,                                                                                                
  f: Int,                                                                                                
  e: Int,                                                                                                
  d: Int,                                                                                                
  c: Int,                                                                                                
  b: Int,                                                                                                
  a: Int                                                                                                 
)                                                                                                        
                                                                                                         
object Test {                                                                                            
  val lgen = LabelledGeneric[Foo]                                                                        
  val foo: Foo = null                                                                                    
  val r = lgen.to(foo)                                                                                   
                                                                                                         
  r('a)                                                                                                  
  r.updated('a, 23)                                                                                      
}
```

milessabin:
Fixed in d4c3c71933e8c4ab6bc1fcde17e92961f9c0f897.

PR #567, opened 2016-03-22
macro based Modifier for Record

eugengarkusha:
1)implemented macro based Modifier
2)implemented Updater and Remover in terms of Modifier
3)fixed the following inconsistency in record syntax: 
**before**: "updated" method was used to _update_ and _add_ new fields and  it was possible to occasionally add another field with same key(in case of providing wrong  value type in updated method).
**now** : "updated" method only updates the field if the record contains appropriate key and value types, "add" method adds new field if record does not contain the field with same key.

**IMPORTANT NOTE:**
This change breaks binary compatibility with 2.3.0.

### DIFF
--- a/core/src/main/scala/shapeless/hlists.scala
+++ b/core/src/main/scala/shapeless/hlists.scala
@@ -111,18 +111,7 @@ object HList extends Dynamic {
     else unsafeGet(c.tail, i-1)
   }
 
-  def unsafeUpdate(l: HList, i: Int, e: Any): HList = {
-    @tailrec
-    def loop(l: HList, i: Int, prefix: List[Any]): (List[Any], HList) =
-      l match {
-        case HNil => (prefix, e :: HNil)
-        case hd :: (tl : HList) if i == 0 => (prefix, e :: tl)
-        case hd :: (tl : HList) => loop(tl, i-1, hd :: prefix)
-      }
 
-    val (prefix, suffix) = loop(l, i, Nil)
-    prefix.foldLeft(suffix) { (tl, hd) => hd :: tl }
-  }
 }
 
 
--- a/core/src/main/scala/shapeless/hlists.scala
+++ b/core/src/main/scala/shapeless/hlists.scala
@@ -111,7 +111,18 @@ object HList extends Dynamic {
     else unsafeGet(c.tail, i-1)
   }
 
+  def unsafeUpdate(l: HList, i: Int, e: Any): HList = {
+    @tailrec
+    def loop(l: HList, i: Int, prefix: List[Any]): (List[Any], HList) =
+      l match {
+        case HNil => (prefix, e :: HNil)
+        case hd :: (tl : HList) if i == 0 => (prefix, e :: tl)
+        case hd :: (tl : HList) => loop(tl, i-1, hd :: prefix)
+      }
 
+    val (prefix, suffix) = loop(l, i, Nil)
+    prefix.foldLeft(suffix) { (tl, hd) => hd :: tl }
+  }
 }
 
 
--- a/core/src/main/scala/shapeless/hlists.scala
+++ b/core/src/main/scala/shapeless/hlists.scala
@@ -18,7 +18,6 @@ package shapeless
 
 import scala.language.dynamics
 import scala.language.experimental.macros
-
 import scala.annotation.tailrec
 import scala.reflect.macros.whitebox
 
@@ -105,12 +104,14 @@ object HList extends Dynamic {
   def selectDynamic(tpeSelector: String): Any = macro LabelledMacros.hlistTypeImpl
 
   @tailrec
+  @deprecated("used for binary compatibility with 2.3.0", "2.3.1")
   def unsafeGet(l: HList, i: Int): Any = {
     val c = l.asInstanceOf[::[Any, HList]]
     if(i == 0) c.head
     else unsafeGet(c.tail, i-1)
   }
 
+  @deprecated("used for binary compatibility with 2.3.0", "2.3.1")
   def unsafeUpdate(l: HList, i: Int, e: Any): HList = {
     @tailrec
     def loop(l: HList, i: Int, prefix: List[Any]): (List[Any], HList) =
@@ -122,6 +123,24 @@ object HList extends Dynamic {
 
     val (prefix, suffix) = loop(l, i, Nil)
     prefix.foldLeft(suffix) { (tl, hd) => hd :: tl }
+  }
+
+  //TODO: remove @noinline after switching to ScalaJs 0.69+
+  @noinline
+  def unsafeCrud(l: HList, i: Int, f: Any => Any, remove:Boolean): (Any, HList) = {
+    @tailrec
+    def loop(l: HList, i: Int, prefix: List[Any]): (List[Any], HList, Any) =
+      l match {
+        //add case
+        case HNil => if(remove)throw new Exception("Index out of bounds.Cannot remove.") else (prefix, f(null) :: HNil, null)
+        //remove / modify case
+        case hd :: (tl: HList) if i == 0 => if(remove)(prefix, tl, hd) else (prefix, f(hd) :: tl, hd)
+        //recursion step
+        case hd :: (tl: HList) => loop(tl, i - 1, hd :: prefix)
+      }
+
+    val (prefix, suffix, v) = loop(l, i, Nil)
+    v -> prefix.foldLeft(suffix) { (tl, hd) => hd :: tl }
   }
 }
 
--- a/core/src/main/scala/shapeless/ops/records.scala
+++ b/core/src/main/scala/shapeless/ops/records.scala
@@ -29,7 +29,7 @@ import poly._
 package record {
   import shapeless.labelled._
 
-  import scala.annotation.tailrec
+  import scala.annotation.{implicitNotFound, tailrec}
 
   /**
    * Type class supporting record field selection.
@@ -163,7 +163,6 @@ package record {
    *
    * @author Ievgen Garkusha
    */
-
   trait Modifier[L <: HList,K, V, R] extends DepFn2[L, V => R] with Serializable { type Out <: HList }
 
   object Modifier {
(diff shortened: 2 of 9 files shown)


eugengarkusha:
Working on implementing Select in terms of Modify. It will turn it into single implementation of all CRUD operations

eugengarkusha:
Introduced _Crud_ typeclass responsible for crud operations on _Record_. It substitutes _Updater_, _Modifier_, _Remover_ and _Selector_. _Selector_ is not yet removed because _Crud_ marco is not being expanded when spawned inside _WitnessWith_.

milessabin:
Thanks!

I don't think it will be possible to merge this as is ... there are many people depending on the current type classes so removing them in this way isn't possible. Alternative implementations of the current type classes are possible (with the proviso that we need to maintain binary compatibility within the 2.3.x series), so maybe you move back to (some iteration of) what you had earlier for me to review?

eugengarkusha:
Would it be acceptable to
1) restore Remover, Updater and Modifier(Selector is still there) for backward compatibility. (Mark them as deprecated?).
2)leave all related operations in syntax.record and lenses  Crud-based. 

Thanks.

milessabin:
I suggest making the smallest change that restores the original operations and binary compatibility whilst adding the new functionality you want so that I have something I can start to review.

eugengarkusha:
Oh, i see that  point 2 is not an option.  Will be back with updates upon restoring of binary compatibility.

alexarchambault:
Like in https://github.com/milessabin/shapeless/pull/563, you can keep binary compatibility with non-implicit stubs along the new code.

eugengarkusha:
Yes looks neat , will  go this way!
 Thanks

eugengarkusha:
@alexarchambault  , your great advice helped me with JVM binary compatibility, but have you any ideas of how to fix ScalaJs compatibility?

eugengarkusha:
blocked by the following ScalaJs issue  https://github.com/scala-js/scala-js/issues/2307

eugengarkusha:
Issue scala-js/scala-js#2307 is fixed by sjrd  and should be available in next ScalaJs release(0.69?)

milessabin:
Good job. I'll bump the Scala.js version as soon as it's available and we can proceed from there.

eugengarkusha:
sjrd advised the following to fix the problem rigtht away:
"As a work around to immediately fix your issue, you can add @noinline on def unsafeUpdate"
It fixes the problem completely!

milessabin:
Can you give us an indication of how this affects compile times? Something along the lines of the timings in #486 and the example in #420.

eugengarkusha:
All tests were made involving get, updated, updateWIth and  remove operations simultaneously:

#420
1)Compile time grows linearly :
initial case class (50 flds)  ~10 seconds
case class with 103 fields  ~20 seconds
2)increasing the amount of operations by 10 times(resulting in 40 operations) has no significant impact on compile times (+ 2-3 seconds)

#486
1)no modifications ˜ 6 seconds
2)'k15 is involved in get, updated, updateWIth and  remove operations  10 times in a row(get: 4, update: 2, updateWith: 2, remove: 2):   ~9 seconds

milessabin:
I don't follow. Can you say what 1) and 2) are?

milessabin:
I would be really helpful to have a couple of examples which are infeasibly slow to compile before this change, but which compile in an acceptable amount of time after it.

eugengarkusha:
I have edited my previous response, hope it is more clear now.
Example of  infeasibly slow compilation before these changes:
1)updateWith operation on Record with 50-70 fields(or many updateWith operations of shorter records)
2)Remove operations were also slower(have not tested the limits)

eugengarkusha:
Should I provide a repo with examples?

milessabin:
That would be really helpful :-)

eugengarkusha:
Here it is :
https://github.com/eugengarkusha/ShapelessPR-567

Currently depends on shapeless 2.3.1-snapshot(locally published branch that contains these changes)
Switch to 2.3.0 to see slow compilation
Thanks!

milessabin:
Thanks for that! :-)

eugengarkusha:
YW, pls notify me if you need something else

milessabin:
I'm good for now. Give me a day or two ...

eugengarkusha:
Thanks !

milessabin:
I've tried out the test project and the speedups are great! Thank you!

I would like you change the structure of this change before I merge it though. Rather than having a single type class which does the work for all the operations could you stick with the current structure (ie. one type class, the existing one for all but the add operation which is new) and have your macro materialize instances of those using the strategy you've adopted here.

The benefit of this is particularly visible wrt update: if the existing `Updater` type class simply has a new materializer implementation, then the `replace` syntax is unnecessary and code using `update` will benefit from your changes as soon as it is recompiled against a new shapeless release.

eugengarkusha:
Great, thanks for your feedback!

Please correct me if I am wrong:  

1)The task is to materialize different type classes(having different method signatures) with a single macro(if it is not the case  then would you pls be so kind to provide a bit more formal task definition ).

2)"replace" semantics differs from "updated" as it supports only replace (not add like "updated" does) so the code  using update wonís t compile in number of places in case of drop-in replacement.

Thanks!

milessabin:
Apologies about the misunderstanding wrt replace vs. update ... yes, a separate replace method makes sense.

Am I right that you're adding two fundamentally new operations on records: replace (which is like the existing `Update` except that it will not add extra fields) and add (which adds a field only if it is not already present)?

If that's the case I would prefer to see those two operations added as independent type classes (eg. `Replacer` and `Adder`). I understand the motivation behind merging them and the existing `Selector`, `Updater`, `Remover` etc. as `Crud`, and it might have made sense to do things that way at the outset but, as things are, I think it's better to continue with these as independent type classes.

Each of these type classes requires a separate implicit macro materializer method, but those can delegate to a single shared method which contains the bulk of the implementation.

As well as that, I'd like to see the method `unsafeUpdate` moved to `object HList` ... maybe renamed to `unsafeUpdateWith` to avoid overloading with the existing `unsafeUpdate`.

You've done all the hard work already, I'm really just proposing shuffling things around a bit ... I hope you don't mind.

eugengarkusha:
1)Regarding  Update and Add: yes you are right.
2)Sure I don't mind, moreover it is an honour for me to  be useful for Shapeless.
Thanks!

eugengarkusha:
The requested implementations is provided.
Only thing that bothers me is that instead of "scrapping my boilerplate" I have written more boilerplate.
Is it really important to stick to the existing conventions even if this requires to act against one of the the main goals this library manifests to achieve?
In future releases It could be possible to replace 4 existing typeclasses with one and avoid adding 2 new typeclasses and materializers, still keeping complexity around the same level .

I apologize in advance in case if these questions are inappropriate, I might have overlooked some details which in this case I am asking you to point out. 

Thanks!

milessabin:
Hmm ... that didn't come out quite the way I was expecting.

Why have you added the `UnsafeXxxx` type classes?

eugengarkusha:
I might have misunderstood you.  Will try to do better today

Hmm ... that didn't come out quite the way I was expecting.

Why have you added the UnsafeXxxx type classes?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/milessabin/shapeless/pull/567#issuecomment-204898093>

eugengarkusha:
Could you pls have a quick glance on this branch again.
Does it look better now?
Pls note that a cast to ```Aux``` in each type class materializer is caued by the fact that  I dont know how to encode ```Out``` type member in quasiquotes(asked in shapeleess and scalamacros gitter channels with no result).

core/src/main/scala/shapeless/hlists.scala
```diff
@@ -123,6 +124,24 @@ object HList extends Dynamic {
     val (prefix, suffix) = loop(l, i, Nil)
     prefix.foldLeft(suffix) { (tl, hd) => hd :: tl }
   }
+
+  //TODO: remove @noinline after switching to ScalaJs 0.69+
+  @noinline
+  def unsafeCrud(l: HList, i: Int, f: Any => Any, remove:Boolean): (Any, HList) = {
```
milessabin: I would really prefer this to be called `unsafeUpdateWith`.

core/src/main/scala/shapeless/ops/records.scala
```diff
@@ -49,37 +50,35 @@ package record {
 
   }
 
+  //used for binary compatibility with 2.3.0
   class UnsafeSelector(i: Int) extends Selector[HList, Any] {
     type Out = Any
-    def apply(l: HList): Any = HList.unsafeGet(l, i)
+    def apply(l: HList): Any = HList.unsafeCrud(l, i, identity, false)._1
```
milessabin: Is this change necessary. I think the original `unsafeGet` will have the same compile time performance, so this change increases the complexity of the code slightly with no obvious gain.

eugengarkusha:
1)got it
2)it expresses an intent of removing the redundant unsafeGet(the same is for unsafeUpdate) method in future releases. Or this also would be not correct approach?

milessabin:
First, my sincere apologies for taking so long to get back to you on this.

On 2) I think it would be better to go with `unsafeGet` and express the intent to switch to a different implementation by creating a follow up issue here marked for the next major release.

eugengarkusha:
Hi!
Just want to confirm, that i have seen your response, thanks!
Totally busy right now, will  get back to this asap

milessabin:
No worries! :-)

milessabin:
I've just pushed [this](https://github.com/milessabin/shapeless/commit/7b90b2e4b536cbdbb0221f44c8acfa4b4089d5bf) which is partly inspired by the work you did on this PR ... many thanks for that.

Some of the changes are cosmetic (eg. I don't particularly like the "crud" terminology). There were some technical issues as well. You were creating anonymous subclasses in your macros rather than instantiating a concrete class ... that can result in the creating of many more class files. I also factored your very general crud method into several simpler methods, this makes the use of `null` unnecessary and eliminates some runtime boxing in the modifier case.

On the additional operations, I added `replace` but used the `Updater` type class with `Selector` used as a constraint to guarantee that the replacement element was of the same type as the element being replaced. And I decided against `add` ... I didn't feel that an append operation was the right thing to add if that was all that was added, and on the other hand there are too many permutations to add all the plausible variants. So instead I added a `LacksKey` type class which can be used as evidence in conjunction with any operation where you want to ensure that only new keys are added. The [test](https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/records.scala#L497-L514) gives an example of how this might be used.

The main thing is that your compile time benchmark project now compiles in a sensible amount of time ... so thank you very much for encouraging me down this route, and I do hope you don't mind me hijacking this PR.

eugengarkusha:
Great news!!
I will study the work you've done in this and try to learn a lesson from it.
Thanks!

PR #568, opened 2016-03-23
Add SYB instances for collections

guersam:
This PR makes `Data` and `DataT` traverse through arbitrary collections.
    
**NOTE**: The behaviour of `Map` instances differs between `Data` and `DataT`. The former treats a map like a sequence of key/value tuples, whereas the latter just ignores keys to avoid shrinking the result due to some duplicated keys.

### DIFF
--- a/core/src/test/scala/shapeless/sybclass.scala
+++ b/core/src/test/scala/shapeless/sybclass.scala
@@ -152,6 +152,10 @@ class SybClassTests {
     typed[Int](e5)
     assertEquals(11, e5)
 
+    val e6 = everything(gsize)(plus)(Map("foo" -> List(1, 2, 3), "bar" -> Nil))
+    typed[Int](e6)
+    assertEquals(14, e6)
+
     val is = gsizeAll2(23)
     typed[Int](is)
     assertEquals(1, is)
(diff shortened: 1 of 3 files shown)


core/src/test/scala/shapeless/sybclass.scala
```diff
@@ -211,6 +226,18 @@ class SybClassTests {
...
+    typed[Vector[Int]](e9)
+    assertEquals(Vector(24), e9)
+
+    val e10 = everywhere(inc)(Vector(Option(23)))
+    typed[Vector[Option[Int]]](e10)
+    assertEquals(Vector(Option(24)), e10)
+
+    val e11 = everywhere(inc)(Map("foo" -> 23))
+    typed[Map[String, Int]](e11)
+    assertEquals(Map("foo" -> 24), e11)
```
guersam: Maybe I need to add some more tests for nested `GenTraversable`s here.

milessabin: Can't hurt.

core/src/test/scala/shapeless/sybclass.scala
```diff
@@ -22,23 +22,30 @@ import org.junit.Assert._
 import test._
 
 class SybClassTests {
-  object gsizeAll extends Poly1 {
-    implicit def caseString = at[String](_.length)
+
+  trait gsizeAll0 extends Poly1 {
```
guersam: Is this name ok? I chose this among `gsizeAllLowPriority`, `gsizeDefault` and PascalCased alternatives.

milessabin: Yes, that's fine. I've adopted the convention of numeric suffixes for low priority traits ... `LowPriority` is just too long winded for an implementation artefact.

milessabin:
Sorry, I'd missed that you'd updated this. Unfortunately it means that you next PR will conflict on the contributors file :-(

Anyhow ... LGTM ... many thanks ... merging!

PR #569, opened 2016-03-26
Add `Reify` ops for `HList` and `Coproduct`

guersam:
This does something similar to `Keys`, but does not require the argument to be a record or a union.


milessabin:

PR #570, opened 2016-03-27
Add `UnzipFields` ops for Records and Unions

guersam:
It combines `Keys` and `Values` for convenience and compile speed.

They extend none of `DepFn0` or `DepFn1` due to conflicting `type Out`. Instead, I gave explicit type member and method for each output.


milessabin:

PR #571, opened 2016-03-27
Improve tests for `Reify` and `UnzipFields`

guersam:
It depends on #570, so please review this one after that.

Added missing serialization tests for `Reify` and `UnzipFields` and modified `Reify` tests to use `assertTypedEquals` to ensure the reified values' singleton-ness.


milessabin:

PR #572, opened 2016-03-27
Add `unfixFold` method to `:+:`

m50d:
This is what I was trying to get at in our conversation on gitter, apologies if I was unclear.

Rationale: I find it hard to confirm [the rules about when pattern matching is safe](http://typelevel.org/blog/2014/11/10/why_is_adt_pattern_matching_allowed.html), particularly in a code review scenario, so I prefer to write code that avoids using `match`/`case` at all (and can therefore be safe on its face without needing to check the context and types involved). This method makes it possible to implement typeclass derivation without using `match`/`case` as per the example. I've called it `unfixFold` by analogy with recursion-schemes `unfix`: `unfix` indicates we are going to act on the type as though it were not recursive.


m50d:
Build failure is on the binary compatibility check. Unless I'm misunderstanding I think this is spurious: `:+:` is sealed so there should not be any code that extends it out there to be broken by the addition of the method.

milessabin:
Gotcha. Yes, I think this is fine, but for symmetry it should also be added to `HList`. And for consistency with Cats `Xor` it should be called `fold`.

Agreed on the MiMa warning being spurious. There will be another one when you add this to `HList` ... please add filters for both of them.

m50d:
The converse is `HCons#head` and `HCons#tail` which already exist, no?

Calling this `fold` would collide with `CoproductOps#fold`, no?

I will look up how to do the MiMa filter.

milessabin:
Ahh, yes of course, `fold` is already taken.

I'm not convinced by `unfix` as a prefix though ... I would expect to see "unfix" used in the context of some sort of fixed point operator, and I don't think that's what's going on here, even though recursion is in play.

This is maybe a bit odd, but how about calling the method `eliminate`? We're getting rid of the coproduct both in the pattern matching case and via this method.

And a method with the exact same signature can be added to `::`, so I think it should be: if it makes sense to use this as an alternative to pattern matching for coproducts then it makes sense to use it as an alternative to pattern matching for products.

m50d:
A recursive type can be expressed as a fixed point; the unfixed version of `Coproduct` would be something like `sealed trait CoproductUF[+A]; sealed trait CNilUF extends CoproductUF[Nothing]; sealed trait :+:UF[H, A] extends CoproductUF[A]` (with unfix returning `CoproductUF[Coproduct]`) and the `fold` on such a type would be this `unfixFold`. The analogy isn't perfect because it discards some type information though. I don't have strong feelings on the name, happy to go with `eliminate` if you like.

I don't see how you'd add a method with the same signature to `::`, since `::` is both rather than either. The corresponding method would be something like `def unfixFold[A](f: (H, T) => A): A = f(head, tail)`, no? While I'm happy to add that it doesn't seem like a useful method.

milessabin:
Agreed that if coproducts were implemented as fixed points that `unfixFold` would be an appropriate name. Let's go with `eliminate` for now.

Sorry about the `HList` confusion ... I'm running ahead to ideas that will show up in shapeless 3.0.

milessabin:
Looks good ... would you mind squashing before I merge?

m50d:
Done

milessabin:
That doesn't look _quite_ right ... what's that merge commit?

m50d:
It's merging your shapeless/master into mine. I'm not sure why github didn't make it a fast-forward merge (had to do it through the web UI because I was traveling). There's no actual merging happening, but rebase -i skips merges.

milessabin:
Umm ... well, whatever, but that first commit shouldn't be there. Could you take another swing at it?

m50d:
Will merge --squash into a new branch and make a new PR

xuwei-k:
> why github didn't make it a fast-forward merge (had to do it through the web UI because I was traveling). There's no actual merging happening, but rebase -i skips merges.

I think GitHub changed default behavior. ( maybe `Require branches to be up to date before merging` settings )

see https://github.com/blog/2137-protected-branches-improvements

m50d:
I think that's unrelated? Maybe github always makes a merge commit for a PR? and I couldn't see any other way to have my repo just pull from miles'.

milessabin:
Thanks for doing that ... sorry for the busywork :-(

PR #575, opened 2016-04-02
Try to fix #574: Don't generate unsafe Typeable instances for generic case classes

### LINKED ISSUE #574: Typeable instance derived for generic case classes can be unsafe
The automatically derived/generated `shapeless.Typeable` instance for generic case classes can be unsafe in some cases. For example:
```
final case class C[A](s: String)(x: A) {
  val y: A = x
}
```
The `Typeable[C[Int]]` instance can cast a `C[Double]` to a `C[Int]`, which then blows up with a `ClassCastException`, due to a `Double` not being an `Int` (full code: https://github.com/durban/akka-typed-examples/blob/7f7378979cbdb70615690a25207c70cf57f332fb/src/main/scala/com/example/shless/TypeableExample.scala). More variations: https://github.com/durban/akka-typed-examples/blob/7f7378979cbdb70615690a25207c70cf57f332fb/src/main/scala/com/example/shless/TypeableExample2.scala and https://github.com/durban/akka-typed-examples/blob/7f7378979cbdb70615690a25207c70cf57f332fb/src/main/scala/com/example/shless/TypeableExample3.scala.

I guess the reason could be that the macro generating the instance misses some definitions involving `A` (it only seems to take into account the case class accessors).

durban:
My attempt at fixing https://github.com/milessabin/shapeless/issues/574

### DIFF
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -396,8 +396,14 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
 
         if(!pSym.isCaseClass)
           c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
-        val fields = tpe.decls.toList collect {
-          case sym: TermSymbol if sym.isVal && sym.isCaseAccessor => sym.typeSignatureIn(tpe)
+        val fields = tpe.decls.sorted collect {
+          case sym: TermSymbol if sym.isVal && sym.isCaseAccessor =>
+            sym.typeSignatureIn(tpe)
+          case sym: TermSymbol if sym.isVal || sym.isVar ||
+              (sym.isParamAccessor && !(sym.accessed.isTerm && sym.accessed.asTerm.isCaseAccessor)) =>
+            // not a case accessor but a val, var or param,
+            // so we won't be able to type check it safely:
+            c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
         }
         val fieldTypeables = fields.map { field => c.inferImplicitValue(appliedType(typeableTpe, List(field))) }
         if(fieldTypeables.exists(_ == EmptyTree))
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -396,14 +396,18 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
 
         if(!pSym.isCaseClass)
           c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
+        val nonCaseAccessor = tpe.decls.exists {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/shapeless/typeable.scala
```diff
@@ -396,8 +396,14 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
 
         if(!pSym.isCaseClass)
           c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
-        val fields = tpe.decls.toList collect {
-          case sym: TermSymbol if sym.isVal && sym.isCaseAccessor => sym.typeSignatureIn(tpe)
+        val fields = tpe.decls.sorted collect {
```
milessabin: Why sorted?

core/src/main/scala/shapeless/typeable.scala
```diff
@@ -396,8 +396,14 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
...
+          case sym: TermSymbol if sym.isVal && sym.isCaseAccessor =>
+            sym.typeSignatureIn(tpe)
+          case sym: TermSymbol if sym.isVal || sym.isVar ||
+              (sym.isParamAccessor && !(sym.accessed.isTerm && sym.accessed.asTerm.isCaseAccessor)) =>
+            // not a case accessor but a val, var or param,
+            // so we won't be able to type check it safely:
+            c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
```
milessabin: I think it might be better to do this test before the collect ... I appreciate that quite a lot of macros terminate early by aborting, but doing so from within a map or collect doesn't feel right.

milessabin:
A couple of minor things, but otherwise good stuff!

Don't forget to add yourself to CONTRIBUTORS and the README.

core/src/main/scala/shapeless/typeable.scala
durban: Because if I understand correctly the scaladoc of `MemberScopeApi.sorted`, otherwise it's not guaranteed that the members are in declaration order, and `caseClassTypeable` seem to depend on that.

core/src/main/scala/shapeless/typeable.scala
durban: Fair enough.

core/src/main/scala/shapeless/typeable.scala
milessabin: Good catch! There are probably a couple of other places where `sorted` should be used. Would you mind creating a ticket for that so that it doesn't get forgotten about?

milessabin:

core/src/main/scala/shapeless/typeable.scala
durban: Done: #576

milessabin: Thanks :-)

PR #577, opened 2016-04-04
Automatic Liftable instances for quasi quoting

JanBessai:
Adds automatically derived instances for Liftable and thereby allows quasi quoting case classes.

### DIFF
--- a/core/src/main/scala/shapeless/liftables.scala
+++ b/core/src/main/scala/shapeless/liftables.scala
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2016 Jan Bessai, Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


JanBessai:
Failure is described [here](https://github.com/scalamacros/paradise/issues/39). I'll need to find out how to apply the workaround to get Liftable for 2.10.6.

milessabin:
Any thoughts on this? Is the work around `new Liftable ...`?

milessabin:
Did you have a chance to try `new Liftable ...`?

JanBessai:
This should be fine now.

The one remaining piece I don't understand is why [this specific location of caching](https://github.com/milessabin/shapeless/pull/577/commits/58739026953b045a73eee21562460eec6b389cc5#diff-802e359a59040545c1312ed2da57ffa8R56) is necessary for compilation in scala 2.10 and not in 2.11+. Anyhow, it is reasonable to perform caching there.

milessabin:
Just repeating here for the record ...

I think this is great stuff, but it should really be an independent project, there's no need for it to be in the shapeless repo.

milessabin:
I'm going to close this now, and reiterate that I think this is great stuff which deserves it's own standalone project :-)

PR #580, opened 2016-04-08
Configure previousArtifacts to ignore for 2.12.0-M4

dwijnand:


dwijnand:
Btw the failure (which is 2.12.0-M4 failing to find a Scala.js for 2.12.0-M4) was expected. Just pushing the problem forward.

Also, an alternative fix to this would be publishing a version of shapeless 2.3.0 for Scala 2.12.0-M4.

dwijnand:
From https://travis-ci.org/milessabin/shapeless/jobs/122327604 the build for 2.12 is still broken, but it looks live you've published a version of shapeless 2.3.0 for Scala 2.12.0-M4.

milessabin:
Yes, I did. Sorry, I should have caught up with this PR sooner :-(

PR #581, opened 2016-04-11
Flexible Ranges

jeffwilde:
I came across the need in one of my projects to create a range of nats that included the ending nat, and also a reversed range (counting down instead of up). Dealing with the extra syntax and type classes for this was a bit awkward, so I created my own type class that handled those two cases. 

I figured that might be useful to others, so I took the opportunity to extend the generality of of my new typeclass, add some fancy syntax, and provide it here if there is interest in integrating it back into shapeless.

The new type class, `FRange` accepts the `Bound` type class for the beginning and ending of the range (instead of bare `Nat`s), which allows the user to specify that the bound is either inclusive or exclusive. Along with some type aliases, this allows syntax like the following to create ranges:

```scala
scala> import nat._; import syntax.nat.range._
...
scala> the[_1 *--* _4].apply() == _1 :: _2 :: _3 :: _4 :: HNil
res1:  Boolean = true

scala> the[_1 :--* _4].apply() == _2 :: _3 :: _4 :: HNil
res2:  Boolean = true

scala> the[_1 :--: _4].apply() == _2 :: _3 :: HNil
res3:  Boolean = true
```
There's also fancy unicode aliases included:
```scala
scala> the[_1 ●--● _4].apply() == _1 :: _2 :: _3 :: _4 :: HNil
res4:  Boolean = true

scala> the[_1 ○--● _4].apply() == _2 :: _3 :: _4 :: HNil
res5:  Boolean = true

scala> the[_1 ○--○ _4].apply() == _2 :: _3 :: HNil
res6:  Boolean = true
```
Counting down also works with all variations:
```scala
scala> the[_4 ●--● _1].apply() == _4 :: _3 :: _2 :: _1 :: HNil
res8:  Boolean = true
```
The behaviour of the original `Range` type class can be duplicated with fancier syntax:
```scala
scala> the[_1 ●--○ _4].apply() == _1 :: _2 :: _3 :: HNil
res8:  Boolean = true
```

If anything doesn't match up with project standards I'm happy to change the syntax aliasing, shuffle around the package naming, change the line-breaks, or rip out the unicode support.

I also have a few outstanding bits that I wasn't able to find any way to work-into the change request. Any suggestions are welcome:
- [ ] The `FRange.SoftInclusive` trait and associated intermediate recursive implicits should actually be marked private, but it didn't seem possible with the way implicit resolution works.
- [ ] Ideally, I would have liked to alter the original `Range` class to accept the `Bound` type class (as well as the original `Nat`) instead of creating a whole new `FRange` type class. I couldn't figure a way to maintain binary compatibility with the existing Range trait, though. 
- [ ] I was going to add an `@implicitNotFound` annotation for `FRange`, but I couldn't think of a case in which this would be usefully displayed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ cross-builds for 2.10.6 and 2.12.x.
 + Huw Giddens <hgiddens@gmail.com>
 + Jason Zaugg <jzaugg@gmail.com> [@retronym](https://twitter.com/retronym)
 + Jean-Remi Desjardins <jeanremi.desjardins@gmail.com> [@jrdesjardins](https://twitter.com/jrdesjardins)
++ Jeff Wilde <jeff@robo.ai>
 + Jisoo Park <xxxyel@gmail.com> [@guersam](https://twitter.com/guersam)
 + Johannes Rudolph <johannes.rudolph@gmail.com> [@virtualvoid](https://twitter.com/virtualvoid)
 + Johnny Everson <khronnuz@gmail.com> [@johnny_everson](https://twitter.com/johnny_everson)
--- a/core/src/main/scala/shapeless/ops/nat.scala
+++ b/core/src/main/scala/shapeless/ops/nat.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-15 Miles Sabin 
+ * Copyright (c) 2011-16 Miles Sabin
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -245,7 +245,7 @@ object nat {
   }
 
   /**
-   * Type class witnessing that `Out` is range `A` to `B`.
+   * Type class witnessing that `Out` is range `A` to `B`, inclusive of `A` and exclusive of `B`.
    *
    * @author Andreas Koestler
    */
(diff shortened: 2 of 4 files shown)


jeffwilde:
I don't believe the build failure is related to this pull request. Once master is building again, I'll rebase or merge it back in.

jeffwilde:
Build looks good now.

milessabin:
My apologies for taking so long to get back to you :-(

This looks great ... thanks!

I'm mostly completely happy to merge this. As you probably guessed, your symbolic operators don't really fit with the general style in shapeless. I really like the unicode version ... I think that's one of the nicest renderings of open/closed I've seen. But for all that I think I'd prefer to limit things to ASCII. I think you're ASCII symbols are fine, but could you make them type aliases for non-symbolic type names?

As a general rule shapeless doesn't use access control, so on your first question ... not a problem.

On the second question, and the issue of the name `FRange`. I think it being a distinct type class from `Range` is fine. I think I would prefer the name `BoundedRange` ... I think it makes it clearer that the intent is a `Range` with a `Bound`.

jeffwilde:
Ten day turnaround is not at all a long time in the world of github pull requests! :)  
(not to mention you had your eyes on the larger prize of SI-2712 right when I submitted this)

All your suggestions sound good–I'll amend the pull request in a jiff.

jeffwilde:
I think that should take care of everything. Let me know if anything else pops out.

milessabin:
Good stuff. I think the only thing I'd change now, is where the aliases go. Why not move them to `object nat` in package `shapeless.syntax`?

jeffwilde:
That's a better place for the aliases, I'll move them.

One more change I was considering, changing `Inclusive` to `Inc` and `Exclusive` to `Exc` in order to cut down on the character count a bit. Let me know which way you lean:
```scala
import ops.nat.BoundedRange; import BoundedRange.{Inclusive, Exclusive}
the[BoundedRange[Inclusive[_1], Exclusive[_4]]].apply()
```
vs
```scala
import ops.nat.BoundedRange; import BoundedRange.{Inc, Exc}
the[BoundedRange[Inc[_1], Exc[_4]]].apply()
```

milessabin:
I'm happy with either. I don't think abbreviation does any harm here. Your call.

jeffwilde:
I decided against the name shortening in favour of keeping them as descriptive as possible. Shortened aliases for the two traits can always be added later if people actually end up using the word syntax over the symbols.

Otherwise, I believe we're ready to go!

milessabin:
Yeah, that looks good ... thank you!

Would you mind squashing before I merge? Unfortunately the new github squash feature doesn't give the option of squashing and creating a merge commit.

jeffwilde:
squashed!

milessabin:
Thank you! Merging :-)

PR #582, opened 2016-04-22
Revert to default CrossType.Pure

dwijnand:
The trade-off is that the path includes a dot, but at least it's a "default" type rather than another difference


milessabin:
I'm not super keen of having code, even test code, buried in a hidden directory. Can you motivate this for me?

dwijnand:
That is indeed the downside of the trade-off. The motivation is the alternative is built on - ie less speciale code. Bit it's just a suggestion from setting up my own scala.js project, I'd understand if you prefer the status quo.

dwijnand:
Another alternative is: keep switching to default CoreType.Pure, but then configure the build so that JVM test sources live in `src/test-jvm/scala`.

Summary:

* JVM test sources in current master: `core/jvm/src/test`
* JVM test sources in current proposal: `core/.jvm/src/test`
* JVM test sources in this proposal: `core/src/test-jvm`

milessabin:
`src/test-jvm/scala` might work.

dwijnand:
Alright, pushed.

milessabin:
Before I merge this ... are you absolutely sure that these tests will actually be run?

InTheNow:
Sorry to jump in at a late stage, but have you tested these changes on different IDE's, as they are all _very_ picky on where source lives relative to the project base (unfortunately)?

Another option (sorry if this is a pain...) is to treat the jvm/js code in the same fashion as 2.10/2.11 code. eg: 

```
core/src/test/scala/shapeless
core/src/test/scala/shapeless_jvm/shapeless
core/src/test/scala/shapeless_2.11+/shapeless
core/src/test/scala/shapeless_jvm_2.11+/shapeless
```

See [an old plugin](https://github.com/InTheNow/sbt-scalajs/blob/master/src/main/scala/SbtScalajs.scala#L236) for an example. I confess that I don't know how this would work with the IDEs today

InTheNow:
FWIW, these scheme is also extendable, just  imagine.....

```
core/src/test/scala_jvm9_2.11+/shapeless
core/src/test/scala_llvm_2.1_linux_centos_7_typlevel_2.11+/shapeless
```

dwijnand:
Making sure that tests actually run is indeed a concern for which I don't know how to automate away..

From inspecting the Travis builds, the one for [this PR](https://travis-ci.org/milessabin/shapeless/jobs/125221864) says:

```
[info] Compiling 42 Scala sources to /home/travis/build/milessabin/shapeless/core/.jvm/target/scala-2.11/test-classes...
[info] Passed: Total 548, Failed 0, Errors 0, Passed 548
```

while the one from [3 days ago](https://travis-ci.org/milessabin/shapeless/jobs/124521419) says:

```
[info] Compiling 43 Scala sources to /home/travis/build/milessabin/shapeless/core/jvm/target/scala-2.11/test-classes...
[info] Passed: Total 547, Failed 0, Errors 0, Passed 547
```

... so I'm not sure.

dwijnand:
*siiiiiiiigh* I'm going to not mess with this. Sorry for the noise.

```diff
diff --git a/core/src/test-jvm/scala/shapeless/lazy.scala b/core/src/test-jvm/scala/shapeless/lazy.scala
index 1575d17..ab2c684 100644
--- a/core/src/test-jvm/scala/shapeless/lazy.scala
+++ b/core/src/test-jvm/scala/shapeless/lazy.scala
@@ -115,6 +115,7 @@ class LazyStrictTestsJVM {
     assert(ccTC3SO)
     assert(genTC3SO)
     assert(listTC3SO)
+    assert(1 == 2)
   }
 }
```

```
> coreJVM/test
[info] Updating {file:/d/shapeless/}coreJVM...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Compiling 78 Scala sources to /d/shapeless/core/.jvm/target/scala-2.11/classes...
[info] Compiling 41 Scala sources to /d/shapeless/core/.jvm/target/scala-2.11/test-classes...
[info] Passed: Total 524, Failed 0, Errors 0, Passed 524
[success] Total time: 200 s, completed 23-Apr-2016 15:53:57
```

milessabin:
Hey ho :-(

PR #583, opened 2016-04-26
Add an unwrapped instance for shapeless tags

clhodapp:
By triggering unification, we can get the compiler to be less-good about computing tightest bounds. This allows us to strip off shapeless tags without writing a special-purpose macro.

Supersedes #538

### DIFF
--- a/core/src/main/scala/shapeless/unwrapped.scala
+++ b/core/src/main/scala/shapeless/unwrapped.scala
@@ -60,6 +60,10 @@ trait UnwrappedInstances extends LowPriorityUnwrappedInstances {
     chain: Strict[Unwrapped.Aux[UI, UF]]
   ) = chain.value.asInstanceOf[Unwrapped.Aux[Newtype[UI, Ops], UF]]
 
+  implicit def tagUnwrapped[T[UI, TT] <: tag.@@[UI, TT], UI, TT, UF](implicit
+    chain: Strict[Unwrapped.Aux[UI, UF]]
+  ) = chain.value.asInstanceOf[Unwrapped.Aux[T[UI, TT], UF]]
+
 }
 
 trait LowPriorityUnwrappedInstances {
--- a/core/src/test/scala/shapeless/unwrapped.scala
+++ b/core/src/test/scala/shapeless/unwrapped.scala
@@ -79,6 +79,23 @@ class UnwrappedTests {
   }
 
   @Test
+  def testShapelessTagged: Unit = {
+
+    import tag._
+
+    val pass = the[Pass[String @@ TestTag]]
+    the[pass.U =:= String]
+    val tagged = tag[TestTag]("testing")
+    val actual = pass.actual(tagged)
+    val wrapped = pass.wrapped(tagged: String)
+    val unwrapped = pass.unwrapped(tagged)
+    assert((actual: String @@ TestTag) == tagged)
+    assert((wrapped: String @@ TestTag) == tagged)
+    assert((unwrapped: String) == (tagged: String))
+    test.illTyped("unwrapped: String @@ TestTag")
+  }
+
+  @Test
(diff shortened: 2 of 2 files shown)


clhodapp:
MiMa's set all the way to "strict", it seems.

milessabin:
Damn. I guess those traits ought to be abstract classes. Can we do that retrospectively?

clhodapp:
I do not believe MiMa will be any happier with `trait` => `class` than it is with adding a method to a trait. I think if you want to preserve bincompat, this change will just have to wait.

clhodapp:
I guess you could also run a survey to find out if anyone has actually extended these traits. I would not be surprised if no one in the world has.

milessabin:
I've lost my bearings about what can go wrong here. Suppose we have library Foo which is built against shapeless-2.3.1 which (hypothetically) includes your change. Foo is then linked into an application which specifies shapeless-2.3.0 as its library dependency. What happens?

clhodapp:
In your scenario, the most likely outcome is that sbt sees two conflicting version of `"com.chuusai" %% "shapeless"` in your transitive library dependencies and performs an eviction of the older version. It is also possible to configure sbt to use other strategies, such as "use the lowest version", "use this insane strategy I just made up" or "blow up". People who have made these sorts of configuration changes are generally very aware of the fact that they have done so because the situation of a library version conflict somewhere in your transitive dependencies is _very_ common, so their strategy is going to be invoked pretty frequently.

As far as what happens to the actual code at runtime... So... in your scenario, the most likely outcome is for it to be just fine. Theoretical shapeless 2.3.1, which hypothetically has my change, has a version of `Unwrapped` that fully implements the interface from the real shapeless 2.3.0. Therefore, any implicitly-inserted calls to its methods should link up and run just fine. However, that isn't the only possible outcome.

The first failure case would happen if the user were to force shapeless 2.3.0 onto their classpath when they also have libraries that were built against 2.3.1 and use this new method. In this case, the user is likely to get a `MethodNotFoundException`. This _could be_ a problem but... we generally discourage users from doing this, as it's pretty uncommon to ensure both forwards and backwards binary compatibility between even minor versions, which is what's needed to prevent this... Ensuring this kind of compatibility is kind of a nightmare, as it pretty much means you can't introduce new public methods (or even certain kinds of private ones, I think). Pretty much the only project that takes on this onus, as far as I know, is Scala itself.

The other failure case (the one I think MiMa is actually trying to protect you from, now that I've thought about it) is the one where the user has extended the trait I'm adding a method to (likely to add their own instances of higher priority in a companion or package object somewhere). I believe that traits with concrete members are compiled down into both a pure interface and a class (to hold the implementations). I think that classes that extend the trait compile to classes/interfaces that implement the interface and have forwarders to a private member to allow the method implementations to be called. This creates a bincompat problem when when you update the definition of the trait later to have a new method: suddenly, all those classes that extend the trait *claim* to implement the new trait interface, but they actually don't (they don't have a forwarder for your new method, even if the trait theoretically provides an implementation). In short, if anyone extended the Unwrapped instance traits to provided their own higher-priority instances, their code would be broken if this PR were to be merged. Also, for what it's worth, I believe that this exact same set of people would have an equivalent type of trouble if were were to change the traits to abstract classes.

milessabin:
OK, I think that means that it's reasonably safe to add a MiMa exclusion for this. I'll flag it up in the release notes.

milessabin:
Oh, damn ... I've just merged #538. What should I do here? Can you update this PR with the MiMa exclusion and include a revert of the #538 commit?

milessabin:
Hmm ... it seems that #538 also breaks binary compatibility. Leave it to me ... I'll revert that, merge this and deal with MiMa.

clhodapp:
Great. I'm actually still kind of surprised that this ended up working, but it does so yay.

PR #586, opened 2016-05-02
improve `illTyped`

larsrh:
* catch parse errors (to distinguish them from type errors)
* blackbox context suffices
* don't abort macro expansion, just report an error


larsrh:
Failure are binary compatibility problems, but I'm not sure whether they're true.

milessabin:
The will be OK if you switch back to whitebox ... maybe open an issue to change to blackbox in 2.4.0?

larsrh:
Done and done (#587).

milessabin:

dwijnand:
cool!

PR #588, opened 2016-05-09
Add FieldType type member to FieldOf (fixes #117).

### LINKED ISSUE #117: Add a field type member to FieldOf trait and use in tests.

peterneyens:
I have added a `FieldType` type member to `FieldOf`.

### DIFF
--- a/core/src/main/scala/shapeless/labelled.scala
+++ b/core/src/main/scala/shapeless/labelled.scala
@@ -79,6 +79,8 @@ trait FieldPoly extends Poly1 {
 trait FieldOf[V] {
   import labelled._
 
+  type F = FieldType[this.type, V]
+
   def ->>(v: V): FieldType[this.type, V] = field[this.type](v)
 }
 
--- a/core/src/test/scala/shapeless/records.scala
+++ b/core/src/test/scala/shapeless/records.scala
@@ -94,14 +94,14 @@ class RecordTests {
   @Test
   def testFieldAt {
     val r1 =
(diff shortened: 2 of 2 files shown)


core/src/main/scala/shapeless/labelled.scala
```diff
@@ -79,7 +79,9 @@ trait FieldPoly extends Poly1 {
 trait FieldOf[V] {
   import labelled._
 
-  def ->>(v: V): FieldType[this.type, V] = field[this.type](v)
+  type FieldType = labelled.FieldType[this.type, V]
+
+  def ->>(v: V): labelled.FieldType[this.type, V] = field[this.type](v)
```
peterneyens: If I try to specify `FieldType` as the return type, `v` is required to be `Nothing`. I don't know if we can fix this ?

```
[error] /home/peter/programming/shapeless/core/src/main/scala/shapeless/labelled.scala:85: type mismatch;
[error]  found   : v.type (with underlying type V)
[error]  required: Nothing
[error]   def ->>(v: V): FieldType = (field[this.type](v): FieldType)
```

milessabin: Interesting. I don't see any obvious reason for that.

It's also annoying that you have to qualify `FieldType` here. Maybe it would be better to go with `type F`? It's being used primarily as an abbreviation, so a single character name would be justified I think.

If you do that then I think the original definition of `->>` should be fine.

milessabin:

Issue #112, opened 2014-06-25
Add permutation methods to HList/Coproduct/tuple.


milessabin:

teldosas:
Is anything more to be done here? It seems that permutation methods have been added both for tuples (#589) and Coproducts (#263).

milessabin:
Well spotted. Yes, this is complete now. Thanks for flagging it up.

PR #589, opened 2016-05-10
Add tuple Align ops and syntax based on HList Align.

peterneyens:
Completes  #112 together with  #263.


milessabin:

PR #590, opened 2016-05-11
Add combinations method for HList

ryoppy:


milessabin:
This is very cool, and I'd be happy to merge it on purely technical grounds. Before I do, can you give an example of a use that would motivate including it in the main library as opposed to, say, as an example?

ryoppy:
I can't give an example. well... I wrote some algorithms in a type level. it's just hobby. I needed a typeclass of combinations for those. and then I found a permutations method in HList, so I was think that a combinations method is also useful for shapeless.

milessabin:
Fair comment wrt permutations.

OK, I'm happy to include this. But I wonder if it would make sense to move `Combinations0` to the top level: if I'm reading it correctly this represents "combinations of some fixed length `N`" and that strikes me as being at least, if not more useful than "all combinations".

ryoppy:
Why do you think 'useful as "all combinations"' ?
Combination0 connot check length alone. I think that it's not necessary move to the top level.

milessabin:
Am I misunderstanding? I read `Combinations0[N, L]` as producing all combinations of elements from `L` of length exactly `N`.

ryoppy:
No you're right. ah.. come to think of it, it could be nice. so, I move it to the top level.

ryoppy:
moved :)

milessabin:
I'm afraid I'm getting a bit confused ... my apologies :-(

I think I misunderstood initially and thought that `Combinations` produced all combinations (or all lengths) of elements from `L`, whereas `Combinations0` produced all combinations of exactly length `N` of elements from `L`. It was on that (mis)understanding that I asked you to move `Combinations0` to the top level.

But looking again now, it seems that `Combinations` is already producing combinations of exactly `N` elements. So what is `Combinations0` doing? It has exactly the same signature as `Combinations`, so why is a distinct type class needed here?

ryoppy:
I wanted check length the same as other type classes, like a "At type class" or "Slice type class". Because I was think that it's more useful. But "Combination" cannot check length alone, so I added "Combination0". I had no choice but to do so...

And I wrote the wrong comment by mistake... my apologies :(
https://github.com/milessabin/shapeless/pull/590/files#diff-03fe632291a854a01a374f4b97f30489R720

ryoppy:
I removed Combination0. This version is cannot check length. like [this](https://github.com/milessabin/shapeless/pull/590/commits/92f2d5e3fede4ab189db686620fa175fe4856e1a#diff-1f14a875ec23de285bb153d924d7e6c2R3303). but simple.
Which do you think is better this version or an old version.

milessabin:
My apologies for taking so long to get back to you on this :-(

This looks just fine ... many thanks! Merging :-)

Atry:
I guess it may be useful for random testing

PR #592, opened 2016-05-11
Fix #591 Add HNil ZipConst

### LINKED ISSUE #591: ZipConst is missing HNil case
Shouldn't the following work? 

```scala
HList().zipConst("anything")
```

doolse:
Added the HNil case and ran the tests, seems fine.


milessabin:

PR #593, opened 2016-05-15
Add the PolyFill type class

Atry:
I added this type class according to the talk about `Zero` on Gitter, May 09.


milessabin:
Thanks for this!

I think that `fillWith` would be a name that's more consistent with conventions used elsewhere in the library. I also think it would be preferable to provide the poly as a value argument rather than as a type argument.

If you make those two changes, the use in the test would end up looking like this,
```scala
HList.polyWith[Int :: String :: Int :: HNil](zero)
```
which I think is preferable.

PR #594, opened 2016-05-15
Configure external links in ScalaDoc

Atry:
I found that Shapeless's [Scaladoc](https://oss.sonatype.org/service/local/repositories/releases/archive/com/chuusai/shapeless_2.11/2.3.1/shapeless_2.11-2.3.1-javadoc.jar/!/index.html) does not link to Scala standard library and other dependencies. This patch fixes the problem.


milessabin:
Thanks!

PR #551, opened 2016-02-25
Adds scoverage support

khronnuz:
Fixes #411 

Preview of how the report looks like is at https://codecov.io/github/khronnuz/shapeless?ref=9a4bc2936bff479547e4cd2bcfbf52c9a78ad163&view=all

PR #595, opened 2016-05-17
Add compileTime macro to measure compilation times

fthomas:
`compileTime` is a simple macro that measures the compilation
time of a code fragment. Like `illTyped` it takes Scala code as
string, measures the time it takes to parse and typecheck it and
returns that time as a `Duration`.

Example usage:
```scala
scala> test.compileTime("Generic[(Int, Option[String])]")
res0: scala.concurrent.duration.FiniteDuration = 34911863 nanoseconds
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
     - $HOME/.sbt
 
 script:
-  - travis_wait 40 scripts/travis-publish.sh
+  - travis_wait 45 scripts/travis-publish.sh
 
 env:
   global:
--- a/core/src/main/scala/shapeless/test/compiletime.scala
+++ b/core/src/main/scala/shapeless/test/compiletime.scala
@@ -23,6 +23,15 @@ import scala.reflect.macros.blackbox
 
 /**
  * Utility that measures the compilation time of a code fragment.
+ *
+ * `compileTime` takes a code fragment as `String`, measures the time
+ * it takes to parse and typecheck it and returns that time as a
+ * `FiniteDuration`.
+ *
+ * Example: {{{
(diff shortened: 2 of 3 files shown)


milessabin:
Awesome! Thank you!

Seeing as there isn't (and can't easily be) a test for this, would you mind putting your description above into a doc comment?

InTheNow:
For the sake of [scoverage that will land soon ](https://github.com/milessabin/shapeless/pull/551), is it perhaps worth adding a "test" that just calls the "code" (including bad code) and checks for a duration `> 0`?

fthomas:
Yes, that is a good idea. I'll add a test and improve the docstring tonight.

fthomas:
I've added a test and updated the documentation. Unfortunately the build for 2.10 failed because it timed out. However this looks unrelated to this change since the build was aborted during `packageLocal`. Could you please rebuild or increase the timeout?

fthomas:
I increased the timeout from 40 to 45 minutes and now it has successfully built in under 30 minutes. :-)

milessabin:
Awesome! Thanks. Yes, the Travis timeout has been borderline for a little while now.

fommil:
@fthomas this could be really amazing in ENSIME to disable slow files. Can you please take a look at https://github.com/ensime/ensime-server/issues/954 and see if you can come up with any ideas?

fommil:
btw, I'm aware this is for code snippets. I was thinking in terms of writing a compiler plugin along the lines of what Jason recommends in that ticket for the source code files that are being compiled.

fthomas:
@fommil I looked at that ticket but that is way above my head.

PR #601, opened 2016-05-28
Fix #600

### LINKED ISSUE #600: Sized.equals fails in 2.3.1
```
scala> Sized(1, 2, 3) == Sized(1, 2, 3)
res4: Boolean = false
```

This returned true in 2.2.5 but returns false in 2.3.1.

Atry:


milessabin:
Unfortunately we can't make `Sized` a case class because we have a custom `apply` and `unapplySeq` which conflict with the ones provided as case class defaults.

PR #602, opened 2016-05-28
RecordArg, inverse of RecordArgs trait

beezee:
This adds a RecordArg trait, which is essentially the inverse of the RecordArgs trait. Specifically, instead of allowing forward from "method" to "methodRecord" where method can be passed named parameters that will be rolled into a record for the forwarded call, this allows forwarding from "methodRecord" to "method", where the signature of method is used to access required parameters from a record passed to the original call.

This allows a more typesafe application of functions that may take a series of similarly typed parameters, so that parameter order cannot be mistaken in such a way as to produce errors or incorrect behavior at runtime.

EG - https://gist.github.com/beezee/fe604473c557564f9a6c1be06bc0960d


beezee:
There's some weirdness with the commit history here, and a leftover, unused import. Closing to amend and re-open.

PR #603, opened 2016-05-28
RecordArg, inverse of RecordArgs trait

beezee:
This adds a RecordArg trait, which is essentially the inverse of the RecordArgs trait. Specifically, instead of allowing forward from "method" to "methodRecord" where method can be passed named parameters that will be rolled into a record for the forwarded call, this allows forwarding from "methodRecord" to "method", where the signature of method is used to access required parameters from a record passed to the original call.

This allows a more typesafe application of functions that may take a series of similarly typed parameters, so that parameter order cannot be mistaken in such a way as to produce errors or incorrect behavior at runtime.

EG - https://gist.github.com/beezee/fe604473c557564f9a6c1be06bc0960d

Re-open of #602 after correcting commit history and cleaning up a stray import and definition.

### DIFF
--- a/core/src/main/scala/shapeless/records.scala
+++ b/core/src/main/scala/shapeless/records.scala
@@ -88,6 +88,29 @@ trait RecordArgs extends Dynamic {
   def applyDynamicNamed(method: String)(rec: Any*): Any = macro RecordMacros.forwardNamedImpl
 }
 
+/**
+ * Trait supporting mapping record arguments to named argument lists, inverse of RecordArgs.
+ *
+ * Mixing in this trait enables method applications of the form,
+ *
+ * {{{
+ * lhs.methodRecord('x ->> 23 :: 'y ->> "foo" :: 'z ->> true :: HNil)
+ * }}}
+ *
+ * to be rewritten as,
+ *
+ * {{{
+ * lhs.method(x = 23, y = "foo", z = true)
+ * }}}
+ *
+ * ie. the record argument is used to look up arguments for a target method
+ * (the called method named minus the "Record" suffix) by name and type and the application
+ * is rewritten to an application of the target method
+ */
+trait RecordArg extends Dynamic {
+  def applyDynamic[L <: HList](method: String)(rec: L): Any = macro RecordMacros.forwardFromRecordImpl[L]
+}
+
 @macrocompat.bundle
 class RecordMacros(val c: whitebox.Context) {
   import c.universe._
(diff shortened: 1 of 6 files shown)


core/src/main/scala/shapeless/records.scala
```diff
@@ -152,4 +193,17 @@ class RecordMacros(val c: whitebox.Context) {
       case(elem, acc) => q""" $hconsValueTree(${promoteElem(elem)}, $acc) """
     }
   }
+
+  def mkParamsImpl[L <: HList](method: MethodSymbol, rec: Expr[L]): List[Tree] = {
+    def mkElem(keyTpe: Type, value: Tree): Tree =
+      q"implicitly[shapeless.ops.hlist.Selector[${rec.actualType}, ${mkFieldTpe(keyTpe, value.tpe)}]].apply($rec)"
```
fthomas: @beezee A general remark when writing macros: Non-interpolated terms and types in quasiquotes (like `implicitly` and `Selector`) should be fully qualified to avoid macro hygiene issues. So in this case you should use `_root_.scala.Predef.implicitly` and `_root_.shapeless.ops.hlist.Selector`.

beezee: Thanks @fthomas that makes sense. I just pushed an update accordingly.

milessabin:
Awesome! Thanks!

I think we need a different name for this though ... `RecordArg` is a bit too close to `RecordArgs` for comfort. I guess that `RecordArgs` should probably have been named `ToRecordArgs` in which case it would make sense to call this `FromRecordArgs`?

Also, I think it would make sense to add a positional variant of this (ie. corresponding to `ProductArgs` rather than `RecordArgs`) ... given that that would most likely share some implementation with this (or if not, at least have a very similar flavour) would you mind adding that too?

beezee:
@milessabin makes sense. I renamed mine as suggested, leaving the original as it was. It occurred to me there's no reason to limit to a single param list, so I adjusted to support it.

I also added FromProductArgs. There's one thing that's bugging me about it right now, which is that it currently ignores extraneous elements on the incoming HList. That is, if the first n elements align with a target method signature, the application is successful and any additional elements are just ignored.

I think it would be preferable to make this a compile time error, but I struggled to get a stable value representing the runtime length of the incoming HList expression for comparison to the target method. Do you know of a way to do this?

beezee:
Weird I did not see this test failure locally. Seems to be a serialization related issue. Looking into it.

milessabin:

PR #608, opened 2016-06-01
Fully qualify scala.Symbol in quasiquote

fthomas:


fthomas:
This looked like a hygiene issue to me, but apparently it is not. I don't know why `scala.Symbol` cannot fully qualified in this extractor quasiquote.

milessabin:
I agree it looks suspicious. I can't remember now if it was deliberate. Were you able to come up with a use site example of unhygiene?

fthomas:
No, I haven't observed any problems at the use site yet. I'll further investigate and report my findings here. In the meantime, feel free to close this PR.

milessabin:
OK, thanks for investigating :-)

PR #610, opened 2016-06-07
Try not to let HNil.type get into the wild

doolse:


alexarchambault:
Alternatively, the `sealed trait HNil` could be removed, and `case object HNil extends HNil` changed to `case object HNil extends HList`, with a type alias `type HNil = HNil.type`.

milessabin:

dwijnand:
Doesn't making the HNil object private mean you can't match `case HNil` anymore and you have to `case _: HNil`?

milessabin:
Yes, that would follow. I'm open to being persuaded that that's not important however ... pattern matching on an `HList` isn't a particularly common operation. I think that's a big enough change that I'd want to punt to 3.0.0.

@doolse can you make a compelling argument in favour of this change? Otherwise I think I'll close.

milessabin:
Closing for now.

PR #613, opened 2016-06-24
Add backtracking example

OlivierBlanvillain:
This PR adds an example of prolog style programing which scala implicits, and illustrates a limitation of the implicit resolution algorithm which does not do backtracking. It then shows a possible workaround of this limitation where the backtracking is "manually" implemented by storing the result of the search in an HList.


milessabin:
Looks good. Could you put it in package shapeless.examples?

OlivierBlanvillain:
Ouch run time coverage is not very useful here... Shall I add something like this:?

```
// $COVERAGE-OFF$This is a compile time only example
```

milessabin:
There are issues with the coverage reports ... I'll merge this as is when Travis goes green.

milessabin:

PR #615, opened 2016-07-03
Fix #604

### LINKED ISSUE #604: Typeable.referenceSingletonTypeable doesn't work after serialization
If a `Typeable.referenceSingletonTypeable` is serialized and deserialized, it stops working, i.e., it rejects the original value (full example: https://github.com/durban/akka-typed-examples/blob/4be6d20ebcc19bc5d687da0e9ff610b2b4052749/src/main/scala/com/example/shless/SingletonTpe.scala).

The reason (probably) is that the value is also serialized and deserialized, and then checked for reference equality (and serialization doesn't preserve that in the general case). To be honest, at the moment I don't see a way to implement this correctly. But since Java serialization is already quite fragile, it's probably not a serious problem. However, it would be nice if these `Typeable`s would at least reject serialization (that way at least it would fail early, with an understandable error).

durban:
Fix #604 by rejecting serialization of `referenceSingletonTypeable`s by default.

However, `object`s and symbols seem to preserve their identity after a serialization/deserialization, so `Typeable`s for them can work correctly after a deserialization (there is also a test for this). Because of this, instances generated for these, do not reject serialization.

### DIFF
--- a/core/jvm/src/test/scala/shapeless/serialization.scala
+++ b/core/jvm/src/test/scala/shapeless/serialization.scala
@@ -89,6 +89,23 @@ object SerializationTestDefns {
     assertSerializable(t)
   }
 
+  def roundtrip[M](m: M): M = {
+    val baos = new ByteArrayOutputStream()
+    val oos = new ObjectOutputStream(baos)
+    try {
+      oos.writeObject(m)
+    } finally {
+      oos.close()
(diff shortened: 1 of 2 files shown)


milessabin:
LGTM ... I'll merge as soon as Travis goes green :-)

PR #616, opened 2016-07-09
Add HasNoGeneric

OlivierBlanvillain:
This PR adds a `HasNoGeneric[T]` type class to witness that `T` has no `Generic` instance. With this addition, it becomes possible to implement a `trait GenericRep[T] { type Repr }` which fully "unrolls" all the `Generic` macros to have the full generic representation of a type. [Example gist](https://gist.github.com/OlivierBlanvillain/4163a3b4cd615a37f62a5006fe7dc34b).

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -264,6 +264,12 @@ object HasCoproductGeneric {
   implicit def apply[T]: HasCoproductGeneric[T] = macro GenericMacros.mkHasCoproductGeneric[T]
 }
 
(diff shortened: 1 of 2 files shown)


OlivierBlanvillain:
I did not notice that `mkProductGeneric` and `mkCoproductGeneric` had a bunch of `abort`s which makes it possible to have neither a `Generic` nor a `HasNoGeneric` instance with this implementaion. I  will persevere in trying to implement the negation of having a `Generic` instance using implicit priority / ambiguity and come up with a more thought out PR it that turns out to be impossible...

PR #617, opened 2016-07-13
`Builder` and `build` method to build an HList from a type and a Poly

stephenjudkins:
This seems like a reasonable utility method; it makes doing a couple things a lot cleaner. I don't know if the `build`/`Builder` names are what we want.

If people think this should be in Shapeless core, I'm happy to add docs and make any changes.

### DIFF
--- a/core/src/main/scala/shapeless/hlists.scala
+++ b/core/src/main/scala/shapeless/hlists.scala
@@ -89,6 +89,8 @@ object HList extends Dynamic {
     */
   def fillWith[L <: HList] = new FillWithOps[L]
 
+  def build[H <: HList](f: Poly)(implicit builder: Builder[f.type, H]): H = builder()
+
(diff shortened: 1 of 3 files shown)


milessabin:
Is this what we were talking about the other day? I'm confused, because it _does_ use inductive implicit, which I thought you were arguing against?

I agree that this is a useful operation though ... in fact we already have it: `fillWith` ... see [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/hlists.scala#L90) with examples of use in [the tests](https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/hlist.scala#L2819-L2828).

stephenjudkins:
Yes, that is exactly what I was looking for. I think we misunderstood each other: I wanted a helper method that didn't require me to write inductive implicits *myself*, not that didn't use one under the hood.

milessabin:
Sorry for being slow on the uptake :-(

PR #621, opened 2016-07-28
Fix type loss issue with `LiftAll`

jeremyrsmith:
I did this because `LiftAll.Aux` can't always resolve correctly with the current implementation.  Because the type `tailInstances.Out` is not a type argument to `hcons`, scalac has trouble resolving implicit `LiftAll.Aux[F, L, I]` if `I` is to be inferred.  With this small change, `I` can always be inferred (assuming all elements of `L` have instances for `F`)


triggerNZ:
LGTM

milessabin:

Issue #622, opened 2016-07-29
Serialisation of type inequality traits fails


AndreasKostler:
The type inequality traits `=:!=` and `<:!<` do not extend `Serializable` and will therefore throw and exception when used in a context where serialisation is required:

```scala
val oos = new ObjectOutputStream(System.out)
oos.writeObject(new =:!=[Int, Doube]{}) /// BOOM!
```

milessabin:
I think this is desirable, or at least not a bad thing: the proof embodied in the value wouldn't survive being serialized and deserialized because of erasure.

AndreasKostler:
Oh...Do you have any suggestion on how to use this in a context where serialisation is required? I.e. is there a way to provide that proof?

milessabin:
Could you give me an example of how you would want to use this before and after serialization?

AndreasKostler:
This is what happens...
```scala
Exception in thread "main" org.apache.spark.SparkException: Task not serializable
    at org.apache.spark.util.ClosureCleaner$.ensureSerializable(ClosureCleaner.scala:304)
    at org.apache.spark.util.ClosureCleaner$.org$apache$spark$util$ClosureCleaner$$clean(ClosureCleaner.scala:294)
    at org.apache.spark.util.ClosureCleaner$.clean(ClosureCleaner.scala:122)
    at org.apache.spark.SparkContext.clean(SparkContext.scala:2021)
    at org.apache.spark.rdd.RDD$$anonfun$map$1.apply(RDD.scala:314)
    at org.apache.spark.rdd.RDD$$anonfun$map$1.apply(RDD.scala:313)
    at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:147)
    at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:108)
    at org.apache.spark.rdd.RDD.withScope(RDD.scala:306)
    at org.apache.spark.rdd.RDD.map(RDD.scala:313)
    at commbank.grimlock.spark.ReducibleMatrix$class.melt(Matrix.scala:749)
    at commbank.grimlock.spark.Matrix5D.melt(Matrix.scala:1438)
    at commbank.grimlock.spark.examples.Conditional$.main(Conditional.scala:65)
    at commbank.grimlock.spark.examples.Conditional.main(Conditional.scala)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #623, opened 2016-07-29
Inequality traits should be serialisable

AndreasKostler:


milessabin:
Much appreciated, but see my comments on #622 ... I'd be happy to reconsider if you can persuade me that this is safe.

milessabin:
Or that even if it isn't safe that there's an overriding reason why we should do it anyway :-)

milessabin:
I'm now persuaded that this is reasonable ... thanks for your persistence ... merging :-)

Issue #624, opened 2016-07-29
`Out` type of `ProductLength` is not a `Nat`


AndreasKostler:
This won't compile

``` scala
trait WantsNat[N <: Nat]

def foo[P <: Product](p: P)(implicit pl: ProductLength[P]): WantsNat[pl.Out] = new WantsNat[pl.Out] {}
```

milessabin:
Good catch. There is a workaround though,
```scala
def foo[P <: Product](p: P)(implicit pl: ProductLength[P] { type Out <: Nat }): WantsNat[pl.Out] =
  new WantsNat[pl.Out] {}
```

AndreasKostler:
That works, kinda breaks symmetry with `HList` `Length` though.

milessabin:
Agreed ... it's a workaround, not the right answer :-)

milessabin:
The fix breaks binary compatibility so this is going to have to wait for 2.4.0.

Ssstlis:
Should we close this issue 'cause of merged #973 ?

joroKr21:
Oh yes thanks fixed by #973 👍

PR #625, opened 2016-07-29
product length is Nat

AndreasKostler:


milessabin:
Given that there's a [workaround](https://github.com/milessabin/shapeless/issues/624#issuecomment-236359843) I think I'd like to be very confident that it's safe to add a MiMa exclusion rule for the bincompat failure that this introduces.

Would you mind generating a binary which exercising `ProductLength` with 2.3.1 then run (without recompilation) it again with a shapeless build which includes this fix. If it's fine then we can add the MiMa rule. If it isn't we'll need to postpone this until 2.4.0.

AndreasKostler:
Miles, I am a bit lost. What do I need to do?

milessabin:
Try something like this ...

 1. Put together a minimal sbt project which exercises `length` compiled against 2.3.1.
 2. Build a version of 2.3.1 with your change included.
 3. Run the binaries built in (1) using the binaries built in (2).

 That will test backwards binary compatibility (old code successfully links to new library). We also need to test forward binary compatibility (new code successfully links to old library). For that ...

 4. Locally publish a version of 2.3.1-SNAPSHOT with your change included.
 5. Build the project from (1) against 2.3.1-SNAPSHOT.
 6. Run the binaries build in (5) using the standard published 2.3.1.

milessabin:

AndreasKostler:
I think it's never been dead really - more like undead. Given it breaks binary compat should this be postponed to 2.4.0?

milessabin:
Yes, I think so ...

joroKr21:
@AndreasKostler sorry for the long wait.
2.4.0 is coming soon. Care to rebase this PR?

joroKr21:
Thanks for the contribution! Merged in #973

PR #627, opened 2016-08-09
Added `UnzipFunctions` typeclass

jeremyrsmith:
This typeclass witnesses that an `HList` type `L` relates two other
`HList` types `Args` and `Returns`, where each element in `L` is of the
form `A => B`, where `A` is the corresponding element from `Args`, and
`B` is the corresponding element from `Returns`. In some ways it
complements `ZipApply`, but to use `ZipApply`, you must already know the
type of `Args`. `UnzipFunctions` allows inferring the `Args` and
`Returns`, either with just `L`, or with `L` and one of `Args` and
`Returns`, or with all three.


milessabin:
This looks great! Thanks :-)

Do you think it would be useful to add an example illustrating use and contrasting with `ZipApply`?

jeremyrsmith:
@milessabin it certainly would! Where should such an example be added? The organization of the examples isn't that straightforward to me 😞

milessabin:
A new example source file under `examples` would be fine.

jeremyrsmith:
So, in the course of writing the example, I realized that `UnzipFunctions` wasn't really necessary. I added it with the idea that it would allow you to infer the argument HList and result HList, without realizing that `ZipApply` can already be used for that purpose.

But, the example of `ZipApply` might still be useful, so I'll leave this PR up with just the example.

milessabin:
Great stuff! Thanks :-)

I don't think the test is necessary though. There's a `runAll` SBT task which runs all the tests and this is exercised by CI.

milessabin:
Hmm ... sorry! I thought I'd merged this for 2.3.2 :-(

I know it probably sounds weird, but I think it would be preferable to drop the test and have the object extend `App` ... that means that it would be tested by Travis.

Would you mind doing that before I merge?

milessabin:
I can't apologise enough for how long this has been sitting in the queue to be merged ... I have no idea how I missed it :-(

Anyhow, LGTM, many (belated) thanks ... merging :-)

PR #628, opened 2016-08-20
Added type class to support zipping a Coproduct with a constant ('zipConst' method)

harveywi:


milessabin:

For consistency with the existing `zipConst` on `HList` would you mind swapping the order of the resulting pairs, ie. to `(<coproduct element type>, <type of constant>)`.

harveywi:
Thanks Miles - these are my first pull requests, and I appreciate your patience and advice.

milessabin:

PR #629, opened 2016-08-20
Added type class to support zipping a Coproduct with an HList ('zipWithHList' method)

harveywi:


milessabin:
Also great stuff :-)

Same comment here ... would you remind swapping the order of the resulting pairs.

I'd also prefer a name without `HList` in it ... would just plain `ZipWith`/`zipWith` work or is that already taken?

harveywi:
It turns out that this `zipWith` method enables at least one nice lightweight/idiomatic technique for folding `Coproduct`s without having to resort to a non-exhaustive match over canonical injectors `Inl` \ `Inr` or a heavyweight polymorphic function requiring a stable identifier (cf. [here](http://stackoverflow.com/q/34107849)).  If it is okay, I can submit a pull request for that too sometime (hopefully this week).

milessabin:

I'd be very interested to see how the coproduct fold plays out with this ... maybe as an example?

PR #631, opened 2016-08-26
Add type logic gates

yilinwei:
Within our code base I've found it useful to sometimes have the ability to do OR, AND, NOT, XOR on a typelevel for evidences. AND is usually fine as you simply need 2 parameters, but OR, XOR and NOT can sometimes cause a few issues.

I'm unsure whether these belong in shapeless at all and this is merely a speculative PR.

### DIFF
--- a/core/src/main/scala/shapeless/package.scala
+++ b/core/src/main/scala/shapeless/package.scala
@@ -106,8 +106,7 @@ package object shapeless {
     val b = Some(evb)
   }
 
-  implicit def abXorAmbigTpe0[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
-  implicit def abXorAmbigTpe1[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
+  implicit def abXorAmbigTpe[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
 
   trait !![A] extends Serializable
 
--- a/core/src/main/scala/shapeless/package.scala
+++ b/core/src/main/scala/shapeless/package.scala
@@ -59,6 +59,61 @@ package object shapeless {
   type ∃[P[_]] = P[T] forSome { type T }
   type ∀[P[_]] = ¬[∃[({ type λ[X] = ¬[P[X]]})#λ]]
 
+  //Type logic-gates
+  trait ||[A, B] extends Serializable {
+    def a: Option[A]
+    def b: Option[B]
+  }
+
+  object || {
+    implicit def aOrTpe[A, B](implicit eva: A): A || B = new ||[A, B] {
+      val a = Some(eva)
+      val b = None
+    }
+    implicit def bOrTpe[A, B](implicit evb: B): A || B = new ||[A, B] {
+      val a = None
+      val b = Some(evb)
+    }
+  }
+
+  implicit def abOrTpe[A, B](implicit eva: A, evb: B): A || B = new ||[A, B] {
+    val a = Some(eva)
+    val b = Some(evb)
+  }
+
+  trait &&[A, B] extends Serializable {
+    def a: A
+    def b: B
+  }
+
+  implicit def abAndTpe[A, B](implicit eva: A, evb: B): A && B = new &&[A, B] {
+    val a = eva
+    val b = evb
+  }
+
+  trait ^^[A, B] extends Serializable {
+    def a: Option[A]
+    def b: Option[B]
+  }
+
+  implicit def aXorTpe[A, B](implicit eva: A): A ^^ B = new ^^[A, B] {
+    val a = Some(eva)
+    val b = None
+  }
+
+  implicit def bXorTpe[A, B](implicit evb: B): A ^^ B = new ^^[A, B] {
+    val a = None
+    val b = Some(evb)
+  }
+
+  implicit def abXorAmbigTpe0[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
+  implicit def abXorAmbigTpe1[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
+
+  trait !![A] extends Serializable
+
+  implicit def nNotTpe[A](implicit eva: A): !![A] = unexpected
+  implicit def notTpe[A]: !![A] = new !![A] {}
+
   /** `Optic` definitions */
   val optic = OpticDefns
   val lens = OpticDefns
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/package.scala
```diff
@@ -59,6 +59,60 @@ package object shapeless {
   type ∃[P[_]] = P[T] forSome { type T }
   type ∀[P[_]] = ¬[∃[({ type λ[X] = ¬[P[X]]})#λ]]
 
+  //Type logic-gates
+  trait ||[A, B] extends Serializable {
+    def a: Option[A]
+    def b: Option[B]
+  }
+
```
milessabin: Can you motivate this for me? Why is this preferable to `Either[A, B]`?

core/src/main/scala/shapeless/package.scala
```diff
@@ -59,6 +59,60 @@ package object shapeless {
...
+      val a = Some(eva)
+      val b = None
+    }
+    implicit def bOrTpe[A, B](implicit evb: B): A || B = new ||[A, B] {
+      val a = None
+      val b = Some(evb)
+    }
+  }
+
+  implicit def abOrTpe[A, B](implicit eva: A, evb: B): A || B = new ||[A, B] {
```
milessabin: Why is this instance not in `object ||` along with the others?

core/src/main/scala/shapeless/package.scala
```diff
@@ -59,6 +59,60 @@ package object shapeless {
...
+    val a = Some(eva)
+    val b = Some(evb)
+  }
+
+  trait &&[A, B] extends Serializable {
+    def a: A
+    def b: B
+  }
```
milessabin: And motivation here ... why not `(A, B)`?

core/src/main/scala/shapeless/package.scala
```diff
@@ -59,6 +59,60 @@ package object shapeless {
...
+  implicit def abAndTpe[A, B](implicit eva: A, evb: B): A && B = new &&[A, B] {
+    val a = eva
+    val b = evb
+  }
+
+  trait ^^[A, B] extends Serializable {
+    def a: Option[A]
+    def b: Option[B]
+  }
+
```
milessabin: OK, this is `Either[A, B]` ...

core/src/main/scala/shapeless/package.scala
```diff
@@ -59,6 +59,60 @@ package object shapeless {
...
+  implicit def bXorTpe[A, B](implicit evb: B): A ^^ B = new ^^[A, B] {
+    val a = None
+    val b = Some(evb)
+  }
+
+  implicit def abXorAmbigTpe[A, B](implicit eva: A, evb: B): A ^^ B = unexpected
+
+  trait !![A] extends Serializable
```
milessabin: Don't we have this already?

core/src/main/scala/shapeless/package.scala
yilinwei: I don't think there was a particular reason, when writing it within our own code base it just fell out that we wrote it as `ev.a.isDefined`. I also originally didn't see a reason to bias the disjunction to the left or the right in the case of the equality. I can change it to `Either[A, B]`.

core/src/main/scala/shapeless/package.scala
yilinwei: I think that in the case that `A` and `B` exists, you'd get amibiguous implicits so I introduced the implicits in a separate scope. I will check that though.

core/src/main/scala/shapeless/package.scala
yilinwei: For consistency with the rest, but I don't feel particularly strongly either way.

core/src/main/scala/shapeless/package.scala
milessabin: You might have to use a super trait to prioritise the instances.

core/src/main/scala/shapeless/package.scala
yilinwei: `^^` is exclusive OR, so `A` has to exist and `B` has to exist, hence the ambiguous type not in the same scope as the other implicits. `||` is OR which means that `A` and `B` can exist.

core/src/main/scala/shapeless/package.scala
milessabin: I guess I'm still looking for a motivating use for this ... can you give an example that's a bit more realistic than the tests?

milessabin: FWIW, the `∃` and `∀` above really don't have a justification and historical baggage ... they won't survive to shapeless 3.0.

core/src/main/scala/shapeless/package.scala
milessabin: That was a comment on `!!`.

core/src/main/scala/shapeless/package.scala
yilinwei: The original motivation comes from line 252 of this [here](https://github.com/to-ithaca/iliad/blob/master/core/src/main/scala/iliad/algebra/matrix.scala)

I essentially need evidence that either `M <: Nat`, `N <: Nat` or both are greater than a certain number as I don't want to be able to call the method unless that is the case. 

I'm not sure these belong in shapeless which is why this is a speculative PR. I'm entirely happy with leaving it in our own codebase.

core/src/main/scala/shapeless/package.scala
yilinwei: I think this works because the package object is counted as the local scope and is a separate scope to the object so it stops searching after the first implicit in the `&&` case. I can check in a separate project though with a local build to make sure though.

core/src/main/scala/shapeless/package.scala
yilinwei: Eep, in which case I apologize. What's it called?

milessabin: See the various operators just above in the same file.

core/src/main/scala/shapeless/package.scala
milessabin: Interesting ... I think I would be tempted to define a `Pad` type class and have separate instances for the three cases ... you should be able to eliminate the pattern match on the options.

core/src/main/scala/shapeless/package.scala
milessabin: Yes, that will work, but IMO it's better practice for instances to all be in the companion object and prioritised via stacked traits.

core/src/main/scala/shapeless/package.scala
milessabin: It's entirely possible that I'm missing the point, in which case a non-toy example would help a lot :-)

yilinwei:
Closing because I've changed the original code and the fact that the usages in which it helps is fairly niche.

PR #632, opened 2016-08-28
Added type class to support folding a Coproduct by handling each type in the order given by the Coproduct type signature.

harveywi:
For instance, a `Coproduct` `c` of type `Int :+: String :+: Boolean :+: CNil`
could be folded into a `Double` as follows:

```
val result = c.foldCases[Double]
               .atCase(i => math.sqrt(i))
               .atCase(s => s.length.toDouble)
               .atCase(b => if (b) 100.0 else -1.0)
```

This provides some benefits over existing methods for folding over
`Coproduct`s.  Unlike the `Folder` type class, this one does not require a
polymorphic function with a stable identifier, so the syntax is somewhat
lightweight and better suited to situations where the folding function is not
reused (e.g., parser combinator libraries).

Additionally, unlike directly folding over a `Coproduct` with pattern matching
over `Inl` and `Inr` injectors, this type class guarantees that the resulting
fold is exhaustive.  It is also possible to partially fold a `Coproduct` (as
long as cases are handled in the order specified by the `Coproduct` type
signature), which makes it possible to incrementally fold a `Coproduct`.

### DIFF
--- a/core/src/main/scala/shapeless/ops/coproduct.scala
+++ b/core/src/main/scala/shapeless/ops/coproduct.scala
@@ -398,6 +398,75 @@ object coproduct {
   }
 
   /**
+   * Type class to support creating a lightweight syntax for folding a `Coproduct`.
+   * Unlike the `Folder` type class, this one does not require a polymorphic
+   * function with a stable identifier.  Additionally, unlike directly folding over 
+   * a `Coproduct` with pattern matching over `Inl` and `Inr` injectors, this type 
+   * class guarantees that the resulting fold is exhaustive.  It is also possible to
+   * partially fold a `Coproduct` (as long as cases are handled in the order specified
+   * by the `Coproduct` type signature), which makes it possible to incrementally fold
+   * a `Coproduct`.
+   * 
+   * @author William Harvey
+   */
+  trait FoldCases[C <: Coproduct, R, D <: Coproduct] extends Serializable {
+    type F <: HList
+    type Out
+
+    def apply(c: C, fs: F): Out
+  }
+
+  type FoldCasesInit[C <: Coproduct, R] = FoldCases[C, R, C] { type F = HNil }
+
+  object FoldCases {
+    trait CaseBuilder[A, B, C] {
+      def atCase(f: A => B): C
+    }
+
+    private[this] object FoldHelper extends Poly1 {
+      implicit def default[T, A]: Case.Aux[(T, T => A), A] = at[(T, T => A)] {
+        case (t, f) => f(t)
+      }
+    }
+
+    def apply[C <: Coproduct, F <: HList, R](
+      implicit foldCases: FoldCases[C, R, C]): Aux[C, R, C, foldCases.F, foldCases.Out] = foldCases
+
+    type Aux[C <: Coproduct, R, D <: Coproduct, F0 <: HList, Out0] = FoldCases[C, R, D] {
+      type F = F0
+      type Out = Out0
+    }
+
+    implicit def baseFoldCases[C <: Coproduct, R, DH, FIn <: HList, FOut <: HList, Z <: Coproduct](
+      implicit prepend: shapeless.ops.hlist.Prepend.Aux[FIn, (DH => R) :: HNil, FOut],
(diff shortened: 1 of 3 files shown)


harveywi:
Potential use case [here](http://stackoverflow.com/questions/34107849/pattern-matching-with-shapeless-coproduct), but I can provide more examples if desired.

milessabin:
I think I like the general idea here. A few things which I'd like to see explored before merging ...

+ The equivalent for `HList` should be very similar and probably has the same justifications.
+ We could also do something similar for `map` (for both `Coproducts` and `HLists`).
+ It would be even more compelling if the result type could be inferred.

It's pretty clearly an straight alternative to a `Poly` and I think I'd like to see that investigated a bit more.

notxcain:
:+1: on this one. I have similar implementation in one of my projects.

notxcain:
My use case is mapping of merged streams.

milessabin:
Can you say why you can't use a `Poly`?

milessabin:
A small self contained example which shows why a `Poly` doesn't do the job would be very helpful.

notxcain:
@milessabin of course I can. However `Poly` feels verbose in this case. So it's basically a sugar.

milessabin:
@notxcain I'm reluctant to add a completely new fold mechanism just for that ... I'd be happy at looking at more succinct ways of defining `Polys` though.

milessabin:
Just to clarify that ... I think it would be possible to define sugar for constructing `Polys` suitable for an `HList` or `Coproduct` fold ... I think that would be preferable to an alternative fold mechanism.

notxcain:
@milessabin I agree, sugar for `Poly`s would be great.

milessabin:
This PR seems to have ground to a halt. I apologise if this is down to my lack of attention. Please update if there's been any developments.

harveywi:
@milessabin No, there have not been any developments.  :(  Argh, I wish I had more time to fiddle with it, but I've had other work to do.

Your recommended path of exploration [here](https://github.com/milessabin/shapeless/pull/632#issuecomment-243387348) seems rather promising - hopefully someone else can take some interest in carrying the torch.

Krever:
I'm willing to push this forward. Recently I came with workflow encoding using shapeless tagged types and coproduct. As a result I quite deep nesting of folding functions wthat looked a bit like following:
```
classifyEvent(event)
.foldCases[Result]
.atCase(created => 
  validate(created)
    .foldCases[Result]
    .atCase(valid => 
      detectDuplicates(valid)
        .foldCases[Result]
        .atCase(...)
        .atCase(...)
    ).atCase(invalid => handleInvalid(invalid))
).atCase(approved => handleApproved(approved))
.atCase(...)
```
Pattern matching failed me. I cant recall the exact error but it was something simillar to "error during expansion of this match expression (this is a scalac bug)"
Using `Poly`s would be extremely cumbersome and verbose.
In case it needs further justification I'm planning to write a blog post about it so I cna link it then.

Before I start working on it I would like to know:

- is this something you're interested in merging @milessabin ?
- what should be done? Should I just replicate the mechanism for HList? And extend for maping?

I think having a sugar for Polys may be nice but the limitation of them being a stable identifier (object) is very cumbersome. Can we overcome this somehow?

milessabin:
Yes I would be interested in merging this. Yup, the mechanism and syntax should be the same for HLists, and handling `map` as well would be great. Bear in mind what I said previously about result type inference.

joroKr21:
Poly functions have concise builder syntax: #676
It still has to be imported though:
```scala
   val myPoly = Poly2
      .at[Int, Int]((acc, x) => acc + x)
      .at[Int, String]((acc, s) => acc + s.length)
      .at[Int, Boolean]((acc, b) => acc + (if(b) 1 else 0))
      .build

    import myPoly._
```

Does this satisfy the need for conciseness?

joroKr21:
I'm going to close this because it doesn't address the root cause of the problem, which is the verbosity of `Poly` functions. Poly function builders somewhat alleviate this issue at the cost of an extra import. Feel free to reopen If you come up with a better way to concisely encode polymorphic functions.

PR #635, opened 2016-09-02
Implementation of LiftAll for Coproduct

AlecZorab:
Straightforward translation of `shapeless.ops.hlist.LiftAll`

### DIFF
--- a/core/src/main/scala/shapeless/ops/coproduct.scala
+++ b/core/src/main/scala/shapeless/ops/coproduct.scala
@@ -1268,4 +1268,30 @@ object coproduct {
         def apply(): Out = wh.value :: rt()
       }
   }
+
+  sealed trait LiftAll[F[_], In <: Coproduct] {
+    type Out <: HList
+    def instances: Out
+  }
+
+  object LiftAll {
+    type Aux[F[_], In0 <: Coproduct, Out0 <: HList] = LiftAll[F, In0] {type Out = Out0}
+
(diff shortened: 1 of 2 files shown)


AlecZorab:
Now with extra added tests

AlecZorab:
I'm not sure exactly why codecov is complaining about this. If someone can point me at what I've done wrong, I'll fix it.

milessabin:
I think this is actually an issue with the scoverage SBT plugin. For the time being I'm ignoring the complaints.

milessabin:

Atry:
Is it possible to make `LiftAll` a `DepFn0`?

core/src/main/scala/shapeless/ops/coproduct.scala
```diff
@@ -1268,4 +1268,30 @@ object coproduct {
         def apply(): Out = wh.value :: rt()
       }
   }
+
+  sealed trait LiftAll[F[_], In <: Coproduct] {
```
Atry: Why not extends `DepFn0`?

milessabin: Yes, I think that would be preferable. PR?

AlecZorab: The main reason for not extending `DepFn0` was that I didn't know it existed. Probably want the sam to be treatment on `HList`

PR #640, opened 2016-09-20
Fix lens example

3tty0n:
Fix typo, and modified some lenses in accordance with the wiki (https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#boilerplate-free-lenses-for-arbitrary-case-classes)

### DIFF
--- a/examples/src/main/scala/shapeless/examples/lenses.scala
+++ b/examples/src/main/scala/shapeless/examples/lenses.scala
@@ -21,7 +21,7 @@ package shapeless.examples
  * 
  * @author Miles Sabin
  */
-object LenseExamples extends App {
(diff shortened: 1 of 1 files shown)


milessabin:
Thanks!

The spelling correction ("Lense" -> "Lens") is good ... mea culpa.

The examples are correct as they are though, and it would be better to update the wiki to match rather than the other way around.

3tty0n:
Thank you for replying for my PR. For now, I will fix only typo.

3tty0n:
I have fixed typo ("Lense" -> "Lens").

milessabin:
Many thanks ... merging :-)

PR #646, opened 2016-10-17
Added orElse to Poly

denisrosca:
Allow composition of two polymorphic functions by specifying
a fallback to be applied where the first function is not
defined.

### DIFF
--- a/core/src/main/scala/shapeless/poly.scala
+++ b/core/src/main/scala/shapeless/poly.scala
@@ -81,6 +81,28 @@ object PolyDefns extends Cases {
   }
 
   /**
+    * Represents the composition of two polymorphic function values by falling back
+    * to the second function where the first is not defined.
+    *
+    * @author Denis Rosca
+    */
+  class OrElse[F,G](f: F, g: G) extends Poly
+
+  object OrElse {
+    implicit def caseF[C, F <: Poly, G <: Poly, T, U]
+      (implicit unpack: Unpack2[C, OrElse, F, G], cF : Case1.Aux[F, T, U]) = new Case[C, T :: HNil] {
+      type Result = U
+      val value = (t : T :: HNil) => cF(t)
+    }
+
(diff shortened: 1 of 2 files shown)


wheaties:
Does that now handle the case of `foo orElse bar orElse baz` in a single line? 'cause back in 2.11.7 and prior you'd lose context of one of those polys.

denisrosca:
@wheaties No, sadly it does not. This is what I tried:
```
  @Test
  def testOrElse {
    object intPlusOne extends Poly1 {
      implicit def caseInt = at\[Int\](_ + 1)
    }
    object stringPlusOne extends Poly1 {
      implicit def caseString = at\[String\](_ + 1)
    }
    object doublePlusOne extends Poly1 {
      implicit def caseDouble = at\[Double\](_ + 1)
    }

    // This don't work:
    val orElse = intPlusOne.orElse(stringPlusOne).orElse(doublePlusOne)
    //val orElse = intPlusOne orElse stringPlusOne orElse doublePlusOne
    
    //This one works
    //val partialFp = intPlusOne orElse stringPlusOne
    //val fp = partialFp orElse doublePlusOne

    val two = orElse(1)
    typed[Int](two)
    assertEquals(2, two)

    val one1 = orElse("one")
    typed[String](one1)
    assertEquals("one1", one1)

    val tmp = orElse(3.45)
    typed[Double](tmp)
    assertEquals(4.45, tmp, 0.0001)
  }
```

The test fails with the following error: 
```
 could not find implicit value for parameter cse: shapeless.poly.Case[orElse.type,shapeless.::[Int,shapeless.HNil]]
```

denisrosca:
@wheaties Looking at the compose example, it does seem to have the same problem:
```
object intToString extends Poly1 {
  implicit def caseInt = at[Int](_.toString)
}
object stringToInt extends Poly1 {
  implicit def caseString = at[String](_.toInt)
}

val cmps = intToString compose stringToInt compose intToString
println(cmps(1))
```

Fails with a compilation error similar to the one above.

PR #647, opened 2016-10-20
More idiomatic <+= replacement

travisbrown:
This is just a minor matter of syntax, but I know I'm not the only person who's cargo-culted their whole boilerplate setup from here, so we might as well be as idiomatic as possible.

(Thanks to @eed3si9n for pointing out that I could do without the `Def.task` in circe.)


milessabin:

PR #650, opened 2016-10-31
Fix type loss with `Comapped` and a few other typeclasses

jeremyrsmith:
Similarly to https://github.com/milessabin/shapeless/pull/621, the implicit recursive case of `Comapped` lacks an explicit type argument for its tail's `Out`. This hurts type inference in some cases. This PR adds the type argument in order to fix this problem.

Although `Comapped` was my immediate problem, I noticed several other typeclasses that appear to have the same problem for which this relatively mechanical fix can be trivially applied. I included a few more of them in this PR, but there are still several others (in general, wherever a recursive implicit case refers to a type member of its implicit argument, type inference can be improved by eliminating that reference and lifting it to a type argument to the implicit derivation). I'm happy to make an effort to apply this change more universally, or users can just continue to make PRs as their need arises. Let me know what you think @milessabin.

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -94,8 +94,10 @@ object hlist {
 
     implicit def hnilComapped[F[_]]: Aux[HNil, F, HNil] = new Comapped[HNil, F] { type Out = HNil }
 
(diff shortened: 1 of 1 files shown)


jeremyrsmith:
Note: this fix does not seem to work trivially for typeclasses involving `Coproduct`. I'm not sure if this is due to `CNil` not existing, but when I tried to apply this for `ToSum` specifically it caused some tests to fail.

milessabin:
I half expected this to fail MiMa :-)

PR #657, opened 2016-11-14
Rough cuts of Interface and LabelledInterface

clhodapp:
This is the PR I promised. It still needs a good deal of polish, but it should be sufficient to completely clarify what I was talking about (at least for this phase 😉). The core idea is to allow two-way conversions between interfaces (purely abstract traits) and HLists of HList-taking functions. This is the sort of thing that should have a tremendous number of use-cases, but the three I've written tests for are:
  * Autogenerating RPC clients/servers a la Autowire: https://github.com/lihaoyi/autowire
  * Autogenerating instances of interfaces (potentially useful in Scalacheck)
  * Defining AOP-like transformations without runtime reflection or client-project-side macros

Open Questions:
  * Should these be their own typeclasses or should the macros just produce instances of `Generic` and `LabelledGeneric`?
  * Currently, the ending of e.g. a method with shape `(Int)(Int) => Int` (two param lists) is `(Int :: HNil) :: (Int :: HNil) :: HNil => Int`, with there being typeclass instances for conversion to a flattened-param-list version (`Int :: Int :: HNil => Int`). The question is whether we care about keeping the param lists separate or whether the initial encoding should just be flat in the first place

### DIFF
--- a/core/src/main/scala/shapeless/interface.scala
+++ b/core/src/main/scala/shapeless/interface.scala
@@ -1,0 +1,412 @@
+/*
+ * Copyright (c) 2016 Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shapeless
+
+import shapeless.ops.adjoin._
+
+import scala.language.experimental.macros
+
+import scala.reflect.macros.whitebox
+
+trait NestedInterface[T] extends Serializable {
+  type Repr
+  def to(t: T): Repr
+  def from(r: Repr): T
+}
+object NestedInterface {
+  type Aux[T, Repr0] = NestedInterface[T] { type Repr = Repr0 }
+  def apply[T](implicit iface: NestedInterface[T]): Aux[T, iface.Repr] = iface
+  implicit def materialize[T, Repr]: NestedInterface.Aux[T, Repr] =
+    macro InterfaceMacros.materializeUnlabelledNestedInterface[T, Repr]
+}
+
+trait FlatInterface[T] extends Serializable {
+  type Repr
+  def to(t: T): Repr
+  def from(r: Repr): T
+}
+object FlatInterface {
+  type Aux[T, Repr0] = FlatInterface[T] { type Repr = Repr0 }
+  def apply[T](implicit iface: FlatInterface[T]): Aux[T, iface.Repr] = iface
+  implicit def materialize[T, Repr]: FlatInterface.Aux[T, Repr] =
+    macro InterfaceMacros.materializeUnlabelledFlatInterface[T, Repr]
+}
+
+@macrocompat.bundle
+class InterfaceMacros(val c: whitebox.Context) extends SingletonTypeUtils {
+
+  import c.universe._
+  import compat._
+
+  case class InterfaceInfo(members: List[MemberInfo])
+  case class MemberInfo(
+    name: TermName,
+    paramLists: List[List[ParamInfo]],
+    returnType: Tree
+  )
+  case class ParamInfo(
+    name: TermName,
+    paramType: Tree
+  )
+
+  sealed trait ExtractionError
+  case class NotTrait(tpe: Type) extends ExtractionError
+  case class ConcreteMember(symbol: Symbol) extends ExtractionError
+  case class GenericMember(symbol: Symbol) extends ExtractionError
+  case class ByNameParameter(param: Symbol) extends ExtractionError
+
+  def traverse[T](
+    items: List[Either[ExtractionError, T]]
+  ): Either[ExtractionError, List[T]] = {
+    items.foldRight[Either[ExtractionError, List[T]]](Right(Nil)) {
+      case (Right(added), Right(current)) => Right(added :: current)
+      case (Right(ignored), Left(existingErr)) => Left(existingErr)
+      case (Left(newErr), dropped) => Left(newErr)
+    }
+  }
+
+  val defaultBaseTypeSymbols = {
+    Set[Symbol](
+      symbolOf[Any],
+      symbolOf[AnyRef],
+      symbolOf[AnyVal],
+      symbolOf[Object]
+    )
+  }
+
+  implicit class BiasEither[T](e: Either[ExtractionError, T]) {
+    def map[U](
+      f: T => U
+    ): Either[ExtractionError, U] = {
+      e.right.map(f)
+    }
+    def flatMap[U](
+      f: T => Either[ExtractionError, U]
+    ): Either[ExtractionError, U] = {
+      e.right.flatMap(f)
+    }
+  }
+
+  def extractInterfaceInfo(
+    tpe: Type
+  ): Either[ExtractionError, InterfaceInfo] = {
+    for {
+      memberSymbols <- {
+        if (tpe.typeSymbol.isClass && tpe.typeSymbol.asClass.isTrait) Right {
+          tpe
+            .members
+            .flatMap { member =>
+              if (member.isTerm) {
+                member.asTerm.alternatives.toList
+              } else List(member)
+            }.filterNot(member => defaultBaseTypeSymbols contains member.owner)
+            .toList
+        } else Left(NotTrait(tpe))
+      }
+      members <- traverse {
+        memberSymbols.map(extractMemberInfo(tpe))
+      }
+    } yield InterfaceInfo(members = members)
+  }
+
+  def extractMemberInfo(
+    containingType: Type
+  )(
+    memberSymbol: Symbol
+  ): Either[ExtractionError, MemberInfo] = {
+    for {
+      signature <- {
+          val signature = memberSymbol.typeSignatureIn(containingType)
+          if (!memberSymbol.isAbstract) Left(ConcreteMember(memberSymbol))
+          else if (signature.typeParams.nonEmpty) Left(GenericMember(memberSymbol))
+          else Right(signature)
+      }
+      name <- Right(memberSymbol.name.toTermName)
+      returnType <- Right(TypeTree(signature.finalResultType))
+      paramLists <- traverse {
+        signature.paramLists.map { paramList =>
+          traverse {
+            paramList.map(extractParamInfo(signature))
+          }
+        }
+      }
+    } yield MemberInfo(name = name, returnType = returnType, paramLists = paramLists)
+  }
+
+  def extractParamInfo(
+    methodSignature: Type
+  )(
+    paramSymbol: Symbol
+  ): Either[ExtractionError, ParamInfo] = {
+    for {
+      paramType <- {
+        if (paramSymbol.asTerm.isByNameParam) Left(ByNameParameter(paramSymbol))
+        else Right(TypeTree(paramSymbol.typeSignatureIn(methodSignature)))
+      }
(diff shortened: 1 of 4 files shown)


clhodapp:
Note that all these tests are impure but it would be quite possible to do all three things being shown in a "pure" way, it would just require more supporting machinery.

clhodapp:
It's always 2.10....

clhodapp:
FWIW, I intend to fix this on the present build error on the `macro-compat` end.

milessabin:
This looks like great stuff, but I'm stilling having trouble disentangling the general components from the particular application. It _looks_ like you're building infrastructure to support proxying `Impl` via `Iface` ... is that the key idea here? If it is, do you think you could simplify things to illustrate that more clearly?

clhodapp:
Sorry for the delay. Work stuff got the better of me last week and I've been sick for most of this. Anyway, I'll try to explain more fully what's going on.

The key change here is the introduction of two new typeclasses: `Interface` and `LabelledInterface`. An instance of `Interface[T]` provides an isomorphism between an `HList` of functions whose shapes match the abstract value members of `T` and a concrete instance of `T`:

```
scala> import shapeless._

scala> trait Ops[T] { def add(a: T, b: T): T }
defined trait Ops

scala> object IntOps extends Ops[Int] { def add(a: Int, b: Int): Int = a + b }
defined object IntOps

scala> val iface = Interface[Ops[Int]]
iface: shapeless.Interface[Ops[Int]]{type Repr = shapeless.::[shapeless.::[shapeless.::[Int,shapeless.::[Int,shapeless.HNil]],shapeless.HNil] => Int,shapeless.HNil]} = $anon$1@5ab486ae

scala> val repr = iface.to(IntOps)
repr: iface.Repr = $anon$1$$Lambda$2549/1953160835@176ef57f :: HNil

scala> val f = repr(0)
f: shapeless.::[shapeless.::[Int,shapeless.::[Int,shapeless.HNil]],shapeless.HNil] => Int = $anon$1$$Lambda$2549/1953160835@176ef57f

scala> f(HList(1,2) :: HNil)
res0: Int = 3

scala> val impl = iface.from(repr)
impl: Ops[Int] = $anon$1$$anon$2@7425f03e

scala> impl.add(1,2)
res1: Int = 3
```

As a careful reading of the above will disocver (and as with `Generic`), `Interface` instances have a type member called `Repr`, which holds the `HList`-ified type that serves as the generic representation of the target type `T`. Of course, there exists a type `Aux` on the `Interface` companion object, in order to allow you to capture `Repr` types into type argument lists for typelevel programming.

Critical to understand is that it's generally only useful to materialize an `Interface` instance for a type with at least one abstract member, but for a lot of use cases, you'll you'll end up _using_ that instance with a fully-concret object. In these cases, the form that the concrete object gets turned into is dictated by the type that the `Interface` instance is for, rather than the type of the concrete object (which will be some subtype of that).

The other typeclass, `LabelledInterface`, is to `Interface` as `LabelledGeneric` is to generic: it's simply a version which captures the names of the functions and function-parameters from the abstract members.

The main benefit to all of this is that it lets you build up, tear down, and transform behaviored objects using the power of inductive typelevel computation: If you can define an inductive way to materialize the `Repr` from a `Interface` instance, you can compile-time generate concrete implementations of an interface implementation (e.g. for mocking or property-based testing). If you can inductively transform that `Repr` while retaining the same function types, you can do a limited sort of aspect-oriented programming (e.g. materialize an "tracing instrumenter" that adds trace logging to type). If you can inductively collapse the `Repr` into a value of a different type, you have defined a converter from instances of an interface to that new type (e.g. collapsing interface implementations into RPC message handling functions).

All three of the just-described examples are presented concretely as tests. Here's a rough explanation for the layout of the tests:

There are three "framework" packages, which contain skeleton versions of code that would appear in a library that wanted to leverage `Interface` or `GenericInterface`. The packages are:
  * `package rpcFramework`: Skeleton for an [autowire](https://github.com/lihaoyi/autowire)-like RPC library
  * `package tracingFramework`: Skeleton for a library to instrument interface instances with tracing information (keeps a log of calls in a `ListBuffer`)
  * `package generationFramework`: Skeleton for a library to allow autogenerating interface instances (a la Scalacheck, but **much** more basic)

There is also an `rpcService` package, which contains the interface we will be operating on and a concrete instance of it.

Finally, `class InterfaceTests` contains the actual tests.

Note: This code is definitely NOT ready to be merged. At the very least, I want to work to remove some of the low-hanging limitations on materializing `Interface` instances and to make a small patch here and in macro-compat to get things working with 2.10 (unless you are planning on dropping 2.10 support from shapeless in the immediate future, @milessabin, which I kind of doubt).

clhodapp:
One thing that is clear once you realize it but not before (:wink:) is that your inductive framework can be restricted to only operating in cases where all of the functions return e.g. a `Task`, which allows you do do things like the tracing purely.

larsrh:
I just came up with this idea independently and then @alexarchambault pointed me here. I'd really love to see this. My use case would indeed be RPC-style; e.g. we could build a Scala version of JNA on top of this.

notxcain:
Wow! It's so awesome when the idea you've got is already a PR in shapeless. I have a library Liberator. The cornerstone of it is the typeclass `FreeAlgebra` https://github.com/aecor/liberator/blob/master/macros/src/main/scala/io/aecor/liberator/FreeAlgebra.scala, it represents the ability to convert `F[A]`, where F: (* -> *) -> * to `FOp ~> A`. Instanes are macro generated. Besides its obvious application to free monads, it has yet unexplored application to finally tagless encoding (https://gist.github.com/notxcain/7dc83acf08b94dcaaa436a12dbcf5df8).

notxcain:
And the idea with typesafe RPC is awesome, and I've been exploring it for a while.

core/src/main/scala/shapeless/interface.scala
```diff
@@ -0,0 +1,424 @@
...
+          val target = TermName(c.freshName("target"))
+            val body = interfaceInfo.members.map {
+              case MemberInfo(name, paramLists, returnType) =>
+                val param = Ident(TermName(c.freshName("param")))
+                val arguments = {
+                  paramLists
+                    .zipWithIndex
+                    .map { case (paramList, listIndex) =>
+                      paramList.zipWithIndex.map { case (name, paramIndex) =>
+                        q"$param.apply($listIndex).apply($paramIndex)" }
```
larsrh: There is a hygiene problem hidden here. `param` is, as far as I understand, an `HList` of `HList`s. Is that correct? In that case, `apply` would require some implicits to resolve. Here's a safer version:

```diff
@@ -258,6 +241,10 @@
   def valueHCons(a: Tree, b: Tree): Tree = q"_root_.shapeless.::($a, $b)"
   val valueHNil: Tree = q"_root_.shapeless.HNil"
 
+  def selectByIndex(l: Tree, n: Int): Tree =
+    if (n == 0) q"$l.head"
+    else selectByIndex(q"$l.tail", n - 1)
+
   def materializeInterface[T: WeakTypeTag, R: WeakTypeTag](
     typeclassName: String,
     paramRepr: ParamInfo => Tree,
@@ -331,7 +318,8 @@
                     .zipWithIndex
                     .map { case (paramList, listIndex) =>
                       paramList.zipWithIndex.map { case (name, paramIndex) =>
-                        q"$param.apply($listIndex).apply($paramIndex)" }
+                        selectByIndex(selectByIndex(q"$param", listIndex), paramIndex)
+                      }
                     }
                 }
                 val paramType = {
@@ -361,7 +349,7 @@
                         q"${param.name}.asInstanceOf[$paramReprTpe]"
                       }.foldRight(valueHNil)(valueHCons)
                     }.foldRight(valueHNil)(valueHCons)
-                q"$target.apply($idx).apply($arg)"
+                q"${selectByIndex(q"$target", idx)}.apply($arg)"
               }
               val parameters = paramLists.map { paramList =>
                 paramList.map { param =>
```

core/src/main/scala/shapeless/interface.scala
```diff
@@ -0,0 +1,424 @@
...
+}
+
+object Interface {
+  type Aux[T, Repr0] = Interface[T] { type Repr = Repr0 }
+  def apply[T](implicit iface: Interface[T]): Aux[T, iface.Repr] = iface
+  implicit def materialize[T, R]: Interface.Aux[T, R] =
+    macro InterfaceMacros.materializeUnlabelledInterface[T, R]
+}
+
+trait FlattenHList[Nested <: HList] {
```
larsrh: This seems to be only needed for the tests; it should probably go somewhere else than this file.

clhodapp: This is for end-user code, which the tests are mock examples of.

This typeclass is what I'm referring to in the initial PR description with:
> Currently, the ending of e.g. a method with shape (Int)(Int) => Int (two param lists) is (Int :: HNil) :: (Int :: HNil) :: HNil => Int, with there being typeclass instances for conversion to a flattened-param-list version (Int :: Int :: HNil => Int). The question is whether we care about keeping the param lists separate or whether the initial encoding should just be flat in the first place

core/src/main/scala/shapeless/interface.scala
clhodapp: This is a good point.

core/src/main/scala/shapeless/interface.scala
larsrh: Understood, but isn't there something like that already in shapeless? `Adjoin` could probably do the job, or a specialization of `FlatMapper` to `identity`.

milessabin:
Where are we at with this?

clhodapp:
I am trying to find a window of time and energy for second round of work on it. Hopefully this weekend I can get 2.10 working. I think I'm close.

milessabin:
That would be completely awesome :-)

kailuowang:
This is really really cool! Does it also support traits that take type parameters (or even higher kinded type parameters like `MyTrait[F[_]]`)?

clhodapp:
@kailuowang It isn't kind-polymorphic but there is e.g. nothing preventing you from using it on `trait Foo[T[_]]`... You'd just have to fill in `T` at materialization time (e.g. `FlatInterface[Foo[List]]`).

durban:
What is the current status of this PR? Can I do anything to help it move along?

larsrh:
@durban As far as I can tell, the build failure stems from a non-existing scala-native build of macro-compat. So maybe [start there](https://github.com/milessabin/macro-compat) and lobby @milessabin into making a release after that :smile:

durban:
Okay, now I really don't understand: shapeless (master) seems to support scala-native (https://github.com/milessabin/shapeless/blob/master/build.sbt#L120), while macro-compat doesn't. How is this even possible?

durban:
@clhodapp I've sent a pull request to your branch: https://github.com/clhodapp/shapeless/pull/1

It rebases the changes to current shapeless master, upgrades the macro-compat dependency to its current master and hopefully fixes the scala native build. Could you add it to this PR, so that we can see if it fixes the travis build?

(If yes, the next step could be a release of the current macro-compat, since for now there is still a source dependency on it.)

durban:
@clhodapp Have you had a chance to look at the Travis problem? (Unfortunately I have no idea how this stuff works.)

clhodapp:
@durban Oh yes! It was a short-lived issue that fixed itself with time. See https://travis-ci.org/milessabin/shapeless/builds/295581126. The only problem now is that the code coverage tool seems to have gotten a bit confused.

durban:
Oh, cool, it seems to be all green on Travis.

@milessabin Would you be able to do a release from the current master of macro-compat? (If I understand correctly, it is required for this PR to work.) After that, we could remove the source dependency from this PR, and AFAICT it would be good to go.

milessabin:
I appreciate that this will probably be disappointing, but I've decided against adding any new macros to shapeless for the time being. By the looks of it, this could happily exist as an independent project with a shapeless dependency.

larsrh:
@clhodapp Do you want to give it a shot to pull it out into a separate project? Alternatively, would you be happy with me ripping this out?

clhodapp:
@larsrh I'm currently feeling like I do want to pull it out into its own library. No commitment on timing, though. I have no objection to you ripping out any of this code and using it in any open-source or even commercial project if you so wish, though, so you don't have to wait for me.

PR #658, opened 2016-12-04
Use MemberScopeApi#sorted where necessary

### LINKED ISSUE #576: Use MemberScopeApi.sorted where necessary
As discussed in #575, members in a scope (as returned by, e.g., `Type.decls`) are not guaranteed to be in declaration order. To guarantee this, `MemberScopeApi.sorted` can be used. According to @milessabin there can be other places where this may be necessary. This issue is to track changing them to use `.sorted`.

durban:
I added the .sorted call where it seemed to make sense. Should fix #576.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -353,7 +353,7 @@ trait CaseClassMacros extends ReprTypes {
     val tSym = tpe.typeSymbol
     if(tSym.isClass && isAnonOrRefinement(tSym)) Nil
     else
-      tpe.decls.toList collect {
+      tpe.decls.sorted collect {
         case sym: TermSymbol if isCaseAccessorLike(sym) =>
           (sym.name.toTermName, sym.typeSignatureIn(tpe).finalResultType)
       }


milessabin:
Could you remind me of the discussion? Just looking at these quickly it's not immediately obvious to me that sorting makes any semantic difference.

alexarchambault:
Or add a test to illustrate the difference (if any)?

durban:
I don't think there is a difference in practice, it's just that the items in `Type.decls` are not guaranteed to be in (source) declaration order. In practice, they _seem_ to be; but with the `.sorted` call, they are now _guaranteed_ to be. (See also #576 and #575.)

milessabin:
In the first and last case in your commit there would only be a difference if there was more than one match. Should it be an error if that's the case? In the second case I don't think sort order is relevant?

durban:
Hm, right, the first and last should have at most one match. The last already seem to check this (`productCtorsOf(tpe).size == 1`). The first one doesn't (although I don't think it can happen that there are more). So you're right, these shouldn't necessarily be sorted.

The second case (if I'm not mistaken) seems to collect the fields of a (e.g.) case class for `Generic`. In this case, I think order could matter?

durban:
I've updated my commit, and removed the original first and last modification. Now only the one in `fieldsOf` remains, which I still think makes sense.

milessabin:
My sincere apologies for taking to long to review and merge this :-(

PR #661, opened 2016-12-11
Add IsHCons.cons

dwijnand:
See https://twitter.com/dwijnand/status/807742485342593024 for motivation.

### DIFF
--- a/core/src/main/scala/shapeless/ops/coproduct.scala
+++ b/core/src/main/scala/shapeless/ops/coproduct.scala
@@ -859,6 +859,7 @@ object coproduct {
 
     def head(c: C): Option[H]
     def tail(c: C): Option[T]
+    def cons(e: Either[H, T]): C
   }
 
   object IsCCons {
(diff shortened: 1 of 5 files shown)


milessabin:
Could you add the corresponding method to `IsCCons` as well. I think it belongs in the same PR. You should be able to make the two completely symmetrical :-)

milessabin:
They key insight is that the kind of product is preserved (`HList` and `Coproduct` are both "products" in this sense). Switching to `Tuple2` and `Either` for the signature might also be reasonable.

dwijnand:
Well, that compiles. Is that what you meant? Does it make sense? I've not played around with Coproduct much..

Also excluded them from MiMa, is that right?

core/src/main/scala/shapeless/ops/coproduct.scala
```diff
@@ -859,6 +859,7 @@ object coproduct {
 
     def head(c: C): Option[H]
     def tail(c: C): Option[T]
+    def cons(e: Either[H, T]): H :+: T
```
milessabin: Close ... this should give us a `C` in the same way that `cons` on `IsHCons` gives us an `L`.

dwijnand:
That's pretty harsh, @codecov-io..

dwijnand:
@milessabin How does this look to you now? GTG?

milessabin:
Yeah, that's about the size of it. Bung in some tests to keep codecov happy and I'll merge :-)

dwijnand:
@milessabin green (including from codecov)

milessabin:

Issue #579, opened 2016-04-05
Typeable instance derived for nested classes can be unsafe


durban:
The automatically generated `Typeable` instance for nested classes can be unsafe, if they refer to a type member of the enclosing class. The macro generating the instance correctly rejects uncheckable generic classes, but allows these; unfortunately these classes are in effect also "generic". The problem affects both regular classes and case classes (examples which cause a `ClassCastException`: https://github.com/durban/akka-typed-examples/blob/ca73c1dc1e1ca729ddc355a175ca95ca4d448b75/src/main/scala/com/example/shless/TypeableExample4.scala and https://github.com/durban/akka-typed-examples/blob/ca73c1dc1e1ca729ddc355a175ca95ca4d448b75/src/main/scala/com/example/shless/TypeableExample5.scala).

I discovered this when I worked on #574. I tried to fix this too, but haven't been able to find a way to distinguish cases when the nested class refers to the type member from the ones when is doesn't. (If it doesn't refer to the type member, I _think_ its not unsafe.)

milessabin:
Good spot, and thanks for the analysis.

PR #668, opened 2016-12-20
Try to fix more unsoundness in Typeable (#579)

allisonhb:
I believe that this fixes the nested type problems in #579.

Specifically, prior to this patch, with a type such as:

```
trait T { type Inner }
class K[P] extends T { type Inner = T; case class Foo(t:T) }
```

the default derived `Typeable` for `K[P]#Foo` would see Foo as a simple unparameterized case class, and perform casting with a simple `isInstanceOf` check.

With this patch, after all other checks for special derived `Typeable`s are made, before we fall back to `simpleTypeable`, we check to see if any values have a type that looks different from the inside as from the outside, which indicates that the erased type will not contain information about that type. At that point, the situation is the same as for a parameterized type; we can safely derive a `Typeable` for a product, but not for an arbitrary type.

### DIFF
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -366,11 +366,11 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
   import internal._
   import definitions.NothingClass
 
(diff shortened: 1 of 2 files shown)


milessabin:
This is great stuff! Many thanks ... merging :-)

PR #669, opened 2016-12-22
Speculative PR for deep generics

zainab-ali:
I have the need for a deep generic representation of a nested ADT.  
This representation is composed of HLists of base elements. e.g.
```scala
case class Address(number: Int, street: String)
case class Person(name: String, address: Address)
the[DeepGeneric[Person]] 
//Repr is (String :: HNil) :: ((Int :: HNil) :: (String :: HNil) :: HNil) :: HNil
```
Is this idea, or a similar one, useful in shapeless itself?

**EDIT**
The representation for `DeepGeneric[Person]` should actually be
```scala
String :: (Int :: String :: HNil) :: HNil
```


milessabin:
Sorry it's taken me so long to get to this ... I'm catching up with things after the break.

shapeless has followed the Haskell generics literature (Scrap Your Boilerplate and GHC Generics in particular) in taking a "one layer at a time" approach to generic representations. In these approaches the deep structure is captured by a traversal, which manifests itself as the kind of inductive traversals you see in examples of type class derivation using shapeless. But there are other points in the design space which are very much worth exploring.

I'm a little reluctant to include this in shapeless core just yet, but I would be keen to include it as an example, especially if you're able to extend it to cover things which have a `Coproduct` generic representation.

zainab-ali:
Thanks for the feedback, I also think this is more appropriate as an example.  There's already a `DeepHLister` which is fairly similar.  I'll expand this to include `Coproduct`, and clean it up a bit.

milessabin:
Thank you!

examples/src/main/scala/shapeless/examples/deepGeneric.scala
```diff
@@ -0,0 +1,98 @@
...
+    type Repr = HNil
+    def to(h: HNil): HNil = h
+    def from(h: HNil): HNil = h
+  }
+
+  implicit def hlistDeepGeneric[H, HG, T <: HList, TR <: HList, HR](
+    implicit hg: Generic.Aux[H, HG],
+    hgen: Lazy[DeepGeneric.Aux[HG, HR]],
+    tgen: Lazy[DeepGeneric.Aux[T, TR]]
+  ): DeepGeneric.Aux[H :: T, HR :: TR] = new DeepGeneric[H :: T] {
```
zainab-ali: This fails to compile with `ambiguous implicit values.  both method hlistDeepGeneric and method hlistHeadNotGeneric match expected type`.  I would have thought that `hlistDeepGeneric` would get prioritized.

Perplexingly, if I add the constraint `HR <: HList` then the code compiles.  Since my goal is to include coproducts, I don't want to do this.  

Any ideas on what I've done wrong?

milessabin: Yes, I would expect the same. Have you tried lowering the priority of `hlistDeepGeneric`?

milessabin:
Apologies for not paying closer attention ... is there anything I can do to help move this along?

zainab-ali:
Apologies for the late update - I've made several attempts at this, but am still perplexed.  I'm going to close this PR, as I think that any solution would be too convoluted to make sense as an example.
I'll repoen it if and when I can get the code to work well.

milessabin:
If you push to the PR we could maybe give you some advice?

PR #670, opened 2017-01-02
Fix code samples in scaladoc for Lazy

savulchik:


milessabin:
Could you elaborate a little on this? What have you fixed?

alexarchambault:
The point of the example is precisely to show that `Lazy` is needed even when there are no recursions.

savulchik:
@milessabin @alexarchambault the first existing code sample didn't produce divergence in the REPL as was stated in the comment for `implicitly[TC[CC]]`. I thought something was wrong so I decided that case class CC could be made recursive to produce divergence as expected. Also the third code sample explaining Lazy implementation was inconsistent to me with the previous CC case class definition.

alexarchambault:
The mistake is that the last line in the example should be
```scala
implicitly[TC[ListCC]]
```
rather than
```scala
implicitly[TC[CC]]
```

Compiling the following file fails then
```scala
import shapeless._

case class ListCC(list: List[CC])
case class CC(i: Int, s: String, i0: Int)

trait TC[T]

object TC {
  implicit def intTC: TC[Int] = ???
  implicit def stringTC: TC[String] = ???
  implicit def listTC[T](implicit underlying: TC[T]): TC[List[T]] = ???

  implicit def genericTC[F, G](implicit
    gen: Generic.Aux[F, G],
    underlying: TC[G]
  ): TC[F] = ???

  implicit def hnilTC: TC[HNil] = ???

  implicit def hconsTC[H, T <: HList](implicit
    headTC: TC[H],
    tailTC: TC[T]
  ): TC[H :: T] = ???
}

object Foo {
  implicitly[TC[ListCC]]
}
```
but succeeds with the `Lazy[...]`s.

savulchik:
@alexarchambault I believe we could just leave the recursive case class for demonstration purposes and remove non-recursive case class then?

savulchik:
@alexarchambault sorry, I didn't notice that there is no recursive case class in your example. I tried it in Ammonite Repl 0.8.1 (Scala 2.12.1 Java 1.8.0_102) and Scala 2.11.8 REPL - it compiles without the expected divergence error:
```scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_102).
Type in expressions for evaluation. Or try :help.

scala> :paste
// Entering paste mode (ctrl-D to finish)

import shapeless._

case class ListCC(list: List[CC])
case class CC(i: Int, s: String)

trait TC[T]

object TC {
  implicit def intTC: TC[Int] = ???
  implicit def stringTC: TC[String] = ???
  implicit def listTC[T](implicit underlying: TC[T]): TC[List[T]] = ???

  implicit def genericTC[F, G](implicit
    gen: Generic.Aux[F, G],
    underlying: TC[G]
  ): TC[F] = ???

  implicit def hnilTC: TC[HNil] = ???

  implicit def hconsTC[H, T <: HList](implicit
    headTC: TC[H],
    tailTC: TC[T]
  ): TC[H :: T] = ???
}

object Foo {
  implicitly[TC[ListCC]]
}

import shapeless._
defined class ListCC
defined class CC
defined trait TC
defined object TC
defined object Foo

scala> Foo
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:230)
  at TC$.intTC(<console>:19)
  ... 44 elided
```

alexarchambault:
@savulchik Yeah, but it fails when it's in a standalone file. Copy-pasting in Ammonite isn't equivalent to putting things as is in `.scala` files, as it "wraps" user code in singletons. Lazy allows to make it work no matter what.

savulchik:
@alexarchambault thanks for clarification. I updated the pull request.

milessabin:
My sincere apologies for taking so long to review and merge this :-(

PR #672, opened 2017-01-03
Make shapeless.the respect @implicitNotFound (Fix #671)

### LINKED ISSUE #671: shapeless.the does not respect @implicitNotFound
``` scala
@ Ordering[Set[Int]] 
cmd5.scala:1: No implicit Ordering defined for Set[Int].
val res5 = Ordering[Set[Int]]
                   ^
Compilation Failed
@ type StringOrderingOps = the.`Ordering[Set[Int]]`.Ops 
cmd5.scala:1: Could not infer implicit value for Ordering[Set[Int]]
type StringOrderingOps = the.`Ordering[Set[Int]]`.Ops
                         ^
Compilation Failed
```

There is an `@implicitNotFound` annotation on `Ordering`, thus I expect the error message for ```the.`Ordering[Set[Int]]`.Ops``` is `No implicit Ordering defined for Set[Int].`. However the error message is `Could not infer implicit value for Ordering[Set[Int]]`

Atry:


milessabin:
Great stuff! I think that `slient = false` means that the following `isEmpty` condition will always be false ... could you check whether or not that's the case and if it is remove the conditional.

Atry:
I removed the conditional and left an assert call instead

milessabin:
Thanks! I think the assert should go though ... it's not validating any logic of ours, so doesn't really belong here.

milessabin:
Sorry to keep chiselling away on this ... would you mind adding a test which verifies that `the` yields values with refined type members as described in the scaladoc?

milessabin:
Bump ... I'll merge as soon as the additional test has been added.

Atry:
Thank you.

Atry:
Don't know how to test the error message

milessabin:
Add a test of the form,
```scala
val foo = ...
foo: <type including refinement>
```
This will verify that the inferred type of `foo` is the expected type, including the expected refinement, without the refinement influencing type inference. Alternatively used `assertTypedEquals`.

Atry:
This PR only changes the error message. How to test if the error message is expected?

milessabin:
Whoops! Sorry, I'm confusing PRs :-)

You should be able to use `illTyped` with an expected error message here.

Please also squash the resulting commits.

AndreasKostler:
I am not sure this does what it's expected to do:
```scala
the.`Ordering[Set[Int]]`
error: exception during macro expansion:
scala.reflect.macros.TypecheckException: implicit search has failed. to find out the reason, turn on -Xlog-implicits
```
expected
```scala
error:  No implicit Ordering defined for Set[Int]
```

AndreasKostler:
I don't want to hijack this PR but I do have a working one if interested.

PR #673, opened 2017-01-06
ToInt macro

eugengarkusha:
Currently the Nat numbers materialization has a runtime cost.

Proposed solution :  
1)eliminates run time cost
2)increases compile time performance

### DIFF
--- a/core/src/main/scala/shapeless/ops/nat.scala
+++ b/core/src/main/scala/shapeless/ops/nat.scala
@@ -604,7 +604,7 @@ object nat {
 
     implicit val toInt0: ToInt[_0] = new Inst[_0](0)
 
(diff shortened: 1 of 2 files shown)


eugengarkusha:
codecov complains on 'toIntSucc'method which is kept only for binary compatibility , marked as deprecated and never supposed to be invoked .
Is it considered safe to keep it uncovered(because covering it will look strange)?

milessabin:
I think you should drop the `@deprecated` ... no ever explicitly called this method anyway, so about the only thing it's doing is causing trouble here.

eugengarkusha:
yes, right

milessabin:
My sincere apologies for taking so long to review and merge this.

eugengarkusha:
Thanks!!

PR #675, opened 2017-01-08
Reorder ops.tuple.Replacer type params to match scaladoc and naming conventions

teldosas:
I will add the description as comments to the edited source code.


core/src/main/scala/shapeless/ops/tuples.scala
```diff
@@ -260,17 +260,17 @@ object tuple {
...
+  trait Replacer[T, U, V] extends DepFn2[T, V] with Serializable
 
   object Replacer {
     def apply[T, U, V](implicit replacer: Replacer[T, U, V]): Aux[T, U, V, replacer.Out] = replacer
 
     type Aux[T, U, V, Out0] = Replacer[T, U, V] { type Out = Out0 }
 
     implicit def replaceTuple[T, L1 <: HList, U, V, L2 <: HList]
-      (implicit gen: Generic.Aux[T, L1], replace: hl.Replacer.Aux[L1, V, U, (V, L2)], tp: hl.Tupler[L2]): Aux[T, U, V, (V, tp.Out)] = new Replacer[T, U, V] {
-        type Out = (V, tp.Out)
```
teldosas: and herer by the type of the implicit parameter `replace`

core/src/main/scala/shapeless/syntax/std/tuples.scala
```diff
@@ -153,7 +153,7 @@ final class TupleOps[T](t: T) extends Serializable {
   def replace[U](u: U)(implicit replacer: Replacer[T, U, U]): replacer.Out = replacer(t, u)
   
   class ReplaceTypeAux[U] {
-    def apply[V](v: V)(implicit replacer: Replacer[T, V, U]): replacer.Out = replacer(t, v)
+    def apply[V](v: V)(implicit replacer: Replacer[T, U, V]): replacer.Out = replacer(t, v)
```
teldosas: The code here has been adapted to the wrong order of the type params.

core/src/main/scala/shapeless/syntax/std/tuples.scala
```diff
@@ -174,7 +174,7 @@ final class TupleOps[T](t: T) extends Serializable {
     (implicit replacer: Replacer.Aux[T, U, U, (U, R)]): R = replacer(t, u)._2
   
   class UpdatedTypeAux[U] {
-    def apply[V, R](v: V)(implicit replacer: Replacer.Aux[T, V, U, (U, R)]): R = replacer(t, v)._2
+    def apply[V, R](v: V)(implicit replacer: Replacer.Aux[T, U, V, (V, R)]): R = replacer(t, v)._2
```
teldosas: Same here.

core/src/main/scala/shapeless/ops/tuples.scala
```diff
@@ -260,17 +260,17 @@ object tuple {
    *
    * @author Miles Sabin
    */
-  trait Replacer[T, U, V] extends DepFn2[T, U] with Serializable
```
teldosas: This becomes apparent here as it is defined as a functions that takes a `T` and a `U`,

milessabin:
I think this is good and a perfectly sensible change. However I haven't been able to completely convince myself that this doesn't silently introduce a semantic change (which isn't tripping MiMa). I've asked on the gitter channel for more eyeballs to take a look :-)

OlivierBlanvillain:
I think this changes semantics: `Replacer[(Int, Boolean), Int, Boolean].apply` used to be a function that takes a `Tuple2[Int, Boolean]` and replaces the 2nd element with something else. After this change, `Replacer[(Int, Boolean), Int, Boolean].apply` replaces the 1st element.

Note that `Replacer[(A, B), B, C]` carries no information about the type of the replacing element, because Tuples are full of `Any`s a `Replacer[(A, B), B, C]` is basically a `Replacer[(A, B), B, X]` for any `X`:

```scala
import shapeless.ops.tuple._

val r = Replacer[(String, Double), Double, Boolean]

r.apply(("", 1.1), true) // (1.1,("",true))

r.asInstanceOf[Replacer[(String, Double), Double, Int]].apply(("", 1.1), 0) // (1.1,("",0))
```

Which is why MiMa did not complain...

teldosas:
@milessabin Should this be flagged for 3.0 then?

milessabin:
Yes, I think so. Sorry it couldn't be merged more eagerly :-(

OlivierBlanvillain:
@teldosas Maybe you could update the scaladoc to match the implementation instead of the opposite?

teldosas:
@OlivierBlanvillain I started #680 which does what you said, but I think it would still make sense to merge this one in version 3.0 cause besides the scaladoc there is also an naming convention in this file and in similar files of the other data structures. So I think it would be more concise this way. This is why I didn't do what you said from the beginning.

joroKr21 (approved):

PR #676, opened 2017-01-10
Add elegant syntax for creating polys from functions

teldosas:
Adds syntax for creating polys like this:
```scala
import poly._

val myPoly = newPoly caseOf[Int]( x => x) caseOf[String](_.length) caseOf[Boolean](if(_) 1 else 0)

import myPoly._

(10 :: "hello" :: true :: HNil).map(myPoly) //  10 :: 5 :: 1 :: HNil
```
Names open to debate off course.

It works for scala 2.12.1 and 2.11.8 and 2.10.6 (for 2.10.6 you have to replace the spaces between the methods with the dot (.) operator).

I will add tests if there is interest in merging this.

### DIFF
--- a/core/src/test/scala/shapeless/poly.scala
+++ b/core/src/test/scala/shapeless/poly.scala
@@ -373,7 +373,7 @@ class PolyTests {
 
   @Test
   def testPoly1Builder {
-    val myPoly = Poly.at[Int]( x => x).at[String](_.length).at[Boolean](if(_) 1 else 0).build
+    val myPoly = Poly1.at[Int]( x => x).at[String](_.length).at[Boolean](if(_) 1 else 0).build
     import myPoly._
 
     val r1 = myPoly(10)
@@ -384,5 +384,23 @@ class PolyTests {
 
     val r3 = myPoly(true)
     assertTypedEquals[Int](1, r3)
+  }
+
+  @Test
+  def testPoly2Builder {
+    val myPoly = Poly2.at[Int, Int]((acc, x) => acc + x).
+                       at[Int, String]((acc, s) => acc + s.length).
+                       at[Int, Boolean]((acc, b) => acc + (if(b) 1 else 0)).
+                       build
+    import myPoly._
+
+    val r1 = myPoly(5, 10)
+    assertTypedEquals[Int](15, r1)
+
+    val r2 = myPoly(5, "hello")
+    assertTypedEquals[Int](10, r2)
+
+    val r3 = myPoly(5, true)
+    assertTypedEquals[Int](6, r3)
   }
 }
--- a/core/src/test/scala/shapeless/poly.scala
+++ b/core/src/test/scala/shapeless/poly.scala
@@ -370,4 +370,19 @@ class PolyTests {
     val r6 = dcis(2.0, 'a', 1, "foo")
     assertTypedEquals[String](s"i: 1, s: foo, d: ${2.0}, c: a", r6)
   }
+
+  @Test
+  def testPolyBuilder {
+    val myPoly = newPoly.caseOf[Int]( x => x).caseOf[String](_.length).caseOf[Boolean](if(_) 1 else 0)
+    import myPoly._
+
+    val r1 = myPoly(10)
+    assertTypedEquals[Int](10, r1)
+
+    val r2 = myPoly("hello")
+    assertTypedEquals[Int](5, r2)
+
+    val r3 = myPoly(true)
+    assertTypedEquals[Int](1, r3)
+  }
 }
--- a/core/src/test/scala/shapeless/poly.scala
+++ b/core/src/test/scala/shapeless/poly.scala
@@ -372,8 +372,8 @@ class PolyTests {
   }
 
   @Test
-  def testPolyBuilder {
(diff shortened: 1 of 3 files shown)


teldosas:
I managed to make it work without the macro @milessabin :)

milessabin:
I'm fairly sure that the use of the type projection (#Out) means that you'll have lost precision in the result type in this version.

Could you add some tests which verify that the result types are as expected?

teldosas:
@milessabin I found a way to overcome the loss of percision, I added tests to verify it. Take a look :)

milessabin:
My sincere apologies for taking so long to get back to you on this.

I think this looks pretty good. For consistency, would it be possible to rename `caseOf` to `at`?

teldosas:
Yes I thought of that too. I'll change it

milessabin:
One more thing, rather than `newPoly`, could `at` be a method on the existing `object Poly`?

teldosas:
> One more thing, rather than newPoly, could at be a method on the existing object Poly?
I'll change that too
>I think this looks pretty good. For consistency, would it be possible to rename caseOf to at?
Because my `Poly1Builder` extends `Poly1` there is a conflict with the at method of `Poly1`

milessabin:
Could you combine your `AtAux` with `CaseBuilder` somehow.

BTW, if you want to test before pushing, the best bet is to use the `scratch` subproject.

teldosas:
>Could you combine your AtAux with CaseBuilder somehow.

I'll think about it

teldosas:
@milessabin I got  close to combining `AtAux` with `CaseBuilder` but the solution was very dirty.
What about a `build` method instead?
```scala
import poly._

val myPoly = Poly at[Int]( x => x) at[String](_.length) at[Boolean](if(_) 1 else 0) build

import myPoly._

(10 :: "hello" :: true :: HNil).map(myPoly) //  10 :: 5 :: 1 :: HNil
```

teldosas:
Idea: I guess it would make sense for this `PolyBuilder`  to be defined at least for `Poly2`. Would it make sense to define a `GenPolyNBuilders` in `Boilerplate.scala` ?

milessabin:
If you can generalize this for polys of all arities, then yes, go for it. And in that case rolling it in to get generated boilerplate would make a lot of sense.

milessabin:
LGTM! Could you squash and rebase before I merge?

teldosas:
I generalized it now it should work for Polys of all arity
I added a companion object for each PolyN. Is that ok?
```scala
val myPoly1 = Poly1 at[Int]( x => x) at[String](_.length) at[Boolean](if(_) 1 else 0) build

import myPoly1._

(10 :: "hello" :: true :: HNil).map(myPoly1) //  10 :: 5 :: 1 :: HNil


val myPoly2 = Poly2.at[Int, Int]((acc, x) => acc + x).
                                at[Int, String]((acc, s) => acc + s.length).
                                at[Int, Boolean]((acc, b) => acc + (if(b) 1 else 0)).
                                build
import myPoly2._

(10 :: "hello" :: true :: HNil).foldLeft(0)(myPoly2) //  16


val myPoly3 = Poly3.at....
.
.
.
val myPoly22 = Poly22.at...

**Comment by milessabin** [comment]
This looks great. Could you a test for a `Poly2` as well ... then I'll merge.

**Comment by teldosas** [comment]
Test added :)

File: core/src/test/scala/shapeless/poly.scala [hunk 1]
```diff
@@ -370,4 +370,37 @@ class PolyTests {
...
+    assertTypedEquals[Int](1, r3)
+  }
+
+  @Test
+  def testPoly2Builder {
+    val myPoly = Poly2.at[Int, Int]((acc, x) => acc + x).
+                                at[Int, String]((acc, s) => acc + s.length).
+                                at[Int, Boolean]((acc, b) => acc + (if(b) 1 else 0)).
+                                build
+    import myPoly._
```
milessabin: Whitespace looks odd here ... stray tabs? Could you either line up the `at`'s or shallow indent in two space increments.

core/src/test/scala/shapeless/poly.scala
```diff
@@ -385,4 +385,22 @@ class PolyTests {
...
     assertTypedEquals[Int](1, r3)
   }
+
+  @Test
+  def testPoly2Builder {
+    val myPoly = Poly2.at[Int, Int]((acc, x) => acc + x).
+                       at[Int, String]((acc, s) => acc + s.length).
+                       at[Int, Boolean]((acc, b) => acc + (if(b) 1 else 0)).
+                       build
+    import myPoly._
```
teldosas: Better?

milessabin: Much ... thanks :-)

milessabin:
Looks perfect ... will merge as soon as Travis goes green :-)

milessabin:
Many thanks ... merging! :-)

milessabin:
@teldosas I just realized that even though you showed `map` and `fold` examples above you didn't add tests for those, just for poly application ... would you mind adding some tests which demonstrate that polys defined this way can be used with operations which require implicit cases?

teldosas:
Addressed with #702 :wink:

PR #680, opened 2017-01-20
Update tuple replacer scaladoc to match implementation

teldosas:
Temporary solution for #675


milessabin:

PR #682, opened 2017-01-22
Improve implicit search times for Filter, FilterNot, Union, Intersection

aryairani:
Hi, 

This patch provides a much faster implementation of `Filter` and `FilterNot`, and somewhat faster implementations of `Union` and `Intersection`.  (Without the patch, my compile times have been prohibitively slow, such that I can only use the static checks for very small operations.)

screen shot 2017-01-22 at 11 19 11 pm

(Note the log-lin scale.)

Codecov check fails only because newly-deprecated code (retained only for bincompat) is no longer used by tests.  The non-covered code can be be deleted at the next minor release.


refried:
codecov check fails because newly-deprecated code (retained only for bincompat) is no longer used by tests.  Hopefully this isn't a deal-breaker!

milessabin:
Can you add tests which explicitly invoke the now redundant methods?

refried:
Sure, I will take a look at that.

refried:
@milessabin Do you know of any other examples of that that I can use to model style?  Otherwise I can wing it.

refried:
Hi — I could use guidance on two issues:

1) In `scalacOptions` we have `-deprecation` and `-Xfatal-warnings`; with the two together, I can't include tests for deprecated methods.  I could not deprecate them, but I think they should be deprecated.

2) I noticed that the hlist syntax methods `.filter` and `.filterNot` [directly summon](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/syntax/hlists.scala#L140-L148) a `Partition[L,U]` instead of `Filter[L,U]` / `FilterNot[L,U]`, which seems like a mistake.  I can:

    a. leave them as-is (linked to the wrong typeclass) or
    b. move them to a `HListOpsDeprecated` trait (this builds, but not sure how it affects bc yet; I havent been able to run mima tests locally; don't know why.)

milessabin:
The old methods won't appear explicitly in any normal user code because they're intended to be invoked implicitly. That being so I think it's right to leave them undeprecated.

A couple of thoughts. Can you get the same improvements for `Partition`? In which case it might be better to leave `Filter` and `FilterNot` as they are, which would also solve the `.filter` and `.filterNot` problem.

Also could you benchmark this change against leaving things as they are but compiling with Typelevel Scala 2.12.1 with the `-Yinduction-heuristics` flag enabled?

refried:
1) With respect to `@deprecation`, I'm happy to replace the annotation with a comment; but remember, the issue with deprecation only came up because codecov wants us to continue referencing the unused methods from tests.  You want them to go away eventually, right?

2) This particular improvement won't help `Partition`, because `Partition` by its contract needs to compute both the `Filter` and `FilterNot` results.  The current implementation of `Filter` and `FilterNot` rely on `Partition` to compute both sets, and then discard one of them.  But computing both seems to be way more expensive than just computing the one that's needed.  

We could probably flip things around so that `Partition` relies on the *new* `Filter` and `FilterNot`, instead of the other way around, and end up with a linear slowdown rather exponential.  This would "help" the `.filter` and `.filterNot` issues somewhat, although it still seems screwy that those methods take a different type class than the ones they're named after.  Maybe the methods predate the corresponding type classes, otherwise I dunno how it came to be the way it is.  They should really use their corresponding type classes though.  

4) I will benchmark against `-Yinduction-heuristics` when I am home next week.  If it's a huge help, then one option would be to forget this, and leave Lightbend folks to suffer; what do you think?

milessabin:
On (1) yes, agreed.

On (2), my intuitions are that if your rework of `Filter`/`FilterNot` can improve the compile time from quadratic to linear, then something similar should be possible for `Partition`. I agree that `Partition` will still be more expensive to compile than either of the two individually, but the margin might be small enough that the benefit of having a single implementation rather than three might outweigh it.

On (3) I think it makes sense to do this whether or not `-Yinduction-heurisitics` improves things, but it would be very useful to see if it helps directly in this case.

refried:
So on (1), would you prefer to (1a) not include the deprecated methods in tests, and ignore the codecov result, (1b) allow deprecated methods in tests by removing `-Xfatal-warnings` in tests, or (1c) not use the `@deprecation` annotation, but just leave a comment that the methods are deprecated?

On (2), I can get it to two implementations, rather than three, but not down to one.  `Partition` is inferring two `HList`s simultaneously; the rework for `Filter`/`FilterNot` was merely to cut them down to each only infer one `HList`.  I can't apply the same rework to `Partition` because it needs to produce both.  However, I can invert the dependence and have `Partition` use `Filter`/`FilterNot` to infer the lists sequentially instead of simultaneously.  Edit: I tried this and surprisingly it didn't help at all  (green line).

With respect to `-Yinduction-heuristics`, it unfortunately seems to make things worse; roughly equivalent to +1 test size.  (purple line)

screen shot 2017-02-14 at 9 36 53 am

Finally, I think I will move the simpler `Union`/`Intersection` part of this PR to a separate PR, because that patch won't cause bincompat issues, so can probably be merged sooner.  Still requires choosing among (1a), (1b), (1c).

Thanks for reading this far ;-)

milessabin:
(1c) ... thanks :-)

milessabin:
@refried could you and @joroKr21 work on combining this PR with #795.

joroKr21:
I don't think #795 is any faster. I hadn't realized `FilterNot` was implemented in such a weird way. This is a simpler and more general solution :+1:

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -1276,10 +1329,10 @@ object hlist {
       }
 
     // let (H :: T) ∪ M  =  H :: (T ∪ M) when H ∉ M
-    implicit def hlistUnion1[H, T <: HList, M <: HList]
+    implicit def hlistUnion1_[H, T <: HList, M <: HList]
       (implicit
        u: Union[T, M],
-       f: FilterNot.Aux[M, H, M]
+       f: NotContainsConstraint[M, H]
```
joroKr21: I wonder if it's better to move `NotContainsConstraint[M, H]` before `Union[T. M]`. In theory it should save some unnecessary computations but in practice I couldn't see a difference.

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -1333,10 +1399,10 @@ object hlist {
       }
 
     // let (H :: T) ∩ M  =  T ∩ M  when H ∉ M
-    implicit def hlistIntersection1[H, T <: HList, M <: HList]
+    implicit def hlistIntersection1_[H, T <: HList, M <: HList]
       (implicit
        i: Intersection[T, M],
-       f: FilterNot.Aux[M, H, M]
+       f: NotContainsConstraint[M, H]
```
joroKr21: ditto

joroKr21:
@aryairani it's been long time. Sorry about that. We are finally looking at a 2.4.0 release soon and I think this would be an important part of it. But the PR needs a rebase to resolve conflicts.

joroKr21:
actually might make sense to check scala 2.12.10 / 2.13.1 first

aryairani:
So are we not worried about bincompat for the purposes of this PR (due to the major version update)?  I see some of the surrounding methods have already seen their signatures changed.

joroKr21:
Yes, that's right we are breaking bincompat for 2.4.0

aryairani:
Okay, tests passed locally for me; let's see what CI says.

joroKr21:
@aryairani thanks for coming back to this. Do you have the original code you used to test the performance? I want to try with newer versions of Scala to see if the performance gains are as dramatic as before.

aryairani:
@joroKr21 Sorry for the delay — 

That's a bit tough 3 years later. 😅 I don't seem to have code immediately accessible, i.e. haven't found it in Dropbox.  The computer I authored it and ran it on is currently in storage, so I can't easily check it.  I may have a backup of said machine in the house, on an external backup appliance that's currently boxed up in the basement.  

How badly do we want that original code? Inventing something from scratch might be easier.

WDYT?

joroKr21:
Ah, don't go through so much trouble. will try something from scratch.

joroKr21:
@aryairani I finally found time to try out the performance. Tried with an `HList` with 100 elements:
```scala
  type B = Boolean
  type I = Int
  type L =
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    B :: I :: B :: I :: B :: I :: B :: I :: B :: I ::
    HNil
  Partition[L, I]
```

I think we can optimise `Partition` and leave `Filter` and `FilterNot` as they are.
The key observation is the difference between:

  * 2 seconds:
```scala
    implicit def hlistPartition2[H, L <: HList, U, LPrefix <: HList, LSuffix <: HList](
      implicit e: U =:!= H, p: Aux[L, U, LPrefix, LSuffix]
    ): Aux[H :: L, U, LPrefix, H :: LSuffix]
```

  * Too long, didn't wait:
```scala
    implicit def hlistPartition2[H, L <: HList, U, LPrefix <: HList, LSuffix <: HList](
      implicit p: Aux[L, U, LPrefix, LSuffix], e: U =:!= H
    ): Aux[H :: L, U, LPrefix, H :: LSuffix]
```

joroKr21:
Do you have time to make that change?

aryairani:
So wait, then is there no performance difference between the new `Partition` and the proposed `Filter`/`FilterNot`?

I missed a comparison in https://github.com/milessabin/shapeless/pull/682#issuecomment-599198932.

joroKr21:
Maybe there is a small difference but nothing noticeable (i.e. all are linear now).

aryairani:
Okay, great.

PR #686, opened 2017-02-19
Reduce type loss in hlist ops with .Aux versions of inductive inputs

refried:
In the spirit of #650

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -205,14 +205,17 @@ object hlist {
   trait HKernelAux[L <: HList] extends DepFn0 { type Out <: HKernel }
 
   object HKernelAux {
+
+    type Aux[L <: HList, Out0 <: HKernel] = HKernelAux[L] { type Out = Out0 }
+
     implicit def mkHNilHKernel = new HKernelAux[HNil] {
       type Out = HNilHKernel
       def apply() = HNilHKernel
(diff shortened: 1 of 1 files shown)


milessabin:
I'm fine with this, but could you deal with the coverage drop?

refried:
Hm, even looking at the report, I have no clue why there is reduced coverage.  As a desperate move, I can try rebasing, but if that doesn't help, I'm not sure what to do.

refried:
(just waking up) I see Github agrees there's only one different file vs master, so I don't think rebasing will help.  

Is there any chance Codecov is wrong?

milessabin:
Hmm ... good point. If you rebase Codecov will rerun ... that might make the problem go away.

refried:
Closing and reopening to try to trigger codecov again.  The current negative coverage delta doesn't make any sense to me, I didn't change any calls to code, only refined the types without breaking any tests.

milessabin:
I've asked for help on the gitter channel ...

As far as I'm concerned this is a Codecov problem, not yours, but I'm going to hold off on merging a little longer in the hope that someone has an idea what's going on.

dwijnand:
@travisbrown is also having coverage problems. He sees it as a scoverage problem (if it's the same problem).

https://github.com/circe/circe/pull/571

refried:
Roger.

milessabin:
Merging this now ... many thanks :-)

PR #688, opened 2017-02-22
Add .codecov.yml and set threshold

n4to4:
Same as https://github.com/circe/circe/issues/572

See also [.codecov.yml spec](https://github.com/codecov/support/wiki/Codecov-Yaml#full-yaml)


milessabin:
LGTM ... many thanks ... merging!

PR #689, opened 2017-02-24
Update Coursier and Ammonite versions for try-shapeless.sh

n4to4:


milessabin:

PR #691, opened 2017-02-28
Adds `TypedRecordArgs`

jeremyrsmith:
`TypedRecordArgs` is just like `RecordArgs`, but allows you to specify the return type of the dynamic methods.

The primary use case is that IDEs like IntelliJ have trouble inferring types that are refined by macro expansion. If the result of `applyDynamicNamed` has a fixed upper bound on its return type before macro expansion, then it can be inferred more easily by using `TypedRecordArgs`.

### DIFF
--- a/core/src/main/scala/shapeless/records.scala
+++ b/core/src/main/scala/shapeless/records.scala
@@ -89,6 +89,17 @@ trait RecordArgs extends Dynamic {
 }
 
 /**
+  * Similar to [[RecordArgs]], but allows the specification of the return type of the
+  * methods.
+  *
+  * @tparam T The return type for all record-accepting methods
+  */
+trait TypedRecordArgs[T] extends Dynamic {
+  def applyDynamic(method: String)(): T = macro RecordMacros.forwardImpl
+  def applyDynamicNamed(method: String)(rec: Any*): T = macro RecordMacros.forwardNamedImpl
+}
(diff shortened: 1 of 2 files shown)


milessabin:
I'm very reluctant to add a workaround for an IDE bug.

jeremyrsmith:
@milessabin I confess that's my motivation, but I think it conceptually makes sense as well. If I have for example:

```scala
implicit class MyOps[T <: HList](self: MyThing[T]) extends RecordArgs {
  def applyRecord[R <: HList](r: R)(implicit selectAll: SelectAll[T, R]): MyThing[T] = ???
}
```

then there's no reason that `applyDynamicNamed` should be typed to return `Any` when it can be constrained to `MyThing[T]`. You could even look at `RecordArgs` as a special case of `TypedRecordArgs[Any]`, but I didn't propose this because it would break MiMa.

It may be true that an IDE's inability to poke deeply into the result type of a `Dynamic` application during editing could be considered a bug, but since it involves expanding macros and resolving implicits, it also makes sense to give it constraints ahead of time.

I'll close this PR if the idea bugs you, but I'll just end up defining my own `applyDynamicNamed` that directly invokes `RecordMacros.forwardNamedImpl` all over the place – and I suspect I'm not the only one (though I have no evidence of that).

PR #692, opened 2017-03-03
Add TypeOf utility

Atry:


milessabin:
Could you add a test where the expression performs a type-level computation via an implicit?

Atry:
@milessabin The last one is via implicit: 
```
val tu2: Either[String, TypeOf.`the.apply: Bar[String]`.U] = Right(23)
```

Note the `the.apply`

milessabin:
Ahh, right. I really meant something a little more sophisticated than that, which justifies using this construct rather than expanding the type by hand. An non-trivial type-level computation would be good here (maybe a `Sum` of `Nat`'s), and maybe a use of `Generic[Foo].Repr`.

`the.apply` is fairly odd usage TBH ... I wouldn't expect to see it in real code.

Atry:
@milessabin I added a zipWithIndex test on HList

Atry:
One more thing, I think it would be possible to create a compiler plugin, which support the following syntax:

``` scala
T@like(expr)
```

which converts to

``` scala
TypeOf.`expr:T`.type
```

Atry:
For example:
```
type T = Any@like(Generic[(Int, String, Boolean)].to((1, "foo", true)).zipWithIndex)
```
As a result, `T` is `(Int, _0) :: (String, _1) :: (Boolean, _2]) :: HNil`

milessabin:
Interesting. I think you're getting into refined and singleton-ops territory here.

milessabin:
A test using `Generic`?

Atry:
@milessabin I hope it can perform type-level ops like `singleton-ops`. However, it is evaluated before method applying, preventing it get the actually type of a value.

milessabin:
Or should that be,
```scala
implicit val genBaz: TypeOf.`Generic[Baz]`.type = cachedImplicit
```

milessabin:
Anyhow, the point is that this would capture the `Repr` type in the type ascription which is quite awkward to do otherwise. See [here](https://github.com/milessabin/shapeless/issues/390#issuecomment-170224924).

milessabin:
Or, rather, [here](https://github.com/milessabin/shapeless/issues/390#issuecomment-183993578).

Atry:
Of course, `TypeOf` can infer `Repr` from implicits. https://github.com/milessabin/shapeless/pull/692/commits/ca4545728c229f3e339b0c5eb551ab521b4a979e#diff-9e42d9f002bf1026dba49cfe00ca4ebbR175

milessabin:
Great stuff. It's not immediately obvious that that would work, because the interactions with `cachedImplicit` and the semantics around implicit definitions with and without type ascriptions are quite subtle. I could easily believe that the val might be initialized to `null` in that case.

It's great that this works ... it's a really useful addition to the toolbox ... thanks! :-)

Atry:
I guess `the` would work as well:
``` scala
implicit val genBaz: the.`Generic[Baz]`.type = cachedImplicit
````

milessabin:
Oh! Interesting ... that hadn't occurred to me ... does it work?

Atry:
Works for me

PR #694, opened 2017-03-04
Fix Mapped for Id and Const

### LINKED ISSUE #693: Mapped broken for Id and Const
It seems that hlist's `Mapped` implicit fails to resolve for `Id` (for `HNil`) and `Const`:
```scala
// ambiguous implicit values:
//  both method hlistIdMapped in object Mapped of type [L <: shapeless.HList]=> shapeless.ops.hlist.Mapped.Aux[L,shapeless.Id,L]
//  and method hnilMapped in object Mapped of type [F[_]]=> shapeless.ops.hlist.Mapped.Aux[shapeless.HNil,F,shapeless.HNil]
implicitly[Mapped.Aux[HNil, Id, HNil]]

// These simply don't resolve
implicitly[Mapped.Aux[HNil, Const[Int]#λ, HNil]]
implicitly[Mapped.Aux[Double :: String :: HNil, Const[Int]#λ, Int :: Int :: HNil]]
```
Tested with scala 2.12.1.

nigredo-tori:
Fixes #693

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -70,7 +70,9 @@ object hlist {
 
     implicit def hnilMapped[F[_]]: Aux[HNil, F, HNil] = new Mapped[HNil, F] { type Out = HNil }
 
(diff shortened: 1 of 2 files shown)


milessabin:

PR #696, opened 2017-03-10
LiftAll for dependent type classes

Atry:
Note that this PR depends on #695

### DIFF
--- a/core/src/main/scala/shapeless/typeoperators.scala
+++ b/core/src/main/scala/shapeless/typeoperators.scala
@@ -201,3 +201,34 @@ object Lub {
   }
 }
 
+trait WitnessThe[TypeClassConstraint] extends DepFn0 {
+
+  type Out <: TypeClassConstraint
+
+}
+
+object WitnessThe {
+
+  type Aux[TypeClassConstraint, Out0] =
+    WitnessThe[TypeClassConstraint] {
+      type Out = Out0
+    }
+
+  final class WitnessTheAux[TypeClassConstraint, Out0 <: TypeClassConstraint](override val apply: Out0)
+      extends WitnessThe[TypeClassConstraint] {
+    type Out = Out0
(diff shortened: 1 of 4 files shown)


core/src/main/scala/shapeless/typeoperators.scala
```diff
@@ -201,3 +201,34 @@ object Lub {
...
+  @macrocompat.bundle
+  private[WitnessThe] final class Macros(val c: whitebox.Context) {
+    import c.universe._
+    def apply[TypeClassConstraint: WeakTypeTag](typeClass: Tree): Tree = {
+      q"new _root_.shapeless.WitnessThe.WitnessTheAux[${weakTypeOf[TypeClassConstraint]}, ${typeClass.tpe}]($typeClass): _root_.shapeless.WitnessThe.Aux[${weakTypeOf[TypeClassConstraint]}, ${typeClass.tpe}]"
+    }
+  }
+
+  implicit def apply[TypeClassConstraint](implicit typeClass: TypeClassConstraint)
+    : WitnessThe.Aux[TypeClassConstraint, /* typeClass.type */ _ <: TypeClassConstraint] =
```
milessabin: Can you motivate the wildcard here for me?

milessabin:
Why is this a new type class rather than an adjustment to the instances for the existing `LiftAll`?

milessabin:
We can move this to 2.4.0 in that case.

joroKr21:
#695 was closed. Does mean this is no longer viable?

milessabin:
@joroKr21 I was never really happy with the way #695 evolved and both it and this PR seem to have been abandoned. So I think close this one too, unless you or @Atry feel motivated to pick them up and get them over the finishing line.

joroKr21:
Closing because the PR this depends on was not accepted.

PR #697, opened 2017-03-10
Don't erase Out type for LiftAll

Atry:


milessabin:
Could you add a test that verifies that the refinements are preserved?

milessabin:
I don't want to merge this without tests.

PR #700, opened 2017-03-17
HList.IndexOf / Coproduct.IndexOf type operation

ctongfei:
Implements HList.IndexOf[L, X], which returns a Nat that is the index of the first occurrence of X in HList L.

### DIFF
--- a/core/src/main/scala/shapeless/ops/coproduct.scala
+++ b/core/src/main/scala/shapeless/ops/coproduct.scala
@@ -1248,6 +1248,32 @@ object coproduct {
   }
 
   /**
+   * Typeclass witnessing a specific type is the ''i''th element of an [[Coproduct]].
+   * @author ctongfei (Tongfei Chen)
+   */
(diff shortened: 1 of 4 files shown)


milessabin:
That's great stuff! Thanks!

Before I merge ... would you mind adding the corresponding for `Coproduct`?

ctongfei:
Coproduct.IndexOf added with accompanying tests.

ctongfei:
Could you merge this now? I'd really like to use this in my project :-)

milessabin:
I must have missed the notification for your more recent comment ... sorry :-(

Before I merge, could you tweak the tests to verify that the expected index type has actually been computed: as it stands you've provided the types explicitly ... we need to be sure that they will be inferred.

ctongfei:
I wonder how I could write such a test.

milessabin:
Maybe something closer to the expected usage? Code a user might write wouldn't hard code the resulting index.

ctongfei:
Will do it in the next few days. Thanks :-)

milessabin:
I don't want to merge this without a test that the expected types have been inferred.

PR #699, opened 2017-03-16
Conversions between Coproducts and Eithers

mgzuber:
Closes #648

PR #701, opened 2017-03-21
Conversions between nested pairs and HLists

mgzuber:


mgzuber:
@milessabin Any comments either way?

milessabin:
Yes, this looks useful. The corresponding definitions for nested `Either` to `Coproducts would also be useful. I think there are a couple of issue which need to be resolved though.

First you've reused a lot of existing names (eg. `ToHList`) and added names which are very close to existing names (eg. `toTuple` vs. `tupled`). Can you come up with something which distinguishes your additions from what's already there and more clearly illustrates the different semantics?

Second, I'm not completely convinced at the way the nested pairs are terminated ... why not have `()` as the rightmost element, analogous to `HNil`?

milessabin:
I'll consider merging this if you rebase and address the issues raised in my previous comment.

mgzuber:
My apologies on having been MIA.

To address your above comments. 
- The definitions for nested `Either` to `Coproduct` was already merged in #699.
- I couldn't find a similar name to `toHList` in the tuple syntax. I'm of the opinion that this is a reasonable name, but am more than happy to change it if you still disagree
- I changed `toTuple` to `nestedProduct`
- Your second point is one of preference I think. I thought it convenient to be able to convert a 2-tuple directly to a HList and a length 2 HList directly to a 2-tuple. However, I don't have a strong opinion on this, so I've changed the behaviour to be as you described.

Let me know if you'd like more changes, I will get to them quickly!

core/src/main/scala/shapeless/syntax/hlists.scala
```diff
@@ -738,4 +738,9 @@ final class HListOps[L <: HList](l : L) extends Serializable {
    * Returns all combinations of exactly length `N` of elements from this `Hlist`
    */
   def combinations(n: Nat)(implicit combinations: Combinations[n.N, L]): combinations.Out = combinations(l)
+
+  /**
+   * Converts this `HList` into a nested pair
+   */
+  def nestedProduct(implicit hListToProduct: HListToProduct[L]): hListToProduct.Out = hListToProduct(l)
```
joroKr21:
```suggestion
  def toNestedProduct(implicit hListToProduct: HListToProduct[L]): hListToProduct.Out = hListToProduct(l)
```

I would also rename then the corresponding one to `toNestedEither`.

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -3030,4 +3030,57 @@ object hlist {
         def apply(l: HNil): Out = HNil
       }
   }
+
+  /**
+   * Type class supporting mappings from type `T` to an `HList`. Currently only supports mapping nested pairs to an `HList`
+   *
+   * @author Michael Zuber
+   */
+  sealed trait ToHList[-T] {
```
joroKr21:
```suggestion
  sealed trait ProductToHList[T] extends DepFn1[T] {
```
I think we should be consistent with #699 

@mgzuber I am also happy to make those changes if it's too late for you.

PR #703, opened 2017-04-13
Add Aux for hlists.SelectAll

lloydmeta:
Useful for the same reason records.SelectAll.Aux is useful.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -43,6 +43,7 @@ Kailuo Wang <kailuo.wang@gmail.com> @kailuowang
 Kenji Yoshida <6b656e6a69@gmail.com> @xuwei_k
 Kevin Wright <kev.lee.wright@gmail.com> @thecoda
 Lars Hupel <lars.hupel@mytum.de> @larsr_h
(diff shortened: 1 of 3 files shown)


lloydmeta:
That's odd. I added 1 line for Aux and a corresponding test for it, but now all these other places I hadn't touched are missing test coverage...any ideas?

milessabin:
The coverage drop looks like a glitch ... not your problem.

I'm a bit puzzled by this PR though ... the `Out` type of `SelectAll[C, P]` should be definitionally equal to `P`, so the align should be unnecessary, and more generally, I'm not completely convinced that the `Aux` is needed at all. Can you show me an example where this is indeed needed?

lloydmeta:
> the Out type of SelectAll[C, P] should be definitionally equal to P

Ah, you are correct, `type Out = S` is in the definition of `hlist.SelectAll`. I was trying to write a structural upcast (like in the [test case](https://github.com/milessabin/shapeless/pull/703/files#diff-1f14a875ec23de285bb153d924d7e6c2R3114)), and thought it might be necessary to use an `Align`, but I see now that I was wrong :)

Closing, because I no longer have a usecase.

PR #705, opened 2017-04-18
Removed allocation in object Tag

Arneball:


milessabin:
This is fine, but it's binary incompatible so it'll have to wait until the 2.4.0.

Can you point to an example project where the allocation you're removing here is an issue in practice?

Atry:
IIRC, Hotspot JVM can easily eliminate these kind of allocation.

Arneball:
Dalvik cannot do escape analysis AFAIK. Saw the allocation in a toy project
of mine.

Den fre 21 apr. 2017 15:21杨博 (Yang Bo) <notifications@github.com> skrev:

core/src/main/scala/shapeless/typeoperators.scala
```diff
@@ -23,14 +23,15 @@ import scala.reflect.macros.whitebox
...
 
 object tag {
-  def apply[U] = new Tagger[U]
+  def apply[U] = Tagger.asInstanceOf[Tagger[U]]
 
   trait Tagged[U]
   type @@[+T, U] = T with Tagged[U]
 
-  class Tagger[U] {
+  trait Tagger[U] {
```
joroKr21: Why does it have to be a trait?

joroKr21:
```suggestion
  class Tagger[U] {
```

PR #710, opened 2017-04-21
Removed unsound cases in appliedTypTree1

### LINKED ISSUE #652: Split1 fails to split monad transformers (from cats)
Example:
```scala
import cats._, data._, instances.all._
import shapeless._

type RSO[T] = Reader[String, Option[T]] // aka ReaderT[Id, String, Option[T]]
Split1[RSO, Monad, Monad] // cats.Id[A] takes no type parameters, expected: one
```

Note that `Id` also has a `Monad` instance, but it's not parameterized by `T`.

joroKr21:
Unsound because they changed the kind of participating types
rather than just replacing the type `param` with the type `arg`.

Incidentally, resolves #652

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -23,6 +23,7 @@ Eugene Burmako <xeno.by@gmail.com> @xeno_by
 Filipe Nepomuceno <filinep@gmail.com
 Frank S. Thomas <frank@timepit.eu> @fst9000
 George Leontiev <folone@gmail.com> @folone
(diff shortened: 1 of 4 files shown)


joroKr21:
There's something wrong with the codecov setup.

milessabin:
Thanks for this. Before I merge it, would you mind checking that this doesn't break kittens (which is currently the main user of `Split1`).

joroKr21:
Tested locally kittens/master with `2.3.3-SNAPSHOT` from this branch and all tests pass successfully.

milessabin:

PR #714, opened 2017-05-06
Deep Merge and Extract ops for Records

eugengarkusha:
Added the following ops:
1)DeepMerger(reused most of the code from Merger typeclass)
2)Shallow/Deep Extractor (extracts super-record from sub-record, witnesses depth/width subtype relation))
3)added an implementation of shapeless.testUtil.assertTypeEquals with more precise type equality check(and a comment describing the flaw in the existing impl).

@milessabin I remember that previously you told me about the rule of 1 TC per operation but maybe a tiny violation could be allowed for the case where operations are very similar?

this PR  follows up on gitter discussion:

_Ievgen Garkusha @eugengarkusha Feb. 24 13:35
Working on translating from dynamic json data model to static (Record-based) one I stumbled upon the absence of the following (potentially) useful operations
1)'deep merge' (currently only 'shallow' merge is supported by 'Merger')
2)'depth subtype test'( currently only breadth subtype test is available via 'SelectAll')
I'd like to provide the draft implementation of the following operations to your consideration and judgement
https://github.com/eugengarkusha/RecordsDeepMerge
@milessabin do you think there might me a place for these ops in Shapeless?
Thanks.

Miles Sabin @milessabin Feb. 24 14:26
@eugengarkusha yes, definitely_


milessabin:
This is great stuff. I think though that it would be better to leave the current `Merger` as it is and add an independent `DeepMerge` type class rather than try and combine them and then differentiate the behaviour via a mode type.

eugengarkusha:
@milessabin 
Thanks!

Justification:
DeepMerger adds only one step to existing Merger logic. The same is true for DeepExtractor(to Extractor) and  MergeWith(to its possible deep counterpart). It possible that other usages may be found for TraversalMode.

Could you pls confirm that I should separate **Merger and DeepMerger** (as well as **Extractor and DeepExtractor**).

milessabin:
Yes, please do. My main concerns here are that,
1. existing code using `Merge` continues to compile.
2. the mode type adds additional complexity.

eugengarkusha:
@milessabin 
I see, thanks for the explanation. 
Done.

milessabin:
Quick question: how does the shallow extractor differ from `SelectAll`?

eugengarkusha:
@milessabin 
While working on the examples I could not find any use case for shallow Extractor. Looks like SelectAll and DeepExtractor will be enough for most cases indeed.

1)removed (shallow)  Extractor 
2)Renamed DeepExtractor to Extractor
3)added example

milessabin:
Great stuff! Many thanks ... merging :-)

eugengarkusha:
@Thanks!

PR #716, opened 2017-05-07
Add PartialComapped type class

Atry:


Atry:
`PartialComapped` is like `Commaped`, except it skips elements that are not match `F[_]`.

milessabin:
Looks interesting, but could you provide a motivating example?

milessabin:
Still waiting for a motivating example.

joroKr21:
There doesn't seem to be much interest in this. Feel free to reopen with a motivating example.
But in the end of the day we can't cover all typelevel functions so it's ok for some of them to live in user land.

PR #722, opened 2017-05-22
Improve error message by adding @implicitNotFound

Atry:


milessabin:
Thanks!

I think the message would be a little clearer if it were worded as follows: "could not find Lazy implicit values for all of the types enumerated in ${T}".

Atry:
You are right, though I still wonder if there is a solution that report the specific type not found in `T`.

milessabin:
I think if you use Typelevel Scala with `-Yinduction-heurisitcs` you should get a more informative error message ... give it a try.

milessabin:
Sorry, this seems to have slipped through the cracks ... many thanks ... merging :-)

Issue #721, opened 2017-05-21
Have a Refute typeclass


zainab-ali:
There was some talk a while back on Gitter about having a `Refute` typeclass:

```scala
/** Evidence that an implicit A is not present */
trait Refute[A]

object Refute {
  implicit def ambiguous1[A](implicit ev: A): Refute[A] = new Refute[A] {}
  implicit def ambiguous2[A](implicit ev: A): Refute[A] = new Refute[A] {}
  implicit def refute[A]: Refute[A] = new Refute[A] {}
}
```

This is a useful tool when using inductive implicits.  Should this exist in shapeless?
If so, I'd be happy to add it

milessabin:
Yes, please do :-)

zainab-ali:
This has now been merged in

milessabin:
Thanks ... sorry I should have closed this earlier :-)

PR #723, opened 2017-05-22
Add Refute typeclass

zainab-ali:
Add `Refute` as discussed in #721

### DIFF
--- a/core/src/main/scala/shapeless/refute.scala
+++ b/core/src/main/scala/shapeless/refute.scala
@@ -1,0 +1,22 @@
+package shapeless
+
+import scala.annotation.implicitNotFound
+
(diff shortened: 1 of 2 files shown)


zainab-ali:
The nature of the ambiguous implicits hack means that I won't be able to hit the codecov checks, despite testing the typeclass

milessabin:
Great stuff! Thanks :-)

The codecov thing is irritating. You should be able to fix that by invoking the ambiguity producing methods explicitly ... I appreciate that this is gaming the coverage tool and doesn't have much value, but I think it's just about worth doing.

milessabin:
Many thanks for your persistence and patience with the coverage tool ... very much appreciated :-)

Anyhow, LGTM ... merging :-)

PR #724, opened 2017-05-23
Default Typeables for simple types and case classes get their names at compile time.

### LINKED ISSUE #707: Typeable.describe is using runtime reflection
I was under the impression that all data was injected at compile time but I was surprised to see in ensime-server that our use of Typeable (when we started using `BigDataFormat` typeclass derivation) was causing a runtime performance problem with `Typeable.describe` calling `Class.getSimpleName`.

A manual inspection of the `typeable.scala` file shows that there are indeed cases where the lookup is at runtime. It would be good to solidify this stuff at compile time.

hrhino:
Doesn't change behavior at all (other than that nested types which previously triggered scala/bug#5425 now are printed like everything else), but does plausibly increase efficiency slightly when large numbers of calls to `.describe` are being made.

Fixes #707.

### DIFF
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -107,20 +107,17 @@ object Typeable extends TupleTypeableInstances with LowPriorityTypeable {
     }
 
   /** Typeable instance for simple monomorphic types */
-  def simpleTypeable[T](erased: Class[T]): Typeable[T] =
+  def simpleTypeable[T](erased: Class[T]): Typeable[T] = {
+    namedSimpleTypeable(erased, safeSimpleName(erased))
+  }
+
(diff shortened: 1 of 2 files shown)


milessabin:
Hmm ... codecov seems to be reporting a coverage decline in _test code_ ... does anyone else think that that shouldn't be happening?

milessabin:
Other than that, this looks good ... thanks!

hrhino:
That's... odd indeed. I pushed up a mildly bogus test of the compatibility stub which will maybe make it less mad. &lt;shrug /&gt;

milessabin:
Apologies for the delaying getting back to you on this ...

fommil:
niiiiice, thanks @hrhino !

PR #726, opened 2017-06-02
Fix ops.record.Selector with tagged type for Scala 2.12+

krzemin:
I've played a bit with record types that contain tagged types inside. It seems that there are inconsistencies of how they are handled between Scala 2.{10,11} and 2.12+ when it comes to `ops.record.Selector` macro. I found that shape of `RefinedType` in https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/generic.scala#L516 is different in newer versions of the compiler.

This PR contains a test case that failed under 2.12+ and workaround solution that seems to fix the issue and works under all supported Scala versions.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -513,8 +513,12 @@ trait CaseClassMacros extends ReprTypes {
       val KeyTagPre = prefix(keyTagTpe)
       val KeyTagSym = keyTagTpe.typeSymbol
       fTpe.dealias match {
-        case RefinedType(List(v0, TypeRef(pre, KeyTagSym, List(k, v1))), _) if pre =:= KeyTagPre && v0 =:= v1 => Some((k, v0))
-        case _ => None
+        case RefinedType(List(v0, TypeRef(pre, KeyTagSym, List(k, v1))), _) if pre =:= KeyTagPre && v0 =:= v1 =>
+          Some((k, v0))
+        case RefinedType(List(v00, v01, TypeRef(pre, KeyTagSym, List(k, v1@RefinedType(List(v10, v11), _)))), _) if pre =:= KeyTagPre && v00 =:= v10 && v01 =:= v11 =>
+          Some((k, v1))
+        case _ =>
+          None
       }
     }
   }
--- a/core/src/test/scala/shapeless/records.scala
+++ b/core/src/test/scala/shapeless/records.scala
@@ -1096,4 +1096,18 @@ class RecordTests {
 
     assertEquals(fields.toList, List('x, 'y, 'z))
   }
+
+  @Test
+  def testSelectorTagged {
+    import shapeless.syntax._
+    import shapeless.tag
+    import shapeless.tag._
+
+    trait TestTag
+    case class FooT(bar: String @@ TestTag)
+    val lgt = LabelledGeneric[FooT]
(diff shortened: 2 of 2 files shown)


milessabin:
Thanks for the diagnosis and the patch!

From the diff it looks as though there an additional parent type in 2.12.x ... have you investigated what it is?

krzemin:
It seems that in 2.12+ tag type alias (`@@`) is resolved and particular parents appears as inlined on the same level in RefinedType, but up to 2.11 another RefinedType for the tagged type is nested as first parent.

I prepared minimal code example that briefly illustrates the difference.

```scala
import scala.reflect.runtime.universe._
import shapeless._
import shapeless.tag._

trait Tag
case class Test(field: String @@ Tag)

val x = tag[Tag]("abc")

reify(LabelledGeneric[Test].to(Test(x))).staticType.asInstanceOf[TypeRef].typeArgs.head.asInstanceOf[RefinedType].parents
```

This snippet can be run in Scastie:
* for 2.11.11: https://scastie.scala-lang.org/krzemin/m8BYOctdQQqFe3GXC4VJCw/1
* for 2.12.2: https://scastie.scala-lang.org/krzemin/m8BYOctdQQqFe3GXC4VJCw/0

I don't think that Shapeless handles tagged types / field types any different between Scala versions. But so far I didn't find any possible root cause of the difference.

krzemin:
What's interesting - when trying to construct `LabelledGeneric` for product by hand by composing `DefaultSymbolicLablleing`, `Generic` and `ops.hlist.ZipWithKeys`, we have:

```scala
val dsl = DefaultSymbolicLabelling[Test]
val gen = Generic[Test]
reify(ops.hlist.ZipWithKeys[dsl.Out, gen.Repr].apply(gen.to(Test(x)))).staticType.asInstanceOf[TypeRef].typeArgs.head.asInstanceOf[RefinedType].parents
```
which behaves identically on both 2.11/2.12, resulting with:

```
List(
  java.lang.String,
  shapeless.tag.Tagged[Playground.this.Tag],
  shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[java.lang.String("field")],java.lang.String with shapeless.tag.Tagged[Playground.this.Tag]]
): List[Type]
```

But when having a function similar to `LabelledGeneric.materializeProduct` which takes all these input type classes as implicit instances with Aux pattern:

```scala
def materializeProduct[T, K <: HList, V <: HList, R <: HList]
  (obj: T)
  (implicit lab: DefaultSymbolicLabelling.Aux[T, K],
   gen: Generic.Aux[T, V],
   zip: ops.hlist.ZipWithKeys.Aux[K, V, R],
   ev: R <:< V) = zip.apply(gen.to(obj))

reify(materializeProduct(Test(x))).staticType.asInstanceOf[TypeRef].typeArgs.head.asInstanceOf[RefinedType].parents
```

The similar difference between 2.11/2.12 is present.

milessabin:
I apologies for not getting back to you on this sooner ... I must have missed a notification from github :-(

#761 looks a little more general than this fix ... I'll ask the submitter to include your test case.

krzemin:
Sure, I'm also for merging more general solution instead of this quick fix.

@joroKr21 feel free to check your fix also on this test:
https://github.com/milessabin/shapeless/pull/726/files#diff-45761b3cb0d957479a96d46507f6b3cbR1100

Closing.

milessabin:
Continued in #761.

PR #730, opened 2017-06-06
Added `Repeat` HList operation and syntax

jeremyrsmith:


milessabin:
This seems to be failing on scaladoc ... would you mind taking a look?

jeremyrsmith:
Looks like it didn't like how I linked the type params. I removed that and it should pass now.

milessabin:

PR #731, opened 2017-06-07
Add syntax for injection into coproducts

SystemFw:
This small PR adds syntax to inject values into coproducts when there is an `Inject` in scope:

```scala
import shapeless_, syntax.inject._

true.inject[String :+: Boolean :+: CNil]
// Inr(Inl(true))
```

### DIFF
--- a/core/src/main/scala/shapeless/syntax/inject.scala
+++ b/core/src/main/scala/shapeless/syntax/inject.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2017 Fabio Labella
+ *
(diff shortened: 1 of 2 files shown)


milessabin:
Would you mind adding a test for this? Otherwise looks good :-)

SystemFw:
ah, sorry, I couldn't find the relative tests, but they appear to be in `test/coproduct`. I'll add one shortly :)

SystemFw:
I have added a few tests, let me know if they are sufficient

milessabin:
Scoverage says yes :-)

PR #733, opened 2017-06-15
Expand gitignore for recent build changes

sarahgerweck:
There's now some more recursive folder nesting, so allow the hidden
files to be ignored deeper than just one level deep. Also, ignore the
`.js` and `.jvm` folders as well as the SBT jar cache folders.

@milessabin if you recall I think I provided the original version of this file, but it's been a year or so and there have been some changes that warrant a small update.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,18 +6,19 @@ scratch/
 target/
 
 # Ignore project files for Eclipse
-/.classpath
-/*/.classpath
-/.settings
-/*/.settings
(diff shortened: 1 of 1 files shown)


milessabin:
My apologies for quibbling last time ...

Many thanks! Merging :-)

sarahgerweck:
@milessabin nah, your initial perspective was reasonable. I have no bad feelings 😄

PR #736, opened 2017-07-06
Simplify and add syntax for runtime inject

SystemFw:
The implementation of RuntimeInject feels slightly more complicated than it could be; this small PR adds a simplified implementation with the following benefits:
 - Shorter.
 - More straightforward: just a linear scan of the Coproduct, rather than mutual typelevel recursion.
 - Roughly half the number of implicit instances needed.
 - Rules out at compile time the impossible case of trying to convert to CNil, instead of allowing it and then always returning None at runtime.

It also adds syntax, and more tests.

I have to say I am unsure whether submitting this PR counts as bikeshedding, if that's the case please consider it withdrawn, with my apologies.
Also, there might be a reason why the current implementation is the way it is, but I'm failing to see it, so I figured this PR would be a good way of asking anyway :)


milessabin:
My apologies for the delay in responding to this ...

Yup, LGTM ... many thanks ... merging :-)

SystemFw:
No worries :) thank you!

PR #737, opened 2017-07-08
Switch to scala-native 0.3.1

alexarchambault:


milessabin:
I'm a bit uncomfortable about the curl | bash ... is this the recommended way of building with scala-native on Travis? Can you point to some discussion somewhere?

alexarchambault:
@milessabin This seems to be common now when using scala-native on Travis CI (see for example, [scodec-bits](https://github.com/scodec/scodec-bits/blob/series/1.1.x/.travis.yml#L36) or [scalacheck](https://github.com/rickynils/scalacheck/pull/337/files#diff-354f30a63fb0907d4ad57269548329e3R36)). This setup is the same for all projects, so it makes sense that it gets maintained from scala-native.

milessabin:
Fair enough.

PR #742, opened 2017-07-26
Respect @implicitNotFound by typechecking apply

AndreasKostler:
Quick stab at fixing #671. 
#672 doesn't work for me...

### DIFF
--- a/core/src/main/scala/shapeless/typeoperators.scala
+++ b/core/src/main/scala/shapeless/typeoperators.scala
@@ -20,7 +20,7 @@ import scala.language.dynamics
 import scala.language.experimental.macros
 
 import scala.reflect.macros.whitebox
-import scala.util.Try
+import scala.util.{ Try, Success, Failure }
 
 object tag {
   def apply[U] = new Tagger[U]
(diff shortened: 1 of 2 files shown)


milessabin:
This looks pretty good to me ... @Atry does this work for you as well?

One improvement would be to lose the `parse` by using the 2.11 `typeCheck` which has typemode support ... macrocompat supports this for 2.10.x as well.

Would you mind losing the minor whitespace changes.

Atry:
@milessabin I did not update the original PR as I prefer the macro less `SingletonOf` you've mentioned.

https://static.javadoc.io/com.thoughtworks.feature/the_2.11/2.0.0/com/thoughtworks/feature/The.html

AndreasKostler:
Is there any more work required?

milessabin:
Before I merge this, can you try with `c.inferImplicitValue(tpe, silent = false)` wrapped in `Try` rather than typechecking `the`. If that does the right thing I think it would be preferable.

AndreasKostler:
I tried that first...Alas, the error is not the one you'd expect:
```scala
Try(c.inferImplicitValue(tpe, silent = false)) match {
      case Success(x) => Literal(Constant(())).setType(x.tpe)
      case Failure(e) => c.abort(c.enclosingPosition, e.getMessage)
    }
```
```scala
type foo = the.`Ordering[Set[Int]]`.Ops
<console>:14: error: implicit search has failed. to find out the reason, turn on -Xlog-implicits
       type foo = the.`Ordering[Set[Int]]`.Ops
```

milessabin:
Thanks for trying that out :-)

OK, LGTM ... many thanks ... merging!

PR #751, opened 2017-08-18
Fix #749

### LINKED ISSUE #749: Cannot implicitly witness a parameter
``` scala
Welcome to the Ammonite Repl 1.0.1
(Scala 2.12.3 Java 1.8.0_131)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import shapeless.Witness 
import shapeless.Witness

@ Seq("xxx").map { s =>
    implicitly[Witness.Aux[s.type]] 
  }  
cmd1.sc:2: could not find implicit value for parameter e: shapeless.Witness.Aux[s.type]
  implicitly[Witness.Aux[s.type]] 
            ^
Compilation Failed
```

Atry:


milessabin:
My hunch is that this might cause problems in cases like,
```scala
def foo(s: String)(implicit w: Witness[s.type]) = ...
```

Atry:
Does TypeLevel Scala's `ValueOf` compile is cases like this?
``` scala
def foo(s: String)(implicit w: ValueOf[s.type]) = ...
```

milessabin:
Superceded by https://github.com/milessabin/shapeless/commit/e70f01ecdcca9c2a991fb045b002dc32e9d88388.

PR #754, opened 2017-08-23
New encoding of Witness

Atry:


milessabin:
I'm going to close this for now ... eventually `Witness` will be retired in favour of SIP-23.

PR #759, opened 2017-09-04
Fix #758

### LINKED ISSUE #758: Generic fails to materialize ADT defined in trait
A bit similar to https://github.com/milessabin/shapeless/issues/212, but with an ADT instead of a case class.

```scala
trait Defs {
  sealed abstract class Foo extends Product with Serializable

  object Foo {
    case object A extends Foo
    case object B extends Foo
  }
}

class Inst extends Defs {
  val gen = Generic[Foo]
}
```
fails to compile with
```
could not find implicit value for parameter gen: shapeless.Generic[Inst.this.Foo]
```

Changing `Generic[Foo]` to `Generic.materialize[Foo, Foo.A.type :+: Foo.B.type :+: CNil]` gives more details:
```
Sealed trait Inst.this.Foo has no case class subtypes
```

Changing the definition of `Defs` above to
```scala
trait Defs {
  sealed abstract class Foo extends Product with Serializable

  case object A extends Foo
  case object B extends Foo
}
```
gives the same errors.

alexarchambault:


alexarchambault:
Not sure the codecov error is meaningful (that PR only modifies one line from the main sources, and adds 19 lines of tests).

milessabin:
LGTM. Many thanks ... merging!

PR #760, opened 2017-09-08
Delayed forcing of type signatures in isCaseAccessorLike

### LINKED ISSUE #757: Generic fails to derive on objects
```scala
import shapeless._

case object Foo {
  implicit val generic = Generic[Foo.type]
}
```

fails to compile with

```
could not find implicit value for parameter gen: shapeless.Generic[Foo.type]
[error]   implicit val generic = Generic[Foo.type]
[error]                                 ^
```

and

```scala
import shapeless._

case object Foo {
  implicit val labelled = LabelledGeneric[Foo.type]
}
```

fails to compile with

```
could not find implicit value for parameter lgen: shapeless.LabelledGeneric[Foo.type]
[error]   implicit val labelled = LabelledGeneric[Foo.type]
[error]                                          ^
```

However, the following succeeds

```
import shapeless._

case object Foo

object Main {
  implicit val generic  = Generic[Foo.type]
  implicit val labelled = LabelledGeneric[Foo.type]
}
```

showing that there is a problem with deriving a `case object`  (or `object`) on itself (i.e. its own companion)

joroKr21:
There is no point in calling `isNonGeneric` if the other tests fail. However, it is more robust to avoid forcing the type signature whenever possible, see the added test e.g.

Resolves #757

### DIFF
--- a/core/src/test/scala/shapeless/generic.scala
+++ b/core/src/test/scala/shapeless/generic.scala
@@ -122,6 +122,10 @@ package GenericTestsAux {
   sealed trait OB extends Overlapping
   case class OBC(s: String) extends OB
   case class OAB(i: Int) extends OA with OB
+
+  case object COSemiAuto {
+    implicit val gen = Generic[COSemiAuto.type]
+  }
 }
 
 class GenericTests {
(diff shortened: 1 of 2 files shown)


core/src/test/scala/shapeless/generic.scala
```diff
@@ -122,6 +122,10 @@ package GenericTestsAux {
   sealed trait OB extends Overlapping
   case class OBC(s: String) extends OB
   case class OAB(i: Int) extends OA with OB
+
+  case object SelfGen {
+    val gen = Generic[SelfGen.type]
```
fommil: can you please make this `implicit`

fommil:
having this in a bugfix 2.3.x release would be really fantastic

core/src/test/scala/shapeless/generic.scala
joroKr21: Sure, although it won't change anything.

milessabin:
LGTM. Please add your details to CONTRIBUTORS and I'll merge :-)

joroKr21:
I am already in the contributors list :smile:

milessabin:
Ah ... right :-)

Merging ... thanks again :-)

PR #761, opened 2017-09-09
Fixed ops.record.Selector for refined types

### LINKED ISSUE #732: Selector doesn't work for swapped record
```
scala> case class Foo(a: Int, b: String)
defined class Foo

scala> val gen = LabelledGeneric[Foo]
gen: shapeless.LabelledGeneric[Foo]{type Repr = shapeless.::[Int with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("a")],Int],shapeless.::[String with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("b")],String],shapeless.HNil]]} = shapeless.LabelledGeneric$$anon$1@4f8bd016

scala> val swap = SwapRecord[gen.Repr]
swap: shapeless.ops.record.SwapRecord[shapeless.::[Int with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("a")],Int],shapeless.::[String with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("b")],String],shapeless.HNil]]]{type Out = shapeless.::[Symbol with shapeless.tag.Tagged[String("a")] with shapeless.labelled.KeyTag[Int,Symbol with shapeless.tag.Tagged[String("a")]],shapeless.::[Symbol with shapeless.tag.Tagged[String("b")] with shapeless.labelled.KeyTag[String,Symbol with shapeless.tag.Tagged[String("b")]],shapeless.HNil]]} = shapeless.ops.record.SwapRecord$$anon$20@7f207278

scala> val field: FieldType[Int, Witness.`'a`.T] = swap().head
field: shapeless.labelled.FieldType[Int,Symbol with shapeless.tag.Tagged[String("a")]] = 'a

scala> Selector[swap.Out, Int]
<console>:34: error: No field Int in record swap.Out
       Selector[swap.Out, Int]
               ^
```

`swap().head` has type `FieldType[Int, 'a]` but `Selector[swap.Out, Int]` can't find a field that has `Int` as key.

### LINKED ISSUE #734: Interaction between records and tagged types
Stumbled over this peculiar behavior.

```scala
sealed trait IdType
val tagged = tag[IdType]("hi")
val head1 = 'id ->> tagged
val head2 = field[Witness.`'id`.T](tagged)
val rec1 = head1 :: HNil
val rec2 = head2 :: HNil
sameTyped(rec1)(rec2) //They should have the same type

showType(rec1) //Prints String @@ A$A90.this.IdType with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("id")],String @@ A$A90.this.IdType] :: shapeless.HNil
showType(rec2) //Prints shapeless.labelled.FieldType[Symbol with shapeless.tag.Tagged[String("id")],String @@ A$A91.this.IdType] :: shapeless.HNil

rec1('id)
rec2('id) //Doesn't work
```

Not sure why, but the problem seems to be that FieldType is inlined in one case, and not the other.

joroKr21:
Refined types can have more than one parent before `KeyTag` and a block of declarations after it in normalized form, e.g.
```scala
Iterable[Int] with Tagged["tag"] with KeyTag["key"] {
  def reverse: Iterable[Int]
}
```
So we have to make sure to match all of them when generating `Selector` instances.

Fixes #734 and fixes #732

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -513,20 +513,16 @@ trait CaseClassMacros extends ReprTypes {
       val KeyTagPre = prefix(keyTagTpe)
       val KeyTagSym = keyTagTpe.typeSymbol
       fTpe.dealias match {
-        case RefinedType(List(v0, TypeRef(pre, KeyTagSym, List(k, v1))), _) if pre =:= KeyTagPre && v0 =:= v1 => Some((k, v0))
+        case RefinedType(v0 :+ TypeRef(pre, KeyTagSym, List(k, v1)), scope)
+          if pre =:= KeyTagPre && refinedType(v0, NoSymbol, scope, NoPosition) =:= v1 =>
+            Some((k, v1))
         case _ => None
       }
     }
   }
 
-  def unpackFieldType(tpe: Type): (Type, Type) = {
-    val KeyTagPre = prefix(keyTagTpe)
-    val KeyTagSym = keyTagTpe.typeSymbol
-    tpe.dealias match {
-      case RefinedType(List(v0, TypeRef(pre, KeyTagSym, List(k, v1))), _) if pre =:= KeyTagPre && v0 =:= v1 => (k, v0)
-      case _ => abort(s"$tpe is not a field type")
-    }
-  }
+  def unpackFieldType(tpe: Type): (Type, Type) =
+    FieldType.unapply(tpe).getOrElse(abort(s"$tpe is not a field type"))
 
   def findField(lTpe: Type, kTpe: Type): Option[(Type, Int)] =
     unpackHListTpe(lTpe).zipWithIndex.collectFirst {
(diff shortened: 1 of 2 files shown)


joroKr21:
I'm not sure if I have to modify `FieldType.apply` accordingly. Ideally I would like to call `.normalize`, but it's deprecated. And scalac doesn't seem to mind the non-normalized types.

joroKr21:
This is similar to #726 but a bit more general.

milessabin:
Could you verify that the test case from #726 passes with this PR, and include it here if so.

joroKr21:
done

milessabin:

cb372:
@milessabin I don't suppose you could release this anytime soon? I've just stumbled across a similar problem and confirmed that this PR fixes it for me.

PR #762, opened 2017-09-09
Don't import tupleNZippedOps from Prelude.

hrhino:
This will break in the next 2.13 milestone release (it's already broken in the community build) on account of the PR scala/scala#6035, which gave `Tuple2` and `Tuple3` an extension method named `transpose`.

That `-Ypredef` PR is looking really useful right about now.

### DIFF
--- a/core/src/test/scala/shapeless/tuples.scala
+++ b/core/src/test/scala/shapeless/tuples.scala
@@ -16,6 +16,8 @@
 
 package shapeless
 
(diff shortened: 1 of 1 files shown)


milessabin:
Thanks for this. I think that `transpose` is the wrong name for the operation in the stdlib though, so I'm going to push back [there](https://github.com/scala/scala/pull/6035#issuecomment-328484411) rather than merging this.

hrhino:
:D even better!

Issue #655, opened 2016-11-07
Drop support for Scala 2.10.x!


milessabin:
This issue is to gauge the level of interest in ongoing support for Scala 2.10.x.

As well as adding a :+1: or a :-1: please leave a comment with your reasons for and against.

vil1:
Being a professional user of scala for more than three years, I've not seen any project using scala 2.10.x for quite a long time. So from my (obviously biased) point of view, support of 2.10.x is an unworthy effort.

joescii:
I have not used shapeless in an sbt plugin myself, but I could see where some folks would want 2.10 until sbt 0.13.x dies off.

cvogt:
@joescii that could be achieved with bug-fixes only legacy releases for 2.10, which would allow master to be 2.11 only.

joescii:
@cvogt sounds good to me. Like I mentioned, I don't personally need it. Just want to make sure this consideration is on the radar.

fommil:
We need shapeless in ensime to support 2.10. although as long as there is a cross build available, we're OK not upgrading shapeless to the latest and greatest.

Milyardo:
I'd imagine the only real concern for dropping 2.10.x support would be for sbt plugins as well.

tpolecat:
I would like to support doobie for 2.10 at least through mid-2017 and possibly until the end of the year, so shapeless support at least through that timeframe would be helpful. But in the worst case I can slice off the API that depends on shapeless and offer it only for 2.11+ so it's certainly not a disaster if you decide to end support earlier.

gruggiero:
Our use case is a project with the Cloudera platform for Hadoop and Spark.
Alas Spark with this platform is still with Scala 2.10 only, so we (and anybody else who uses this platform) cannot upgrade to 2.11 until Cloudera does

shawjef3:
@gruggiero Is your team OK using the current shapeless release for 2.10, or do you expect new features that will be in 2.11 and 2.12 releases?

gruggiero:
Actually we use the 2.2.5 as we import it as a transitive dependency of another library, which is an old version for 2.10 compatibility.
This means that even an upgrade to a new Shapeless 2.10 version is not possible in this project.
The choice is between being OK with 2.2.5 or nothing :)

fanf:
We succesfully ended all support of scala 2.10 this spring, so we only have 2.11 now in Rudder. 
[...]

PR #764, opened 2017-09-12
Fixed semiauto derivation of Default instances

### LINKED ISSUE #747: Implicit Default[] instantiation fails when base trait with abstract method is used for companion object
The following example doesn't compile with Scala 2.11 and shapeless 2.3.2 (implicit Default[Sample] not found):

```scala
case class Sample(a: Int)
trait Companion {
  def default: Default[Sample]
}
object Sample extends Companion {
  val default = Default[Sample]  // Error: No Default[Sample] found
}
```

The compile error is triggered by defining the default method in the base trait and implementing it as a `val` in the object that derives from the trait. One of the following changes makes the code compile:
  
  * Removing the method in the base trait
  * Adding an explicit type for the val; e.g. `val default: Default[Sample] = ...`
  * Changing the `val` to a `def`
 
The code compiles with Scala 2.12.3.

joroKr21:
Achieved via refactoring of `DefaultMacros.materialize`.

Fixes #747

### DIFF
--- a/core/src/main/scala/shapeless/default.scala
+++ b/core/src/main/scala/shapeless/default.scala
@@ -217,80 +217,73 @@ class DefaultMacros(val c: whitebox.Context) extends CaseClassMacros {
 
   def materialize[T: WeakTypeTag, L: WeakTypeTag]: Tree = {
     val tpe = weakTypeOf[T]
-
-    if (!isCaseClassLike(classSym(tpe)))
-      abort(s"$tpe is not a case class or case class like")
+    val cls = classSym(tpe)
 
     lazy val companion = companionRef(tpe)
+    def altCompanion = companion.symbol.info
 
-    // Fixes https://github.com/milessabin/shapeless/issues/474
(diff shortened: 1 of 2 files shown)


joroKr21:
Some missing functionality in Scala 2.10 makes things more verbose. Depending on how #655 turns out, `DefaultMacros` can be simplified at a later point (probably true for other macros as well).

milessabin:

milessabin:
Although Travis showed this as green previously, rerunning the build after clearing the build caches results in a build failure on 2.10.x: https://travis-ci.org/milessabin/shapeless/builds/274500921.

Would you mind taking a look?

joroKr21:
This looks like a spurious error. I saw it in #765 but I couldn't reproduce it locally. The failing example has nothing to do with `Default`. This commit fixes the build: a531c48ef00df82b19c046828e5c0fea5fb3d305

milessabin:
Could you pull that commit out as a separate PR?

PR #765, opened 2017-09-16
Fixed some scoping issues with Generic Coproducts

### LINKED ISSUE #612: Generic.apply[T] finds instances for undesirable types
The following code works as expected:

``` scala
scala> import shapeless._
import shapeless._

scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed trait Foo
case class FooA(a: Int) extends Foo
case class FooB(b: Int) extends Foo

defined trait Foo
defined class FooA
defined class FooB

scala> Generic[Foo]
res0: shapeless.Generic[Foo]{type Repr = shapeless.:+:[FooA,shapeless.:+:[FooB,shapeless.CNil]]} = anon$macro$1$1@481aacda
```

However, when I add a method that produces an anonymous instance of `Foo`, I would want it to not produce a `Generic[Foo]` (compile time error), but it does anyway:

``` scala
scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed trait Foo
case class FooA(a: Int) extends Foo
case class FooB(b: Int) extends Foo
def fooC(c: Int): Foo = new Foo { val cc: Int = c }

defined trait Foo
defined class FooA
defined class FooB
fooC: (c: Int)Foo

scala> Generic[Foo]
res1: shapeless.Generic[Foo]{type Repr = shapeless.:+:[FooA,shapeless.:+:[FooB,shapeless.CNil]]} = anon$macro$2$1@4c1b4994
```

If `FooC` is a `class` (not `case class`), the compiler catches the issue:

``` scala
scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed trait Foo
case class FooA(a: Int) extends Foo
case class FooB(b: Int) extends Foo
class FooC(c: Int) extends Foo

defined trait Foo
defined class FooA
defined class FooB
defined class FooC

scala> Generic[Foo]
<console>:16: error: could not find implicit value for parameter gen: shapeless.Generic[Foo]
       Generic[Foo]
              ^
```

Is it possible / desirable to have the compiler act similarly in the second case as in the first?

### LINKED ISSUE #614: Compile error with Generic + ADT in a method
This code produces a compile error (full example: https://github.com/durban/akka-typed-examples/blob/d784a3132efbd9255437fcd709389bb5a27fe687/src/main/scala/com/example/shless/NestedGeneric.scala):

``` scala
package com.example.shless

import shapeless._

class NestedGeneric {

  def foo(): Unit = {

    sealed trait U
    object U {
      final case class C(i: Int) extends U
      final case object D extends U
    }

    Generic[U]
  }
}
```

The error is somewhat strange:

```
[trace] Stack trace suppressed: run last compile:compileIncremental for the full output.
[error] (compile:compileIncremental) scala.reflect.internal.Types$TypeError: value <none> is not a member of com.example.shless.NestedGeneric
```

If I run `last compile:compileIncremental`, I get a big stacktrace (attached). If I comment out the line `Generic[U]`, there is no error.

joroKr21:
Finding a stable path from the macro call site to all subtypes of a sealed trait is neither trivial, nor always possible.

Fixes #612 and fixes #614


joroKr21:
What dark magic is this? How is it possible it can find an implicit on my laptop but it can't find it on the Travis servers? Lowest priority implicit scope - the brains of unsuspecting programmers?

milessabin:
This is quite a drastic refactoring in a fairly sensitive area ... could you walk through it a bit?

core/src/main/scala/shapeless/generic.scala
```diff
@@ -306,8 +306,10 @@ trait CaseClassMacros extends ReprTypes {
       appliedType(tpe, List(typeOf[Any])).dealias
     else tpe
 
-  def isProductAux(tpe: Type): Boolean =
-    tpe.typeSymbol.isClass && (isCaseClassLike(classSym(tpe)) || HasApplyUnapply(tpe) || HasCtorUnapply(tpe))
+  def isProductAux(tpe: Type): Boolean = tpe.typeSymbol.isClass && {
+    val cls = classSym(tpe)
+    isCaseObjectLike(cls) || isCaseClassLike(cls) || HasApplyUnapply(tpe) || HasCtorUnapply(tpe)
```
joroKr21: The `isCaseObjectLike` check is (mostly) redundant, but it's much cheaper than `isCaseClassLike` when the object is not `case`. It will also cover objects with fields which failed before:
```scala
object O { val x = 42 }
Generic[O.type]
```

core/src/main/scala/shapeless/generic.scala
```diff
@@ -351,12 +353,11 @@ trait CaseClassMacros extends ReprTypes {
 
   def fieldsOf(tpe: Type): List[(TermName, Type)] = {
     val tSym = tpe.typeSymbol
-    if(tSym.isClass && isAnonOrRefinement(tSym)) Nil
-    else
-      tpe.decls.sorted collect {
-        case sym: TermSymbol if isCaseAccessorLike(sym) =>
-          (sym.name.toTermName, sym.typeSignatureIn(tpe).finalResultType)
-      }
+    if (tSym.isClass && (isCaseObjectLike(tSym.asClass) || isAnonOrRefinement(tSym))) Nil
```
joroKr21: ditto

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
+          val subst = thisType(sym).baseType(baseSym).typeArgs
+          for (param <- sym.typeParams)
+            yield baseArgs(subst.indexWhere(_.typeSymbol == param))
+        }
+
+        // Handle anonymous subclasses.
+        val isNamed = !isAnonOrRefinement(sym)
+        if (isNamed && sym.isStatic) {
+          if (sym.isModuleClass) sym.toType
+          else appliedType(sym.toType, substituteArgs)
```
joroKr21: If the subclass is globally accessible, there is no point to try the expensive stuff below.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
+        } else {
+          // Look for a path from the macro call site to the subtype.
+          val owner = sym.owner
+          val owners = ownerChain(if (isNamed) owner else owner.owner)
+          val prePaths = for (pre <- Iterator.iterate(basePre)(prefix).takeWhile(_ != NoPrefix))
+            yield (pre, owners.iterator.dropWhile(pre.baseType(_) == NoType))
```
joroKr21: There are two thing going on here:
1. We try all (sub)prefixes, so that if the super class is `outer.inner.T`, we look for the subclass both in `inner` and in `outer`, in that order. Before it was only `inner`.
2. For each prefix we try to find a path to the subclass. Such a path exists if one of the classes in the owner chain is a super class of the prefix. The change here is to use `baseType` instead of symbol equality (to accomodate inheritance).

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
+
+          // Find a path from a (sub-)prefix or the enclosing owner.
+          val (pre0, path) = prePaths.find(_._2.nonEmpty).getOrElse {
+            val enclosing = ownerChain(enclosingOwner)
+            val common = owners zip enclosing indexWhere { case (o1, o2) => o1 != o2 }
+            (NoPrefix, if (common < 0) Iterator.empty else owners.iterator drop common - 1)
```
joroKr21: This is a last attempt. If none of prefixes worked or there was no prefix, maybe there is a path from the enclosing owner of the macro call site to the subclass. Note that there is always a common owner due to `_root_`. If all owners are common, we just reference the raw type, otherwise we use the differing owners as a path (we drop the common to accommodate nesting and relative accessibility).

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
-              }
-            }
-          if(!isAccessible(ctor))
-            abort(s"$tpe has an inaccessible subtype $ctor")
-          if(ctor <:< baseTpe) Some(ctor) else None
+          // Construct a stable prefix from the path.
+          val pre = path.drop(1).foldLeft(pre0) { (pre1, part) =>
+            if (part.isType) part.asType.toTypeIn(pre1)
+            else abort(s"$tpe has a subtype $sym with unstable prefix")
+          }
```
joroKr21: At this point we have a (possibly empty) path from the macro call site to the subclass and we can make a prefix out of it. Since we already have all the symbols (in the owner chain), there is no need to fall back to calling `.member` with names like the old implementation. Instead we just make sure all symbols in the path are types (otherwise the subclass is hidden in a val, var or def) and call `toTypeIn`.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
+
+          if (isNamed) {
+            if (sym.isModuleClass) sym.toTypeIn(pre)
+            else typeRef(pre, sym, substituteArgs)
+          } else {
+            def ownerIsSubType = owner.typeSignatureIn(pre) <:< baseTpe
+            if (owner.isTerm && owner.asTerm.isVal && ownerIsSubType) singleType(pre, owner)
+            else abort(s"$tpe has a subtype $sym with unstable prefix")
```
joroKr21: The only change here is to check the owner of an anonymous class more strictly:
1. It should be a val (because `Generic` then uses its singleton type)
2. It should be a subtype of the base trait

milessabin:
Thanks for that!

I'm still nervous though. As you've probably seen typechecking is up to its eyeballs in mutable state. This means that even quite simple reorderings of things which might force a lazy typecheck can change behaviour, possibly resulting in hard to diagnose errors ... SI-7046 and related things are particularly sensitive in this respect.

Before I merge this, would you mind building and running the tests of a few representative shapeless-using projects?

joroKr21:
I guess I can. Which projects do you have in mind?

milessabin:
I guess circe and scodec would be a couple of good candidates.

joroKr21:
Ok, I successfully tested those and a few others I have locally. Only JVM though, did not have the patience for Scala.js
- [x] Circe
- [x] Scodec
- [x] Kittens
- [x] Scalacheck-shapeless
- [x] Frameless

I'm not sure how much that proves though. If there are any problems, they will arise with more complex user defined data types.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
   }
 
   def ctorsOfAux(tpe: Type, hk: Boolean): List[Type] = {
-    def collectCtors(classSym: ClassSymbol): List[ClassSymbol] = {
```
milessabin: Would you mind removing the whitespace changes and removals of redundant braces for now. I'm not opposed, but it's making it harder for me to review this PR.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
-      classSym.knownDirectSubclasses.toList flatMap { child0 =>
+    def collectCtors(classSym: ClassSymbol): List[ClassSymbol] =
+      classSym.knownDirectSubclasses.toList.flatMap { child0 =>
         val child = child0.asClass
         child.typeSignature // Workaround for <https://issues.scala-lang.org/browse/SI-7755>
-        if (isCaseClassLike(child) || isCaseObjectLike(child))
```
milessabin: Eg. here ... this is just a whitespace change but superficially looks more drastic.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
-        else {
-          val tc = tpe.typeConstructor
-          val paramSym = tc.typeParams.head
-          val paramTpe = paramSym.asType.toType
-          appliedType(tc, paramTpe)
-        }
+      val baseTpe = if (!hk) tpe else {
+        val tc = tpe.typeConstructor
+        appliedType(tc, tc.typeParams.head.asType.toType)
+      }
```
milessabin: Similarly with the removal of the intermediate val ... I'm not opposed to the change in principle, but it is making it harder for me to review this PR.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -380,82 +381,85 @@ trait CaseClassMacros extends ReprTypes {
...
-            val subst = c.internal.thisType(sym).baseType(baseSym).typeArgs
-            sym.typeParams.map { param =>
-              val paramTpe = param.asType.toType
-              baseArgs(subst.indexWhere(_.typeSymbol == paramTpe.typeSymbol))
-            }
+        def substituteArgs = {
+          val subst = thisType(sym).baseType(baseSym).typeArgs
+          for (param <- sym.typeParams)
+            yield baseArgs(subst.indexWhere(_.typeSymbol == param))
+        }
```
milessabin: I think the only real change here is the use of `param` rather than `paramTpe.typeSymbol`. I think I'm fine with that change ... would you mind minimizing the delta to just that?

core/src/main/scala/shapeless/generic.scala
milessabin: As with the other refactorings, I'm not opposed, but I would like some sense that this reordering is winning us something ... do you have any compile time benchmarks which show that this is a win?

milessabin:
Thanks for that.

I've gone through the PR fairly closely now. It's great stuff, but I'd like you to try and separate out the semantic/bugfix parts of it and move the whitespace and general refactoring changes to a separate PR.

Apologies for being so pernickety about this but, as I said, this is a particularly sensitive area.

core/src/main/scala/shapeless/generic.scala
joroKr21: No and I don't plan on making any just for this. It definitely won't make any difference compared to implicit resolution times. So it's fine to remove this branch. Maybe then coverage won't decrease.

joroKr21:
@milessabin I minimized the diff as much as possible. I guess it will always be controversial how people format branchy code.

joroKr21:
In a conversation with @fommil it occurred to me that subclasses might have extra type parameters that have to be existentially quantified
```scala
sealed trait Foo[A]
case class Bar[A](a: A) extends Foo[A]
case class Qux[A, B](a: A, b: B) extends Foo[A]
val gen = Generic[Foo[Int]]
implicitly[gen.Repr =:= Bar[Int] :+: Qux[Int, _] :+: CNil]
```

milessabin:
I think this looks great, but I'm going to punt it to 2.4.0 to give it a chance to go through a full RC cycle.

joroKr21:
I'm getting the feeling that I might have to revert b14a373, because `existentialAbstraction` performs eager beta-reduction on the underlying type which might result in failing unifications (esp. for type aliases) that worked before. Then again, this is a problem that has plagued Scala for a long time and Dotty was only able to solve it by proving first class support for type lambdas.

joroKr21:
Instead of reverting the existential quantification, I modified it to apply only when the subclass has free type parameters.

fspillner:
Would this change-set fix my issue described here: https://stackoverflow.com/questions/49851942/why-could-it-not-resolve-implicits-if-a-coproduct-member-is-generic-in-my-adt ?

joroKr21:
This won't work:
```scala
final case class GenericFragment[T](value: T) extends Fragment
object GenericFragment {
  implicit def encoder[T](implicit encoder: Encoder[T]): Encoder[GenericFragment[T]] =
    Encoder.instance(fragment => encoder.encode(fragment.value))
}
```
You can only go between:
```scala
Fragment <-> TextFragment :+: NumberFragment :+: GenericFragment[_]
```
I.e. you would need:
```scala
implicit def encoder: Encoder[GenericFragment[_]] = ???
```

joroKr21:
@milessabin this is the last item on the 2.4.0 milestone 😅 
But we should also take a look at krzemin/octopus#28

milessabin:
> this is the last item on the 2.4.0 milestone

:clap: 

> But we should also take a look at krzemin/octopus#28

Agreed ... what are your thoughts on it?

joroKr21:
> Agreed ... what are your thoughts on it?

Commented there, but I'm not sure why it was working better in Scala 2.12

joroKr21:
@milessabin I think we are now ready to release an RC. We can the check if it still has performance issues. Perhaps also try with a nightly Scala build.

PR #774, opened 2017-09-29
Fix exception message in csv example

custommonkey:
Pass error string to super RuntimeException so that .getMessage behaves as expected.

### DIFF
--- a/examples/src/main/scala/shapeless/examples/csv.scala
+++ b/examples/src/main/scala/shapeless/examples/csv.scala
@@ -45,7 +45,7 @@ Markus,Persson,32"""
 // Implementation
 
 /** Exception to throw if something goes wrong during CSV parsing */
-class CSVException(s: String) extends RuntimeException
+class CSVException(s: String) extends RuntimeException(s)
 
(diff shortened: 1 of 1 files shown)


milessabin:
I'm completely baffled as to why the 2.10 build is failing given the tiny delta in this PR ... do you have any ideas?

custommonkey:
No I'm stumped as well. I can't reproduce it locally, compiling against 2.10 works perfectly.

joroKr21:
This commit fixes the 2.10 build: a531c48ef00df82b19c046828e5c0fea5fb3d305. For some reason having more `Lazy` than necessary breaks implicit resolution for the huge example in `sexp.big` only on 2.10. I'm also convinced that `Lazy` and friends swallow OOM and other fatal errors (probably via `c.inferImplicitValue(tpe, silent = true)`. But obviously we can't be sure whether running out of memory is the problem or something else entirely.

milessabin:
No idea why Codecov suddenly woke up now!

Anyhow, this LGTM ... many thanks ... merging :-)

PR #775, opened 2017-09-29
Remove faulty implementation of Coproduct ZipOne

### LINKED ISSUE #772: ZipOne on coproducts throws runtime exceptions in most cases
Coproduct ops include a `ZipOne` typeclass, witnessing that zipping two coproducts is the coproduct of tuples of elements taken pairwise, e.g. 
```scala
ZipOne[Int :+: String :+CNil, Boolean :+: Double :+: CNil]
res0: ZipOne {type Out = (Int, Boolean) :+: (String, Double) :+: CNil
```
However, this evidence is clearly false, as you can't guarantee that you'll always have both elements when you zip. Matter of fact, it only compiles due to the use of  `Option#get` on the Coproduct head.

As a result, if you pass in anything but two `Inl`, it throws an exception.

Is this intended behaviour? Imho this op should either be removed, or changed to return a more meaningful type, i.e. a Coproduct of cats `Ior`. WDYT?

SystemFw:
Fixes #772. Obviously breaks binary compatibility.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -333,7 +333,9 @@ lazy val mimaSettings = mimaDefaultSettings ++ Seq(
       exclude[MissingMethodProblem]("shapeless.CaseClassMacros.FieldType"),
       exclude[MissingMethodProblem]("shapeless.SingletonTypeUtils.parseSingletonSymbolType"),
       exclude[MissingMethodProblem]("shapeless.ops.hlist#IsHCons.cons"),
-      exclude[MissingMethodProblem]("shapeless.ops.coproduct#IsCCons.cons")
+      exclude[MissingMethodProblem]("shapeless.ops.coproduct#IsCCons.cons"),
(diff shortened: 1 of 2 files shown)


SystemFw:
Is there anything holding this PR back? The removed code compiles but crashes at runtime, very non-shapeless behaviour :)

milessabin:
Sorry, it looks like this slipped though the cracks ... could you fix up the MiMa failures?

SystemFw:
Ehm, the failure is due to the fact that the typeclass is removed. Can one preserve binary compatibility when removing something? (If yes, I'll research and fix it)

milessabin:
You can add a MiMa exclusion for the removed type.

SystemFw:
Cool, I'll look into that and fix it. A bit busy at the moment but I'll get to it

SystemFw:
@milessabin this should be all ready now

milessabin:

PR #778, opened 2017-09-30
Add support for Generic mutable case-like classes

### LINKED ISSUE #767: Add support for Generic case-like classes with vars
... or drop case-like classes altogether. Either way make it consistent:
```scala
case class A(i: Int); Generic[A] // ok
case class B(var i: Int); Generic[B] // ok
class C(val i: Int); Generic[C] // ok
class D(var i: Int); Generic[D] // mutable AND no case? No Generic for you sir!
```

joroKr21:
Fixes #767

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,2 +1,4 @@
--Xms2048m
--Xmx2048m
+-Xms2g
+-Xmx3g
+-Xss2m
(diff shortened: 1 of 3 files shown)


milessabin:
LGTM ... many thanks ... merging.

PR #779, opened 2017-10-06
add missing type parameter in Generic.Aux scaladoc

jbgi:
Not completely sure it was not intended, let me know.


milessabin:
Well spotted ... thanks! It's just a doc change so I'm ignoring the unrelated 2.10.x build failure.

PR #780, opened 2017-10-07
Handle consecutive Option columns in examples/csv deriveHConsOption, …

hywelandrews:
…and update CSVExample

This commit fixes an issue in the CSVExample where consecutive empty columns handled by deriveHConsOption would attempt to parse the remaining string unconditionally. Guard against this when head is None and orElse is called.

Useful for those basing work from examples/csv.scala


milessabin:
Would you mind rebasing this ... hopefully the CI stack size increase that I just merged will make the 2.10.x failure go away.

milessabin:

PR #781, opened 2017-10-18
MkString on HList was overconstrained

Jacoby6000:


milessabin:
Looks good, but unfortunately not binary compatible.

I don't know if it's worth it, but I think a binary compatible alternative would be to use the existing signature, but give the implicit parameter a `null` default value and implement as you have done here.

Jacoby6000:
A binary compatible alternative wouldn't be any more useful than the
existing implementation.

Jacoby6000:
Oh my bad. I didn't read your entire message. It's late 😅  I'll try that tomorrow.

milessabin:
Looks good ... maybe add a comment to the effect that the implicit parameter list should be removed in the next binary compatibility breaking release?

Jacoby6000:
@milessabin done!

milessabin:

PR #782, opened 2017-10-21
Fix link to SYB paper in README

howyp:
The link to Simon Peyton Jones's paper on scrap your boilerplate seems to get re-directed to his home page. It looks like it used to point directly to the paper - is this the right link now?


milessabin:
Good catch! Yes, that's the right link ... many thanks ... merging :-)

Issue #783, opened 2017-11-01
shapeless.ops.record.RemoveAll Not Serializable


khsibr:
Hi,
An obvious and quick to fix issue, shapeless.ops.record.RemoveAll is not serializable. 

Ryadh

milessabin:
Well spotted ... fancy taking a swing at it?

khsibr:
sure :)

joroKr21:
Fixed by #784

PR #784, opened 2017-11-01
Fix for shapeless.ops.record.RemoveAll Not Serializable #783

khsibr:


milessabin:
Thanks! Could you add yourself to the contributors list as well?

milessabin:

PR #787, opened 2017-11-19
Use mkHListValue in LabelledMacros

limansky:
I've found that LabelledMacros contains a copy of mkHListValue inside. Just to avoid duplication.


milessabin:
Well spotted! Many thanks ... merging :-)

PR #788, opened 2017-11-27
Add an example - recursion schemes for ADTs

joroKr21:
Inspired by Tim Williams' talk ["Recursion Schemes by Example"](https://youtu.be/Zw9KeP3OzpU).

The example shows how we can automatically generate a functor type and morphisms for an ADT.

### DIFF
--- a/examples/src/main/scala/shapeless/examples/recursionschemes.scala
+++ b/examples/src/main/scala/shapeless/examples/recursionschemes.scala
@@ -1,0 +1,466 @@
+/*
+ * Copyright (c) 2014 Georgi Krastev (@joroKr21)
+ *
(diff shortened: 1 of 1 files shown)


milessabin:
This is great stuff! Many thanks :-)

I know it's a pain, but do you think you could dig in to why it's not compiling on 2.10.x?

joroKr21:
In 2.10 `ArrowAssoc` and `Ensuring` add an `x` member to anything via implicit conversion :grin:

joroKr21:
Should be fixed now.

Hit some issues related to pattern matching and dependent types solved by using the `Aux` pattern.

milessabin:

PR #791, opened 2017-12-06
Add implicit OrElse combinator

joroKr21:
Spin off from milessabin/kittens#73
* Like `Option.orElse` on the type level
* Like `Either` on the value level

I only added `fold` and `unify` to `OrElse`, but other methods from `Either` might also be useful.

### DIFF
--- a/core/src/main/scala/shapeless/orelse.scala
+++ b/core/src/main/scala/shapeless/orelse.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2017 Georgi Krastev
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


joroKr21:
Why does codecov say 0% of diff hit? That can't be true.

milessabin:
The codecov result are pretty flaky IME. I've almost reached the point where I think it might just be better to remove it from the build.

milessabin:
Anyhow ... good stuff ... many thanks ... merging :-)

joroKr21:
I think I may have made a mistake. I wanted to submit a version with strict `Primary` and lazy `Secondary`, but instead I submitted one where both are lazy. But I'm not sure which one is more useful in general.

PR #792, opened 2017-12-08
Make OrElse Primary subclass strict

joroKr21:
Follow up #791 

My only argument in favour is that this works best for kittens.


joroKr21:
The extra `Lazy` increases compile time :/

milessabin:

PR #794, opened 2017-12-09
  Refactoring of apply/unapply/constructor extractors 

### LINKED ISSUE #790: Figure out what to do with multiple parameter lists (esp. implicits) in Generic
```scala
case class Ord1[A: Ordering](value: A)
class Ord2[A: Ordering](val value: A)
Generic[Ord1[Int]] // works
Generic[Ord2[Int]] // doesn't work
```

joroKr21:
* Utilize for-comprehensions on `Option` to make them more readable.
* Consistently use `patchedCompanionSymbolOf`.
* Add tests for the status quo of implicit parameter lists.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -621,20 +621,14 @@ trait CaseClassMacros extends ReprTypes {
   }
 
   def isCaseClassLike(sym: ClassSymbol): Boolean = {
-    def checkCtor: Boolean = {
-      def unique[T](s: Seq[T]): Option[T] =
-        s.headOption.find(_ => s.tail.isEmpty)
-
-      val tpe = sym.typeSignature
-      (for {
-        ctor <- accessiblePrimaryCtorOf(tpe)
-        params <- unique(ctor.asMethod.paramLists)
-      } yield params.size == fieldsOf(tpe).size).getOrElse(false)
-    }
-
-    sym.isCaseClass ||
-    (!sym.isAbstract && !sym.isTrait && !(sym == symbolOf[Object]) &&
-     sym.knownDirectSubclasses.isEmpty && checkCtor)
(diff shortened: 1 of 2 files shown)


milessabin:
I'd like to see this split into separate PRs. I think the `alignFields` stuff is most likely reasonable, but I think it's going to change behaviour in some cases, so I'd like to see a test which illustrates that.

I'm not sure about the `nonImplicitParamsOf` stuff. I'm pretty sure that `val bar = gen.from(gen.to(foo))` could result in `bar` having different implicit arguments from `foo` ... the implicits of `bar` will be whatever is in scope at the point at which `gen` was materialized. In any case, I'd like to see this split out as a separate PR with some tests that explore that.

The `patchedCompanionSymbolOf` stuff is spot on.

The refactorings are probably fine, but I'd like to see them split out as a separate PR just so that they're disentangled from anything which changes semantics.

joroKr21:
When I think more about the implicit arguments, I don't see a reasonable way to capture them at call sites of `from`. I guess the current behaviour for case classes of capturing at definition site of `Generic` happened "by accident" rather than "by design". Maybe there should be no `Generic` instance in this case?

milessabin:
If the implicit constructor arguments aren't accessible vals then there shouldn't be a `Generic` instance.

joroKr21:
You mean that if they are accessible they should become part of the `Repr`?

milessabin:
I think that if we supported multiple constructor argument lists then if this,
```scala
class Foo(val a: A)(val b: B)
```
has a `Generic` instance then so should,
```scala
class Foo(val a: A)(implicit val b: B)
```

joroKr21:
I don't have motivation to deal with this right now, so I just removed `nonImplicitParamsOf`. The tests just assert the current behaviour.

joroKr21:
@milessabin this PR now only contains refactoring, `alignFields` is at #808

milessabin:

Issue #309, opened 2015-01-11
Interaction between shapeless FieldType, and Scalaz tagged types


alexarchambault:
When using _at the same time_, `Lazy`, labels, and _scalaz_ tagged types, implicits resolution with `Lazy` seems to fail:

``` scala
import scalaz.@@

case class Dummy(i: Int @@ CustomTag)
case class DummyTagged(b: Boolean, i: Int @@ CustomTag)

implicitly[TC[Int @@ CustomTag]]
implicitly[TC[Boolean]]


implicitly[TC[Dummy]] // Works for this one...

type R = Record.`'i -> Int @@ CustomTag`.T
val lgen = LabelledGeneric[Dummy]
implicitly[lgen.Repr =:= R]
implicitly[TC[R]] // ...but not for its record

implicitly[TC[DummyTagged]] // Does *not* work for this one...

type RT = Record.`'b -> Boolean, 'i -> Int @@ CustomTag`.T
val lgent = LabelledGeneric[DummyTagged]
implicitly[lgent.Repr =:= RT]
implicitly[TC[RT]] // ...nor for its record
```

Full example here: https://gist.github.com/alexarchambault/ebf6ac15fceccbd37fc5#file-tc-scala

(Original reason for using scalaz `@@` was https://github.com/scalaz/scalaz/issues/747, which was addressed in scalaz but not in shapeless at some point.)

When using shapeless' `@@` type instead of the one of scalaz, one runs into _runtime_ exceptions instead (I'll post an issue about it).

When using `Generic` instead of `LabelledGeneric` (like in https://gist.github.com/alexarchambault/ebf6ac15fceccbd37fc5#file-tc-nonlabelled-scala), it works.

milessabin:
I think if it works with `Generic` but not `LabelledGeneric` then there's a reasonably good chance that this  has a fairly straightforward fix.

milessabin:
This appears to be an interaction between the Scalaz encoding of tagging and shapeless's encoding of fields. 

With the following local definition of Scalaz tags,
```scala
type Tagged[A, T] = { type Tag = T; type Self = A }
type @@[T, Tag] = Tagged[T, Tag]
```
we get the following behaviour,
```scala
val rec = 'i ->> 23.asInstanceOf[Int @@ CustomTag] :: HNil
rec('i)

[...]

PR #797, opened 2017-12-24
Fix Lazy for tagged types

### LINKED ISSUE #584: Interaction between Lazy and tagged types
There is an interaction between `Lazy` and tagged types with respect to the automatic derivation of typeclasses; to reproduce it this is a simple example:

```
sealed trait Readable[T]

object Readable {
  implicit def taggedReadable[A, T, M[_, _]](
    implicit
    r: Lazy[Readable[A]],
    ev: @@[A, T] =:= M[A, T]
  ): Readable[M[A, T]] = ???

  implicit val stringReadable: Readable[String] = ???

  implicit val hnilReadable: Readable[shapeless.HNil] = ???

  implicit def hconsReadable[H, T <: HList](
    implicit
    rh: Lazy[Readable[H]],
    rt: Readable[T]
  ): Readable[H :: T] = ???

  implicit def genericReadable[T, TGen <: HList](
    implicit
    gen: shapeless.Generic.Aux[T, TGen],
    genR: Readable[TGen]
  ): Readable[T] = ???
}

type CustomerTag = W.`'Customer`.T
type CustomerId = String @@ CustomerTag

case class Customer(id: CustomerId, name: String)

implicitly[Readable[CustomerId]]
implicitly[Readable[Customer]]

sealed trait BookTag
type BookId = String @@ BookTag

case class Book(id: BookId, title: String)

implicitly[Readable[BookId]]
implicitly[Readable[Book]]
```

In this case the derivation fails for both `Customer` and `Book`. Interestingly enough removing the `Lazy` in `hconsReadable` it compiles; in this case I could just remove it but when the derivation is used in third party libraries that won't compile without Lazy this could be a problem.

joroKr21:
Fixes #584 

Funny thing, this is also a partial fix for #309, but it looks like some types are automatically dealiased by Scala (e.g. `lgen.Repr` and `Record.{..fields}.T`) and in this case scala/bug#10506 kicks in.

### DIFF
--- a/core/src/main/scala/shapeless/lazy.scala
+++ b/core/src/main/scala/shapeless/lazy.scala
@@ -180,7 +180,7 @@ trait OpenImplicitMacros {
   def openImplicitTpeParam: Option[Type] =
     openImplicitTpe.map {
       case TypeRef(_, _, List(tpe)) =>
-        tpe.map(_.dealias)
+        tpe.dealias
       case other =>
         c.abort(c.enclosingPosition, s"Bad materialization: $other")
     }
--- a/core/src/test/scala/shapeless/labelledgeneric.scala
+++ b/core/src/test/scala/shapeless/labelledgeneric.scala
@@ -93,7 +93,7 @@ object ScalazTaggedAux {
     def apply(): String
   }
 
-  object TC {
+  object TC extends TCLowPriority {
     implicit val intTC: TC[Int] =
       new TC[Int] {
         def apply() = "Int"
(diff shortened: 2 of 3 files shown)


melrief:
This is great, thank you @joroKr21 !

milessabin:
Does this change actually break the test you've modified?

joroKr21:
Yes, the unmodified test fails because of ambiguous implicits, since the workaround for #309 is not necessary when deriving a typeclass instance for case class. However, it cannot be completely removed, because it is still necessary for the tests in `LabelledGeneric211Tests`. The difference is that the types in `LabelledGeneric211Tests` are eagerly dealiased (I realized this happens in `existentialAbstraction`).

milessabin:
Gotcha ... I'm going to defer this to 2.4.0 and hopefully get a fix for https://github.com/scala/bug/issues/10506 into 2.13.0.

joroKr21:
@milessabin should we merge this for 2.4.0?

milessabin:
Yes, I think so.

PR #799, opened 2017-12-28
A few lintings

som-snytt:
The missing interpolator seems to be a mistake; the tuple adaptation a matter of style or tolerance.

The unused imports were unused.

A possible build tweak to lint against 2.12.x, whitelisting certain dubious practices, as when Woody Allen as Miles Monroe wakes up in Sleeper and everything bad for you is good. That was back when Woody Allen was good.

Left as separate commits for easy cherry-picking and choosing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,12 +40,17 @@ lazy val commonSettings = Seq(
 
   scalacOptions := Seq(
     "-feature",
-    "-language:higherKinds",
-    "-language:implicitConversions",
+    "-language:higherKinds,implicitConversions",
     "-Xfatal-warnings",
     "-deprecation",
     "-unchecked"
   ),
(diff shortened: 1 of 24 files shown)


som-snytt:
`PackExamples` fakes out the "selecting a val from `DelayedInit`" test because the selection is `examples.this.PackExamples.sb` instead of `PackExamples.this.sb`. Turned off that lint, but left lint on for examples. Certainly examples are the primary source for newcomers wondering "what do I need to import?"

som-snytt:
Sorry, I don't understand what codecov is trying to tell me.

milessabin:
@som-snytt thanks for this!

I also can't make a lot of sense of the codecov output ... I think I'm going to take it out altogether.

milessabin:
This all looks good, but I'm going to punt it until 2.4.0.

milessabin:
Merging now ... many thanks :-)

PR #802, opened 2018-01-01
Actualize versions in readme

limansky:
I've just updated shapeless and Scala versions in readme.


limansky:
Oops. Already done.

PR #798, opened 2017-12-27
Fix Typeable for inner classes

joroKr21:
Fixes #748

PR #803, opened 2018-01-05
  Avoid creating Typeable for general type projections 

### LINKED ISSUE #579: Typeable instance derived for nested classes can be unsafe

### LINKED ISSUE #848: TypeCase broken on shapeless 2.3.3 only in scala 2.12.x
I was upgrading a current project to scala 2.12.6 and got into the following behaviour:

  * TypeCase compiles and works for shapeless version 2.3.2 on both scala 2.11.x and 2.12.x
  * TypeCase fails to compile using shapeless version 2.3.3 on scala 2.12.x but compiles as works fine on scala 2.11.x

The following error is thrown:

```
No default Typeable for type Protocol.SpecificCommandOne capturing an outer type variable
[error]   private val someMessage = TypeCase[Payloads.Payload[SpecificCommandOne]]
[error]                                     ^
[error] one error found
```

I managed to extract an example with a structure very similar to what I have on the project (which is much more complex) here - https://gitlab.com/vascorsd/broken-shapeless

Also made a scastie version here - https://scastie.scala-lang.org/6cODaCGGR62ZXsxoLiYfkg

They could probably be reduced much more, but that's what I got.

On the project I was only using shapeless 2.3.2 and didn't get that error until I also upgraded some dependencies and the shapeless version got forced to become the 2.3.3. Here the evicted log from sbt:

```
[info]     * com.chuusai:shapeless_2.12:2.3.3 is selected over 2.3.2
[info]         +- io.circe:circe-generic_2.12:0.9.3 ()               (depends on 2.3.3)
[info]         +- org.tpolecat:doobie-core_2.12:0.5.3                (depends on 2.3.3)
[info]         +- org.reactivemongo:reactivemongo_2.12:0.12.5 ()     (depends on 2.3.2)
```

Currently I solved the problem with the followng line on my sbt, forcing the shapeless version down as I really wanted to do the upgrade of the project to scala 2.12.6:

```
  dependencyOverrides ++= Seq(
    "com.chuusai" %% "shapeless" % "2.3.2"
  )
```

Hope this helps. I'm not sure if my current solution may break something else unexpectedly.

I read other similar bugs about on the bug tracker about TypeCase and Typeable but they don't seem to be this exact case.

joroKr21:
Unless:
* the outer type is final and not parametrized
* the inner type is a case class

All other type projections are potentially unsound because the
inner type can capture type members of the outer type.

Closes #579
Closes #848


milessabin:
Could you add a test that `None` is yielded appropriately?

joroKr21:
You mean for an unrelated type?

milessabin:
I'm looking at the comment above the code you replaced,
> There is potential unsoundness if we allow a simple cast between two unparameterized types, if they contain values of an abstract type variable from outside of their definition. Therefore, check to see if any values have types that look different from the inside and outside of the type.
I'd like to see a test which verifies that unsound cases are still rejected.

milessabin:
Something like this maybe?
```scala
trait Outer {
  type T
  class Inner(val t: T)
  def boom(a: Any): Option[Inner] = a.cast[Inner]
}
val i = new Outer { type T = Int }
val ii = new i.Inner(23)
val s = new Outer { type T = String {
s.boom(ii) // If Outer compiles then this should be None
```

joroKr21:
There are existing tests for type projections in `testNested`. The example you described should be better handled in #798 but to test this I need to combine it with this PR or wait for one or the other to be merged.

joroKr21:
I found an unsound example:
```scala
  @Test
  def testTypeProjection(): Unit = {
    trait Outer {
      type T
      class Inner(val t: T)
    }

    trait OuterInt extends Outer {
      type T = Int
    }

    trait OuterStr extends Outer {
      type T = String
    }

    val typ = Typeable[OuterInt#Inner]
    val outStr = new OuterStr { }
    assertEquals(None, typ.cast(new outStr.Inner("boom")))
  }
```
However, note that the check in `master` is not enough to catch all such cases. It doesn't deal with: inheritance, methods, even more nested classes, etc. That's a very deep rabbit hole.

One option would be to just reject all type projections.

milessabin:
When you say "type projection" do you mean "abstract type member"?

milessabin:
Ahh, you mean type projections as the target of the cast?

joroKr21:
Yes, I mean instances of `Typeable[Type#Projection]`

joroKr21:
Apropo, regular pattern matching has the same problem. It will warn for erasure on `case _: Outer[Int]#Inner`, but not on `case _: OuterInt#Inner`

milessabin:
I just checked my earlier example,

```scala
@Test
def testRefined: Unit = {
  trait Outer {
    type T
    class Inner(val t: T)
    def boom(a: Any): Option[Inner] = a.cast[Inner]
  }
  val i = new Outer { type T = Int }
  val ii = new i.Inner(23)
  val s = new Outer { type T = String }
  //assert(!s.boom(ii).isDefined) // If Outer compiles then this should be None}
  s.boom(ii).map(_.t.length) // CCE
}
```
This blows up with a `ClassCastException`.

Conclusion: there should be no `Typeable` instances for types with abstract type members.

milessabin:
Interestingly, in your example, I think there _should_ be an instance for `Typeable[OuterInt#Inner]` because as seen from `OuterInt` `T` is _not_ abstract.

joroKr21:
@milessabin in your example we don't have a type projection, but a path dependent type. As such, the outer reference should be compared for equality (see #798).

Yes, I guess my example should have a `Typeable` instance according to the comments on `master` (although it is still unsound), but it doesn't. Looking more closely at the code `rtpe.asSeenFrom(tpe, tpe.typeSymbol.owner)` is nonsensical. We can't replace references to `Outer.this.type` with `OuterInt#Inner` because `Inner` is not a subclass of `Outer`. In effect removing this call to `asSeenFrom` would change nothing.

But what if there is a third class nested inside `Inner` that captures type variables? Should we check all nested `decls`? What about inheriting captured variables from a superclass? Check all nested `members`? Even if all cases are covered, the error won't make much sense to the user any more.

milessabin:
I'm going to defer to your judgement on this one.

joroKr21:
If you approve I will merge it 😄 - it's the last of a series of fixes in `Typeable` and doesn't make much sense to leave it for later.

Issue #728, opened 2017-06-02
Coproduct constructor does not respect covariance


atemerev:
```scala
import shapeless._
import shapeless.ops._
import shapeless.ops.coproduct._

trait A
case object Aa extends A

trait B
case object Bb extends B

type AB = A :+: B :+: CNil

val cA = Coproduct[AB](Aa) // does not compile :(
// could not find implicit value for parameter inj: shapeless.ops.coproduct.Inject[AB, Aa.type]

val aa: A = Aa
val cA = Coproduct[AB](aa) // does compile :)
```

milessabin:
It's the `Inject` type class rather than the constructor which is to blame here. Do you want to take a swing at a fix for this?

atemerev:
I will try, sure. Thank you for the pointer!

Atry:
Similar to #715

nickvcb:
Seems to affect Options as well

```
type AB = A :+: Option[B] :+: CNil

val bb: B = Bb
val cB = Coproduct[AB](Some(bb)) // does not compile :(
```
"could not find implicit value for parameter inj: shapeless.ops.coproduct.Inject[AB,Some[B]]"

@milessabin care to share more insights please?

Context:
Trying to use for marshallers between domain model and [`skeuomorph`](https://github.com/higherkindness/skeuomorph) generated Scala class from protobuf IDL.

`protobuf` 3 no longer allows `required` so all is optional => ``skeuomorph` generates` `coproduct`s of `Option` :-(

SemanticBeeng:
Seems to affect Options as well

```
type AB = A :+: Option[B] :+: CNil

val bb: B = Bb
val cB = Coproduct[AB](Some(bb)) // does not compile :(
```
"could not find implicit value for parameter inj: shapeless.ops.coproduct.Inject[AB,Some[B]]"

@milessabin care to share more insights please?
Is there any way to go around in latest version?

Context:
Trying to use for marshallers between domain model and [`skeuomorph`](https://github.com/higherkindness/skeuomorph) Scala class generated from protobuf IDL.

`proto3` no longer allows `required` so all is optional => ``skeuomorph` generates` `coproduct`s of `Option` :disappointed:

milessabin:
It's fixed in #804, but not released because it's been queued waiting for a binary breaking release. I guess that's really long overdue now.

PR #804, opened 2018-01-05
Make Inject respect covariance

jdrphillips:
Addresses issue https://github.com/milessabin/shapeless/issues/728

Note I've chosen the constraint `HH <: H` rather than an implicit `HH <:< H` because the latter causes an ambiguous implicit error, the fix for which is either an `<:!<` in the other implicit method or a `LowPriority` trait. I felt both of those would compile too slowly for such a common class.


milessabin:

PR #805, opened 2018-01-08
Ensure LowPriority is fine with the Aux pattern

alexarchambault:
I ran into issues with `LowPriority` when using it for type classes relying on the aux pattern… That PR fixes that.

### DIFF
--- a/core/src/main/scala/shapeless/lowpriority.scala
+++ b/core/src/main/scala/shapeless/lowpriority.scala
@@ -67,11 +67,18 @@ class LowPriorityMacros(val c: whitebox.Context) extends OpenImplicitMacros with
 
   def strictTpe = typeOf[Strict[_]].typeConstructor
 
+  // FIXME Also in LazyMacros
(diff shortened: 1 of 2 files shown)


alexarchambault:
I tested that a bit with my [case-app](https://github.com/alexarchambault/case-app) project, it seemed fine (in 2.12 at least).

milessabin:
I think you're the main user of `LowPriority`, so if you're happy with it that's fine by me :-)

core/src/main/scala/shapeless/lowpriority.scala
```diff
@@ -67,11 +67,18 @@ class LowPriorityMacros(val c: whitebox.Context) extends OpenImplicitMacros with
 
   def strictTpe = typeOf[Strict[_]].typeConstructor
 
+  // FIXME Also in LazyMacros
+  def stripRefinements(tpe: Type): Option[Type] =
```
jvican: Is there any reason why this wasn't done in `LazyMacros` too yet?

PR #806, opened 2018-01-10
Don't lose type information in HList ops apply

joroKr21:
Added a few missing `apply` methods while at it.


milessabin:

PR #807, opened 2018-01-13
Make FieldType and @@ abstract

### LINKED ISSUE #44: Runtime exception if the value of record field is Any or AnyRef
To reproduce:

```
scala> import shapeless._, record._, syntax.singleton._
scala> "foo" ->> (1: Any)
```

java.lang.ClassCastException: java.lang.Integer cannot be cast to shapeless.record$KeyTag
....

### LINKED ISSUE #309: Interaction between shapeless FieldType, and Scalaz tagged types

### LINKED ISSUE #557: tag[Tag](value) doesnt work with explicitly declared target type
```
@ import shapeless.tag._
import shapeless.tag._
@ trait NameTag; type Name = String @@ NameTag
defined trait NameTag
defined type Name
@ val name1 = tag[NameTag]("ben")
name1: String @@ NameTag = "ben"
@ val name2: Name = tag[NameTag]("ben")
Compilation Failed
Main.scala:321: type mismatch;
 found   : String("ben")
 required: shapeless.tag.Tagged[cmd4.NameTag] with String
val name2: Name = tag[NameTag]("ben")
                               ^
```

It seems to be a side-effect of type inference from the target (ie the `val`) flowing back to the expression assigned to it. Scalac is simplistically setting the type T in Tagger#apply to be the "target" type of the val, eg Name. gitter [context](https://gitter.im/milessabin/shapeless?at=56d82cbd048f9e65291bb89d).

Reordering the tag operation, to collect the value first and the tag type second, seems a nice solution:

```
import shapeless.tag
import shapeless.tag.@@
object tag2 {
    def apply[T](t: T) = new TaggedValue[T](t)
    class TaggedValue[T](t: T) {
      def @@[U]: T @@ U = tag[U].apply[T](t)
    }
  }

def name: Name = tag2("ben").@@[NameTag]
```

I like the symmetry of using `@@` here myself. 

Ive trialed this pattern in my own codebase with success, and wonder if it would be welcomed pushed up to shapeless proper? I'd propose offering the new form, retaining but deprecating the existing `tag` operation.

### LINKED ISSUE #584: Interaction between Lazy and tagged types

### LINKED ISSUE #644: Look into type tagging and value classes
See the conversation in the typelevel/scala Gitter room on [October 11, 2016 8:58 PM](https://gitter.im/typelevel/general?at=57fd44740aa72e3c5bf38a8e).

joroKr21:
aka. "You shall not dealias"

Fixes a bunch of problems caused by:
1. the compiler normalizing types behind your back;
2. the compiler failing to unify the normalized types.

Also has the additional benefit of printing more nicely in the REPL.
This will always box however. Would that be a problem?

TODO: Check which bugs are fixed, reference them here and add regression tests:
- [x] Should close #644 - check that value classes and arrays work (boxed)
- [x] Should fix #584 in a different way than #797 - add regression test
- [x] Should fix #557 - add regression test
- [x] Fully fixes #309 - removed workaround
- [x] Should fix #44 - similar to #644 (with boxing)


core/src/main/scala/shapeless/singletons.scala
```diff
@@ -147,22 +147,21 @@ trait SingletonTypeUtils extends ReprTypes {
...
-    val TaggedSym = typeOf[tag.Tagged[_]].typeConstructor.typeSymbol
+    val AtatTpe = typeOf[_ @@ _].typeConstructor
+    val AtatSym = AtatTpe.typeSymbol
 
-    def unrefine(t: Type): Type =
-      t.dealias match {
-        case RefinedType(List(t), scope) if scope.isEmpty => unrefine(t)
-        case t => t
-      }
+    def unrefine(t: Type): Type = t.dealias match {
```
joroKr21: This is probably unnecessary

joroKr21:
Sadly I think this is not allowed on Scala 2.10 so this has to wait for #655

core/src/main/scala/shapeless/labelled.scala
```diff
@@ -17,16 +17,19 @@
...
+import scala.reflect._
 
 import scala.reflect.macros.whitebox
 
 object labelled {
   /**
    * The type of fields with keys of singleton type `K` and value type `V`.
    */
-  type FieldType[K, +V] = V with KeyTag[K, V]
+  type FieldType[K, +V] <: V with KeyTag[K, V]
```
joroKr21: What's the purpose of `V` in `KeyTag`?

joroKr21: How about calling `FieldType[K, V]` `K ->> V`?

joroKr21:
`ClassTag` magic doesn't work. HALP!

core/src/main/scala/shapeless/typeoperators.scala
```diff
@@ -19,17 +19,24 @@ package shapeless
...
 
+import scala.reflect.ClassTag
 import scala.reflect.macros.whitebox
 import scala.util.{ Try, Success, Failure }
 
-object tag {
-  def apply[U] = new Tagger[U]
-
+private[shapeless] trait Tag {
+  type @@[+T, U] <: T with Tagged[U]
```
kwark: I have been giving this encoding a spin, for encoding/decoding tagged types to/from json:

* with the `@@` part of the private `Tag` class, I run into a problem when using play-json's format macro: 

```No instance of play.api.libs.json.Format is available for xxx.Tag.$at$at[java.lang.String, TaggingTest.FooTag] in the implicit scope (Hint: if declared in the same file, make sure it's declared before)```

* If I move the `@@` type to  the `tag` object (just a few lines below), it does work. Check-out this scastie: https://scastie.scala-lang.org/kwark/rnWsgsTQQQ2Fmsh97NMRdQ

* circe-generic seems to work with both variants

joroKr21: I introduced the trait only because Scala 2.10 doesn't support abstract types in objects / classes, but somehow it supports them when inherited.

joroKr21: Yes, that's true unfortunately.

milessabin:
Continuing the discussion from https://github.com/scala/community-build/pull/1047#issuecomment-576539231.

I think it's worth investigating which approach most smoothly transfers to Dotty. At least some uses of tagged types will be replaced by opaque types, but probably not all. It'd be good to see what works best for the cases which aren't handled well by opaque types.

joroKr21:
Progress!

The boxing issue is solved by scala/scala#8720 so we can have our cake and eat it too.

Dotty has better type inference and some simple cases that I tried work better than in Scala 2 with the current encoding of tagged types. I still need to come up with a better comparison (and also check how they are erased in dotty).

I agree that perhaps tagged types will not be completely replaced (because they are subtypes) but as a native solution I expect opaque types to become much more widely used, perhaps even as much as to make tagged types on shapeless 3 unwarranted.

joroKr21:
Changed milestone to 2.5.0 which will be 2.13 only.
If scala/scala#8720 is merged (I see no reason why not) then this has no downsides.

joroKr21:
> Changed milestone to 2.5.0 which will be 2.13 only.
> If scala/scala#8720 is merged (I see no reason why not) then this has no downsides.

In the end it was not merged. In Scala 3 we have opaque types.

PR #808, opened 2018-01-17
 Match fields by name as well as by type in alignFields 

joroKr21:
`alignFields` just zips fields and arguments and checks they match
(both name and type), instead of guessing based on the type.
Byname parameters are checked only by type to enable the emulation
of lazy parameters.

based on #794


joroKr21:
Fixed the 2.10 build

milessabin:

PR #809, opened 2018-01-20
Update README.md: fix output from try-shapeless.sh

flyrev:
From the try-shapeless.sh example output, remove "Loading..." after the list value is printed.


milessabin:
Well spotted! Thanks :-)

PR #816, opened 2018-02-28
Remove GenTraversable constraint from ToSizedHList

ceedubs:
The instance method needs this sort of constraint, but I don't know of a
reason that the type class itself needs this constraint.

@tpolecat I think that you originally contributed this type class. What
do you think?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -347,7 +347,11 @@ lazy val mimaSettings = mimaDefaultSettings ++ Seq(
 
       // Implicit reorderings
       exclude[DirectMissingMethodProblem]("shapeless.LowPriorityUnaryTCConstraint.hnilConstUnaryTC"),
-      exclude[ReversedMissingMethodProblem]("shapeless.LowPriorityUnaryTCConstraint.hnilUnaryTC")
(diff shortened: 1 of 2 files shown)


ceedubs:
This is failing on a binary compatibility check. I'm not sure what the current policy and/or branching strategy is for that. I'll await further guidance.

milessabin:
Unless anyone objects, I'd be happy to go with a MiMa exclusion for this one: I believe the bincompat issue would only be for other implementers of the type class trait, and I'd be surprised if there were any.

ceedubs:
@milessabin agreed, I'd be surprised if there were any. But personally I also am not in a big rush to have this changed, so depending on the anticipated shapeless timeline maybe an exception isn't warranted. I'm fine either way -- do you want me to add an exception to the PR?

milessabin:
It'd be great if you could add a MiMa exclusion to this PR :-)

ceedubs:
@milessabin I've added the mima exception.

milessabin:
Awesome ... thanks :-)

PR #823, opened 2018-03-30
Move Mapped for Id and Const to low priority

joroKr21:
Improves compile time for the common case when `F[_]` is a proper
type constructor.


milessabin:

PR #825, opened 2018-04-01
add -Xfuture option. fix warnings

xuwei-k:
procedure syntax and view bounds are deprecated


milessabin:
Many thanks for that ... it's saved me from a huge chore ... very much appreciated :-)

PR #830, opened 2018-05-06
upgrade sbt-api-mappings so build works on JDK 9

SethTisue:
this came up in the context of the Scala community build, at
https://github.com/scala/community-builds/issues/609

also threw in an sbt version bump just 'cause


milessabin:

PR #833, opened 2018-05-18
remove duplicate scala-js plugin

xuwei-k:
https://github.com/milessabin/shapeless/blob/d3da9a8aca02a6a9b2a2fcbb39a8014a9a33b4b7/project/plugins.sbt#L7-L9


milessabin:
Good catch! Many thanks ... merging :-)

PR #835, opened 2018-05-27
Define a typeclass `Perhaps` to reify implicit resolution failure

missingfaktor:
In a similar spirit as `OrElse`.

### DIFF
--- a/core/src/main/scala/shapeless/perhaps.scala
+++ b/core/src/main/scala/shapeless/perhaps.scala
@@ -1,0 +1,18 @@
+package shapeless
+
+/**
+  * Reifies implicit resolution failure as a value.
+  *
+  * Useful for expressing "If the implicit is available, do X. Else do Y.”
+  */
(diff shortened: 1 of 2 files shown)


milessabin:
I'm not opposed to this in principle, however, I think you can get pretty much the same effect by giving a (normal, language built-in) default value for your implicit argument ... does this PR cover any cases that can't be handled that way?

missingfaktor:
Nothing really. It encapsulates that pattern in a data type is all. It may be useful when used in conjunction with other similar types. For instance, `ev: A OrElse Perhaps[B]`, something that will be harder to do otherwise.

missingfaktor:
Actually never mind... In my last example, you could get the same effect with `A OrElse B = null`. 

Happy to close the PR if you don't think this adds much.

milessabin:
It's a nice idea, but TBQH I don't really think it adds enough. Thanks anyhow :-)

PR #846, opened 2018-07-14
added support for hierarchies of parameterized annotations

b4eEX:
Added support for field annotation type hierarchies, so that type boundary in `Annotations[Anno[Param], Class]()` properly respects variances.

### DIFF
--- a/core/src/main/scala/shapeless/annotation.scala
+++ b/core/src/main/scala/shapeless/annotation.scala
@@ -135,9 +135,9 @@ class AnnotationMacros(val c: whitebox.Context) extends CaseClassMacros {
     }
 
     if(isCaseClass || hasNonGenericCompanionMember("apply"))
-      args => q"${companionRef(tpe)}(..$args)"
+      args => c.typecheck(q"${companionRef(tpe)}(..$args)", c.TERMmode)
     else
-      args => q"new $tpe(..$args)"
+      args => c.typecheck(q"new $tpe(..$args)", c.TERMmode)
   }
 
   def materializeAnnotation[A: WeakTypeTag, T: WeakTypeTag]: Tree = {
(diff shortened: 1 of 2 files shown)


milessabin:
Thanks ... looks promising.

I'm afraid I don't think we can change the semantics of the existing type class (viz your changes to the tests) because people are relying on it. To move this forward we'll need a new type class and corresponding tests ... see `shapeless.ops.hlist.Unifier` vs. `shapeless.ops.hlist.SubtypeUnifer` for a precedent.

b4eEX:
I believe the only case when this PR actually changes the semantics of
`Annotation` is if somebody is using it for processing annotations with
covariant/contravariant parameters, but expects them to actually act as if
they were invariant. The doc is vague and doesn't explain that this is the
indended behavior, so shouldn't this be considered a bug or, at least,
undocumented behavior?

b4eEX:
@milessabin I guess I could effectively extract/make a new type class with those changes, but that would mean modifying the existing docs to specifically state that the current behavior is the spec (now it's just `annotations of type A`, which could be interpreted as both `=:=` and `<:<`). Another option would be to maybe pull the type equality check up into an implicit with a default value. The difference at this point is really just `s/=:=/<:</g`, so spinning up an entire new type seems a bit heavy-handed to me. What do you think?

joroKr21:
How often is subtyping used for annotations? I have no clue.

joroKr21:
Closing for inactivity. As already mentioned, feel free to reopen with a new typeclass instead of modifying the existing one.

joroKr21:
@milessabin actually looking at #715 - it's very similar to this, but seems like you were inclined to accept it. I think we should have a consistent stance on variance in all typeclasses. We should either:
  1. Accept making existing typeclasses covariant (I don't think there's any case for contravariant)
  2. Accept only new typeclasses which are covariant
  3. Reject variance altogether as out of scope for shapeless

Either way I would not have time and motivation to go though all existing typeclasses and try to make them covariant or add covariant versions, so the question is more about what to do with such contributions.

milessabin:
I think any drastic variance related changes should wait for shapeless 3. I'll close #715.

PR #854, opened 2018-08-23
Remove spurious word from comment in klist example

antbell:
Fix a minor documentation issue in klist example.


milessabin:
Well spotted! Many thanks.

PR #856, opened 2018-08-23
Fix scaladoc for Cached

DmytroMitin:


milessabin:
Well spotted! Many thanks :-)

PR #859, opened 2018-09-05
Use Strict in hlist.LeftFolder and hlist.RightFolder to avoid diverging implicit expansion

DmytroMitin:
Details are here: https://stackoverflow.com/questions/52175141/diverging-implicit-expansion-error-on-foldleft-hlist/

`RightFolder` works properly without `Lazy`.


milessabin:
Do you have a test which fails before this PR and succeeds afterwards?

Why have you enabled `scala.language.reflectiveCalls`?

DmytroMitin:
@milessabin 
Before PR the following code doesn't compile:
```
object poly extends Poly2 {
  implicit def wrapMap[T]: Case.Aux[T, Int, Map[List[Int], T]] = at[T, Int]((end, t) => Map(List(t) -> end)) 
}

implicitly[LeftFolder.Aux[Int :: Int :: HNil, Int, poly.type, Map[List[Int], Map[List[Int], Int]]]]
```

After PR it does.

Without `scala.language.reflectiveCalls` test `poly.scala` doesn't compile.

https://travis-ci.org/milessabin/shapeless/jobs/424845742

```
[error] /home/travis/build/milessabin/shapeless/core/src/test/scala/shapeless/poly.scala:427:56: reflective access of structural type member method allCases should be enabled
[error] by making the implicit value scala.language.reflectiveCalls visible.
[error] This can be achieved by adding the import clause 'import scala.language.reflectiveCalls'
[error] or by setting the compiler option -language:reflectiveCalls.
[error] See the Scaladoc for value scala.language.reflectiveCalls for a discussion
[error] why the feature should be explicitly enabled.
[error]     val r = (10 :: "hello" :: true :: HNil).foldLeft(0)(myPoly)
[error]                                                        ^
```

joroKr21:
It works in Scala 2.13. `Strict` might be a better choice here.

joroKr21:
Also `RightFolder` might have the same problem and this needs a test.

core/src/test/scala/shapeless/poly.scala
```diff
@@ -16,6 +16,8 @@
 
 package shapeless
 
+import scala.language.reflectiveCalls
+
```
joroKr21:
```suggestion
```

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -429,10 +429,10 @@ object hlist {
       }
 
     implicit def hlistLeftFolder[H, T <: HList, In, HF, OutH, FtOut]
-      (implicit f : Case2.Aux[HF, In, H, OutH], ft : LeftFolder.Aux[T, OutH, HF, FtOut]): Aux[H :: T, In, HF, FtOut] =
+      (implicit f : Case2.Aux[HF, In, H, OutH], ft : Lazy[LeftFolder.Aux[T, OutH, HF, FtOut]]): Aux[H :: T, In, HF, FtOut] =
```
joroKr21:
```suggestion
      (implicit f : Case2.Aux[HF, In, H, OutH], ft : Strict[LeftFolder.Aux[T, OutH, HF, FtOut]]): Aux[H :: T, In, HF, FtOut] =
```

milessabin:
Apparently there really is some unexpected reflective access there ...

joroKr21:
> Apparently there really is some unexpected reflective access there ...

I have no clue why this is happening

milessabin:
Missing explicit refinement somewhere maybe?

joroKr21:
@milessabin I figured out the cause of the reflective calls warning - `PolyNBuilders` were missing type annotations. Check the last commit.

milessabin:
> I figured out the cause of the reflective calls warning

Good catch!

Issue #860, opened 2018-09-10
Sometimes derivation fails if you use -yno-predef


MateuszKubuszok:
I am using custom Predef. I noticed the issue when I started to use more type-class derivation in my project (Kittens).

I have a hard time creating a reproducible test case, but I noticed that one of things that breaks (and that I can fix) is `classOf` - if is something generated by the compiler, so I cannot simply add `def classOf[T] = scala.Predef.classOf[T]` in my Predef (it would probably work out as a macro though), but macros start to work again when I add `import scala.Predef.classOf`.

I believe the cause might be macros using `classOf` and assuming that it is always there e.g. [here](https://github.com/milessabin/shapeless/blob/6b54d3c33c83cc730c472a91c9d874ac925f8147/core/src/main/scala/shapeless/typeable.scala#L461).

Would changing `classOf` to `_root_.scala.Predef.classOf` be an acceptable solution/PR?

milessabin:
Yup, that's a hygiene problem ... good catch!

joroKr21:
Fixed in #861 :tada:

PR #861, opened 2018-09-10
Make Typeable macros not dependent on default Predef presence

MateuszKubuszok:
Typeable macros used classOf directly - it breaks if one uses settings like -Yno-predef or similar (https://github.com/milessabin/shapeless/issues/860). This PR should fix it.


joroKr21:
@MateuszKubuszok hope you don't mind I rebased the PR myself.

Thank you for the fix! :+1:

MateuszKubuszok:
No problem, thanks! I'm on vacation so I don't look at github very often.

PR #867, opened 2018-10-04
Support for Annotation[Option[A], T]

guersam:
Currently, `Annotation` just aborts compilation when type `T` is not
annotated with `A`. Although it's possible to allow non-annotated case
by implicit priorities, It would be better to use `Option[A]` to avoid
duplication in complex generic derivations.

This PR additionaly allows `Annotation[Option[A], T]` by returning
`Option[A]` in runtime, keeping current behavior of `Annotation[A ,T]`
the same.


core/src/main/scala/shapeless/annotation.scala
```diff
@@ -45,15 +45,19 @@ trait Annotation[A, T] extends Serializable {
   def apply(): A
 }
 
-object Annotation {
+object Annotation extends Annotation0 {
+  implicit def materialize[A, T]: Annotation[A, T] = macro AnnotationMacros.materializeAnnotationRequired[A, T]
```
guersam: I want to prioritize the default behavior not to decrease compilation speed, but it might be just okay to keep them in a single object.

core/src/main/scala/shapeless/annotation.scala
```diff
@@ -154,10 +165,14 @@ class AnnotationMacros(val c: whitebox.Context) extends CaseClassMacros {
...
-      case Some(annTree) =>
+    (annTreeOpt, optional) match {
+      case (Some(annTree), true) =>
+        q"_root_.shapeless.Annotation.mkAnnotation[${appliedType(optionTpe, annTpe)}, $tpe](_root_.scala.Some($annTree))"
+      case (Some(annTree), false) =>
         q"_root_.shapeless.Annotation.mkAnnotation[$annTpe, $tpe]($annTree)"
-      case None =>
+      case (None, true) =>
+        q"_root_.shapeless.Annotation.mkAnnotation[${appliedType(optionTpe, annTpe)}, $tpe](_root_.scala.None)"
+      case (None, false) =>
```
guersam: Sorry about potential boolean blindness here. Any suggestion to improve?

guersam: Nevermind, figured it out :)

guersam:
Oops, just realized that I was running `sbt> coreJVM/testOnly *annotation*`, not `*Annotation*`.

guersam:
What does that `codecov/path` failure mean?

milessabin:
@guersam codecov is very flaky ... TBQH I think it's more noise than signal. Don't worry about it.

core/src/main/scala/shapeless/annotation.scala
```diff
@@ -45,15 +45,19 @@ trait Annotation[A, T] extends Serializable {
   def apply(): A
 }
 
-object Annotation {
+object Annotation extends Annotation0 {
+  implicit def materialize[A, T]: Annotation[A, T] = macro AnnotationMacros.materializeAnnotationRequired[A, T]
+}
+
+private[shapeless] trait Annotation0 {
```
guersam: Does it affect the binary compatibility? I'm not familiar with JVM internals :(

core/src/main/scala/shapeless/annotation.scala
joroKr21: I don't think it matters for compilation speed.

PR #869, opened 2018-10-18
simplify foldLeft to forall where possible

teresy:
The `foldLeft(true)(_ && _)` calls can be simplified to `forall(identity)`


milessabin:
I take your point, but the test here is really supposed to be demonstrating the equivalence of `foldMap` to a `map` followed by a `foldLeft` ... I think rewriting in the way you're proposing makes that less clear.

teresy:
Ah, thanks for pointing that out - that does makes sense.

PR #874, opened 2019-01-11
Add inverse method to Prepend

antbell:
Add an `inverse` method to `Prepend`, splitting the final HList back into a tuple `(prefix, suffix)`.

This allows composition of objects having HLists in contravariant position (e.g. as method arguments).

Example:

```scala
trait HListShow[A <: HList] { top =>
  def show(l: A): String

  def ~[B <: HList](other: HListShow[B])(implicit p: Prepend[A, B]): HListShow[p.Out] = new HListShow[p.Out] {
    def show(l: p.Out): String = {
      val (l1, l2) = p.inverse(l)
      top.show(l1) + "~" + other.show(l2)
    }
  }
}
```

Something like this was suggested by @travisbrown in reply to [this SO question](https://stackoverflow.com/questions/32366527/splitting-an-hlist-that-was-concatenated-using-prependa-b).

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -2016,7 +2016,10 @@ object hlist {
    *
    * @author Miles Sabin
    */
-  trait Prepend[P <: HList, S <: HList] extends DepFn2[P, S] with Serializable { type Out <: HList }
+  trait Prepend[P <: HList, S <: HList] extends DepFn2[P, S] with Serializable {
+    type Out <: HList
+    def inverse(p: Out): (P, S)
+  }
 
   trait LowestPriorityPrepend {
     type Aux[P <: HList, S <: HList, Out0 <: HList] = Prepend[P, S] { type Out = Out0 }
@@ -2026,6 +2029,11 @@ object hlist {
       new Prepend[PH :: PT, S] {
         type Out = PH :: PtOut
         def apply(prefix : PH :: PT, suffix : S): Out = prefix.head :: pt(prefix.tail, suffix)
+
+        def inverse(p: PH :: PtOut): (PH :: PT, S) = {
+          val e = pt.inverse(p.tail)
+          (p.head :: e._1, e._2)
+        }
       }
   }
 
@@ -2036,20 +2044,24 @@ object hlist {
      */
     override type Aux[P <: HList, S <: HList, Out0 <: HList] = Prepend[P, S] { type Out = Out0 }
 
-    implicit def hnilPrepend0[P <: HList, S <: HNil]: Aux[P, S, P] =
+    implicit def hnilPrepend0[P <: HList, S >: HNil.type <: HNil]: Aux[P, S, P] =
       new Prepend[P, S] {
         type Out = P
         def apply(prefix : P, suffix : S): P = prefix
+
+        def inverse(p: P): (P, S) = (p, HNil)
       }
   }
 
   object Prepend extends LowPriorityPrepend {
     def apply[P <: HList, S <: HList](implicit prepend: Prepend[P, S]): Aux[P, S, prepend.Out] = prepend
 
-    implicit def hnilPrepend1[P <: HNil, S <: HList]: Aux[P, S, S] =
+    implicit def hnilPrepend1[P >: HNil.type <: HNil, S <: HList]: Aux[P, S, S] =
       new Prepend[P, S] {
         type Out = S
         def apply(prefix : P, suffix : S): S = suffix
+
+        def inverse(p: S): (P, S) = (HNil, p)
       }
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -2036,20 +2044,24 @@ object hlist {
      */
     override type Aux[P <: HList, S <: HList, Out0 <: HList] = Prepend[P, S] { type Out = Out0 }
 
-    implicit def hnilPrepend0[P <: HList, S <: HNil]: Aux[P, S, P] =
+    implicit def hnilPrepend0[P <: HList, S <: HNil](implicit n: HNil.type <:< S): Aux[P, S, P] =
```
joroKr21: Better to just remove the `S` type parameter and use `HNil` (which is also binary compatible).
I wonder if there was a reason to use `S` in the first place.

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -2036,20 +2044,24 @@ object hlist {
...
+
+        def inverse(p: P): (P, S) = (p, n(HNil))
       }
   }
 
   object Prepend extends LowPriorityPrepend {
     def apply[P <: HList, S <: HList](implicit prepend: Prepend[P, S]): Aux[P, S, prepend.Out] = prepend
 
-    implicit def hnilPrepend1[P <: HNil, S <: HList]: Aux[P, S, S] =
+    implicit def hnilPrepend1[P <: HNil, S <: HList](implicit n: HNil.type <:< P): Aux[P, S, S] =
```
joroKr21: Same comment about the type parameter.
Let's at least try it out and see if all tests pass.

core/src/main/scala/shapeless/ops/hlists.scala
antbell: I have given it a try, but `HNil` doesn't seem to cover some cases where `HNil.type` is needed/expected.

```
shapeless/core/src/test/scala/shapeless/hlist.scala:427:53: could not find implicit value for parameter prepend: shapeless.ops.hlist.Prepend[L,shapeless.HNil.type]
```

I think the `S` type parameter is used to cover both `HNil` and `HNil.type` types in one go.
The alternative is writing separate implicits for `HNil` and `HNil.type`, which makes tests pass, but involves some additional code duplication.

core/src/main/scala/shapeless/ops/hlists.scala
antbell: Similar problem as above:
```
shapeless/core/src/test/scala/shapeless/hlist.scala:426:55: could not find implicit value for parameter prepend: shapeless.ops.hlist.Prepend[shapeless.HNil.type,L]
```

core/src/main/scala/shapeless/ops/hlists.scala
joroKr21: Ok a quick search for `<: HNil` reveals that it's used 25 times. But definitely not every time which is quite inconsistent. How about doing the ugly thing here and just use `asInstanceOf`?

In hindsight I think it would have been better if didn't expose `object HNil` but just `val HNil: HNil = new HNil {}`. @milessabin WDYT, is this something we ought to do in 2.4?

milessabin: I'm not enthusiastic about using `asInstanceOf` unless there's absolutely no alternative.

> is this something we ought to do in 2.4

I'm in two minds about this. On the one hand I think it'd be good to experiment ... I dimly remember there might have been a good reason for it being the way it is, but maybe it's historical, or maybe it was a bad choice that I wouldn't make now.

On the other, it's been this way since the beginning and I'm sure there will be client code that relies on the more precise type. I'm not sure I want to break source compatibility in that way for 2.4. 2.5 _will_ be breaking souce compatibility, so maybe it'd be better to leave it until then.

Also bear in mind that 3.0 will very likely be deprecating `HList`s in favour of `Unit` and tuples, so it might not be worth changing it at all in the 2.x series.

joroKr21: > Also bear in mind that 3.0 will very likely be deprecating HLists in favour of Unit and tuples, so it might not be worth changing it at all in the 2.x series.

Alright, makes sense. Let's not touch it then.

@antbell sorry it took me so long to notice this but we should be able to say `S >: HNil.type <: HNil` right?

antbell: > @antbell sorry it took me so long to notice this but we should be able to say `S >: HNil.type <: HNil` right?

That works, but I think it's not necessary, since `HNil` is a sealed trait, only implemented by the `HNil` object.

joroKr21: Right, I meant instead of the implicit `<:<` evidence

antbell: That works indeed and is more elegant. Nice! I have pushed the change.

joroKr21:
Hi @antbell, one last thing before merge, do you mind squashing to one commit?

antbell:
> Hi @antbell, one last thing before merge, do you mind squashing to one commit?

Sure, I have squashed the commits and force-pushed.

joroKr21:
Merged. Thank you! :+1:

PR #878, opened 2019-02-21
Fix bug in RotateLeft/RotateRight for Poly with more than 1 Case

DmytroMitin:
Type parameter `RL` should be inferred before `L`, otherwise RotateLeft/RotateRight is not resolved for Poly with more than 1 Case.

```
[error] /home/travis/build/DmytroMitin/shapeless/core/src/test/scala/shapeless/poly.scala:312:17: could not find implicit value for parameter cse: shapeless.poly.Case[sdi.type,String :: Double :: Int :: shapeless.HNil]
[error]     val r2 = sdi("foo", 2.0, 1)
[error]                 ^
[error] /home/travis/build/DmytroMitin/shapeless/core/src/test/scala/shapeless/poly.scala:315:18: could not find implicit value for parameter cse: shapeless.poly.Case[sdi.type,Char :: Boolean :: Long :: shapeless.HNil]
[error]     val r21 = sdi('f', true, 1L)
[error]                  ^
[error] /home/travis/build/DmytroMitin/shapeless/core/src/test/scala/shapeless/poly.scala:320:17: could not find implicit value for parameter cse: shapeless.poly.Case[dis.type,Double :: Int :: String :: shapeless.HNil]
[error]     val r3 = dis(2.0, 1, "foo")
[error]                 ^
[error] three errors found
```
https://travis-ci.org/DmytroMitin/shapeless/jobs/496497621

### DIFF
--- a/core/src/main/scala/shapeless/poly.scala
+++ b/core/src/main/scala/shapeless/poly.scala
@@ -88,7 +88,7 @@ object PolyDefns extends Cases {
 
   object RotateLeft {
     implicit def rotateLeftCase[C, P <: Poly, N <: Nat, L <: HList, LOut, RL <: HList]
-      (implicit unpack: Unpack2[C, RotateLeft, P, N], cP: Case.Aux[P, L, LOut], rotateRight: hl.RotateRight.Aux[RL, N, L])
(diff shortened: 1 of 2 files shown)


milessabin:
LGTM! Many thank ... merging :-)

PR #879, opened 2019-02-22
Fix bug in Witness for TypeRef singleton type (error "type argument A.type is not a singleton type")

DmytroMitin:
Add one more matching case to `SingletonTypeMacros#extractSingletonValue(..)` when a singleton type (namely, type of an object) is a `TypeRef`.
Without this fix the following code doesn't compile
```
trait B
case object A extends B

trait B1 {
  type T <: B
  def getT(implicit w: Witness.Aux[T]): T = w.value
}
case class A1() extends B1 {
  type T = A.type
}

assertTypedEquals[A.type](A1().getT, A)
```

```
Information: shapeless.Witness.apply is not a valid implicit value for shapeless.Witness{type T = A.type} because:
hasMatchingSymbol reported error: Type argument A.type is not a singleton type
    assert(A1().getT == A)

Error: could not find implicit value for parameter w: shapeless.Witness{type T = A.type}
    assert(A1().getT == A)

Error: not enough arguments for method getT: (implicit w: shapeless.Witness{type T = A.type})A.type.
Unspecified value parameter w.
    assert(A1().getT == A)
```

Motivated by the discussion https://stackoverflow.com/questions/54786083/get-instance-of-singleton-type-in-scala

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,11 +118,11 @@ resolvers ++= Seq(
 [ci]: https://travis-ci.org/milessabin/shapeless
 
 
-Builds are available for Scala 2.10.x, 2.11.x and for 2.12.x. The main line of development for
-shapeless 2.3.3 is Scala 2.12.4. Scala 2.10.x is supported via the macro paradise compiler plugin.
+Builds are available for Scala 2.10.x, 2.11.x, 2.12.x and for 2.13.0-M5. The main line of development for
+shapeless 2.3.3 is Scala 2.12.8. Scala 2.10.x is supported via the macro paradise compiler plugin.
 
 ```scala
-scalaVersion := "2.12.4"
+scalaVersion := "2.12.8"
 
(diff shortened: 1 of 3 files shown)


DmytroMitin:
This change would be consistent with that similar code works properly:

```
  trait B
  case object A extends B

  trait B1 {
    type T <: B
    def getT(implicit w: Witness.Aux[T]): T = w.value
  }
  case object A1 extends B1 {
    type T = A.type
  }

  assertTypedEquals[A.type](A1.getT, A)
```

and 

```
  trait B
  case object A extends B

  trait B1[T <: B] {
    def getT(implicit w: Witness.Aux[T]): T = w.value
  }
  case class A1() extends B1[A.type]

  assertTypedEquals[A.type](A1().getT, A)
```

milessabin:

PR #885, opened 2019-04-03
Add SelectAny typeclass to shapeless.ops.hlist

akiomik:
Hi, I added `SelectAny` to `HList`.
I think this is useful for the following case:

```scala
final case object TeamA
final case object TeamB
final case object Admin
final case object Operator

type AllowedRoles = TeamA.type :: Admin.type :: Operator.type :: HNil
def access[A <: HList](roles: A)(implicit any: SelectAny[AllowedRoles, A]) = true

access(HNil)                       // compile fails!
access(TeamA :: HNil)              // true
access(TeamB :: HNil)              // compile fails!
access(Admin :: HNil)              // true
access(Operator :: HNil)           // true
access(TeamA :: Operator :: HNil)  // true
access(TeamB :: Operator :: HNil)  // true
access(TeamB :: Admin :: HNil)     // true
```

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -955,7 +955,36 @@ object hlist {
       }
   }
 
+  /**
+   * Type class supporting access to the first element of `S` in this `HList`.
+   * Available only if this `HList` contains at least one element of `S`.
+   *
+   * @author Akiomi Kamakura
+   */
+  @implicitNotFound("Implicit not found: shapeless.Ops.SelectFirst[${L}, ${S}]. You requested an element of ${S}, but there is none in the HList ${L}.")
+  trait SelectFirst[L <: HList, S <: HList] extends DepFn1[L] with Serializable { type Out }
 
+  object SelectFirst {
+    def apply[L <: HList, S <: HList](implicit sel: SelectFirst[L, S]): Aux[L, S, sel.Out] = sel
+
+    type Aux[L <: HList, S <: HList, Out0] = SelectFirst[L, S] { type Out = Out0 }
+
+    implicit def selectAny[L <: HList, H, S <: HList](
+      implicit sel: Selector[L, H]
+    ): Aux[L, H :: S, H] =
+      new SelectFirst[L, H :: S] {
+        type Out = H
+        def apply(l: L): Out = sel(l)
+      }
+
+    implicit def recurse[L <: HList, H, S <: HList, SelOut](
+      implicit sel: Aux[L, S, SelOut]
+    ): Aux[L, H :: S, SelOut] =
+      new SelectFirst[L, H :: S] {
+        type Out = SelOut
+        def apply(l: L): Out = sel(l)
+      }
+  }
 
   /**
    * Type class supporting partitioning this `HList` into those elements of type `U` and the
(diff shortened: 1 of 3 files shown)


milessabin:

Did you consider implementing this directly rather than in terms of `Selector`?

akiomik:
Yes, but I came to the conclusion that using `Selector` is a simple way because `Selector` is equivalent to `contains` in the following code.

```scala
def any[A](a: List[A], b: List[A]): A = {
  def rec(a: List[A]): A = a match {
    case h :: t if b contains h => h
    case _ :: t                 => rec(t)
  }
  rec(a)
}
```

milessabin:
It's mainly compile times I'm concerned about.

When resolving `SelectAny[L, S]` we'll be using a `Selector[L, s]` which scales at best linearly in the size of `L`. In your examples `L` is bigger than `S` ... if you think that will be typical there might be some benefit in reworking this a little. Of course if you expect the sizes of the arguments to be fairly small then this isn't much of an issue (at least, no more so than the compile times are an issue for shapeless generally).

Incidentally, I think `SelectFirst` might be a better name for this type class.

joroKr21:
Do you think a direct implementation would be more efficient than `O(n * m)`? I don't see how.

joroKr21:
Given that `SelectAll` also relies on `Selector` I would say it's ok here as well.
Agree that probably `SelectFirst` is a better name.
Although we lose the symmetry with `SelectAll`, but it's better to be precise.

akiomik:
Really sorry for the deley. I renamed `SelectAny` to `SelectFirst` for now.

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -955,7 +955,36 @@ object hlist {
       }
   }
 
+  /**
+   * Type class supporting access to the first element of this `HList` in `S`.
```
joroKr21: I thought about this some more. Note that there are two options:
  1. Select the first element of `L` which is present in `S`
  2. Select the first element of `S` which is present in `L`

The difference is subtle. I read this comment as describing 1. but the implementation as 2.
We should take a step back to make sure that:
  * We implement the semantics that make more sense. Not sure which one, but the user could easily switch by swapping the arguments.
  * The comment and the implementation align.

Do I interpret this correctly?

core/src/test/scala/shapeless/hlist.scala
```diff
@@ -1424,6 +1424,32 @@ class HListTests {
...
+    assertEquals(2.0, sd, Double.MinPositiveValue)
+
+    val sib = sl.selectFirst[Int::Boolean::HNil]
+    assertTypedEquals[Int](1, sib)
+
+    val sulb = sl.selectFirst[Unit::Long::Boolean::HNil]
+    assertTypedEquals[Boolean](true, sulb)
+
+    val ssbi = sl.selectFirst[String::Boolean::Int::HNil]
+    assertTypedEquals[String]("foo", ssbi)
```
joroKr21: Alright this is actually the relevant test - we get `String`, although `Boolean` comes first in `sl`

core/src/main/scala/shapeless/syntax/hlists.scala
```diff
@@ -137,6 +137,12 @@ final class HListOps[L <: HList](l : L) extends Serializable {
 
   def selectRange(a : Nat, b : Nat)(implicit sel: SelectRange[L,a.N,b.N]): sel.Out = sel(l)
 
+  /**
+   * Returns the first element of this `HList` in `S`.
```
joroKr21: Same issue here about the comment. I think we need to swap:
> Returns the first element of `S` in this `HList`

core/src/main/scala/shapeless/ops/hlists.scala
joroKr21: Considering the `selectFirst` syntax I think the implementation (option 2.) makes sense and we need to tweak the comment.

PR #890, opened 2019-04-21
Make OrElse Serializable

joroKr21:
Ref typelevel/kittens#130


milessabin:

PR #893, opened 2019-05-10
Ensure default values are only run when Default.apply is called

alexarchambault:
Before this PR, getting a `Default` instance, like `Default[CC]`, was calling the default values straightaway. For example, the following would throw:
```scala
case class CC(n: Int = ???)
Default[CC] // throws because of ???, even though we didn't ask for the actual default values yet
```

With the changes here, default values are only called when `Default.apply` is called. Only the following would throw:
```scala
case class CC(n: Int = ???)
val default = Default[CC] // does not throw
default() // trying to get the default values, throws because of ???
```

### DIFF
--- a/core/src/main/scala/shapeless/default.scala
+++ b/core/src/main/scala/shapeless/default.scala
@@ -39,7 +39,11 @@ trait Default[T] extends DepFn0 with Serializable {
 object Default {
   def apply[T](implicit default: Default[T]): Aux[T, default.Out] = default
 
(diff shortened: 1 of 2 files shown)


milessabin:
Great stuff! Thanks :-)

Would you mind splitting the scala-native stuff off to a separate PR?

alexarchambault:
> Would you mind splitting the scala-native stuff off to a separate PR?

Sure

milessabin:

PR #900, opened 2019-06-11
Loosen `Typeable` restriction

### LINKED ISSUE #812: Typeable failing in 2.3.3 for code that works in 2.3.2
(Tested under scala 2.11)

The following code works as expected with shapeless 2.3.2
````scala
object X {
  case class A()
  case class B(a: A) {
    private[this] val aa = a
  }
}
object Test {
  shapeless.Typeable[X.B]
}
````

But when using shapeless 2.3.3 I get:

    <console>:13: error: No default Typeable for parametrized type X.B
             shapeless.Typeable[X.B]
                               ^

### PR DESCRIPTION
**Pull request by jeremyrsmith**
Fixes #812

Don't reject materializing a `Typeable` instance due to the presence of a non-case-accessor value, unless its type is abstract. The abstract type is what makes it unsafe, so we should only reject if that's the case.

### DIFF
--- a/core/src/main/scala/shapeless/typeable.scala
+++ b/core/src/main/scala/shapeless/typeable.scala
@@ -466,13 +466,21 @@ class TypeableMacros(val c: blackbox.Context) extends SingletonTypeUtils {
   private def mkCaseClassTypeable(tpe: Type): Tree = {
     // an unsafe accessor is one that isn't a case class accessor but has an abstract type.
     def isUnsafeAccessor(sym: TermSymbol): Boolean = {
+
       if (sym.isCaseAccessor) {
         false
-      } else if (sym.typeSignature.typeSymbol.isAbstract) {
-        sym.isVal ||
-        sym.isVar ||
-        (sym.isParamAccessor && !(sym.accessed.isTerm && sym.accessed.asTerm.isCaseAccessor))
-      } else false
+      } else {
+        val symType = sym.typeSignature.typeSymbol
+        val isAbstract =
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jeremyrsmith** [comment]
Interestingly `isAbstract` seems not to work in Scala 2.10 (hence the failure). Could be a macro-compat issue worth looking at (if Scala 2.10 still matters)

**Comment by jeremyrsmith** [comment]
To clarify the issue, this check was added in #575:

```scala
val nonCaseAccessor = tpe.decls.exists {
          case sym: TermSymbol if !sym.isCaseAccessor && (sym.isVal || sym.isVar ||
              (sym.isParamAccessor && !(sym.accessed.isTerm && sym.accessed.asTerm.isCaseAccessor))) => true
          case _ => false
        }
        if (nonCaseAccessor) {
          // there is a symbol, which is not a case accessor but a val,
          // var or param, so we won't be able to type check it safely:
          c.abort(c.enclosingPosition, s"No default Typeable for parametrized type $tpe")
        }
```

What this amounts to is rejecting any case class that defines a `val` in its body. For example, I won't be able to get a `Typeable` instance for:

```scala
case class Foo(a: Int) {
  val foo = 22
}
```

This is a bit too strict (and leads to bug #812). Instead, there should also be a check as to whether the value's type is parametric. So this PR adds a check that this is the case.

It's possible that this is *still* even too strict, as in here:

```scala
case class Foo[A](a: A) {
  val aa = a
}
```

Here, the presence of `aa` does not render the whole thing un-`Typeable`, because `aa` depends only on `a` which will be checked during a cast. This will still fail to get an instance after this PR. But, it should be a good improvement.

milessabin:
Yup, this LGTM! You've won your t-shirt!

PR #901, opened 2019-06-16
added an json ADT example, which uses the most recent dotty syntax

borissmidt:
Hey miles,

As i promised i tried to implement a simple example for Json and this is what i came up with.
It was much easier to implement then the old API.

The hardest part for me was getting the polymorphic method definition correctly, 
because the compiler was not very helpful in its error message.
```
[error]     |Found:    Object with PolyFunction {...}
[error]     |Required: PolyFunction{apply: [t](x$1: Acc, x$2: shapeless.Json[t]): (Acc, Option[t])}
```

I however couldn't get the deriving for OptionInt working, i suspect this is the coproduct part?

I will clean up the PR (typos) later this week.

kind regards, Boris Smidt

### DIFF
--- a/core/src/test/scala/shapeless/deriving.scala
+++ b/core/src/test/scala/shapeless/deriving.scala
@@ -15,7 +15,6 @@
  */
 
 package shapeless
-
 import org.junit.Test
 
 import scala.compiletime._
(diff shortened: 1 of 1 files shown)


milessabin:
This looks like great stuff! Thanks for trying this out :-)

milessabin:
@borissmidt things have moved on a little since June ... do you want to see if you can get this over the line?

I think it might be best to split this out as a separate test.

borissmidt:
Oke i will have a look at it on Monday when i have the time, the main issue
for me was syntax highlighting and linting for dotty in the tooling

borissmidt:
I think it is better to wait till  dotty stabilizes so there is better tooling around it. But it was nice trying it out for once to see how easy it will become.

milessabin:
Thanks for giving it a try. Looking forward to seeing what you come up with later.

Issue #922, opened 2019-09-09
Add support for type annotations


pgrandjean:
According to the scala [documentation](https://www.scala-lang.org/files/archive/spec/2.12/11-annotations.html), annotations may apply to classes, variables, types and expressions. Currently, `Annotations` allows to extract variable annotations from case classes, such as:
```scala
case class Foo(
  @Bar field1: String
)
```

The aim of this request is to add support for type annotations, such as:
```scala
case class Foo(
  field1: String @Bar
)
```

PR #925, opened 2019-09-13
Add support for type annotations (issue #922)

pgrandjean:


pgrandjean:
working to fix macros for scala 2.10

pgrandjean:
Build fails with 2.13.1, but it looks like the error is related to SBT.

milessabin:
We're waiting on Scala.js to propagate ... won't be long.

milessabin:
Actually, it's scoverage: https://github.com/scoverage/sbt-scoverage/issues/295.

milessabin:
You should be good if you rebase on master.

pgrandjean:
Rebased.

core/src/main/scala_2.10/shapeless/annotationmacros.scala
```diff
@@ -0,0 +1,65 @@
...
+        annots.collectFirst {
+          case ann if ann.tree.tpe =:= annTpe => construct0(ann.tree.children.tail)
+        }
+      case _ => None
+    }
+  }
+}
+
+/*
+ * Workaround for Scala 2.10, which needs an object.
```
milessabin: macrocompat should allow you to keep to the same bundle style as 2.11+ ... can you say what problems you were seeing?

pgrandjean: If I remove the companion object, I get compilation errors saying "\<macro> is not a member of object shapeless.AnnotationMacros":

```
[error] /Development/shapeless/core/src/main/scala/shapeless/annotation.scala:56:77: value materializeAnnotation is not a member of object shapeless.AnnotationMacros
[error]   implicit def materialize[A, T]: Annotation[A, T] = macro AnnotationMacros.materializeAnnotation[A, T]
[error]                                                                             ^
[error] /Development/shapeless/core/src/main/scala/shapeless/annotation.scala:117:89: value materializeVariableAnnotations is not a member of object shapeless.AnnotationMacros
[error]   implicit def materialize[A, T, Out <: HList]: Aux[A, T, Out] = macro AnnotationMacros.materializeVariableAnnotations[A, T, Out]
[error]                                                                                         ^
[error] /Development/shapeless/core/src/main/scala/shapeless/annotation.scala:171:89: value materializeTypeAnnotations is not a member of object shapeless.AnnotationMacros
[error]   implicit def materialize[A, T, Out <: HList]: Aux[A, T, Out] = macro AnnotationMacros.materializeTypeAnnotations[A, T, Out]
```
That made we believe that the problem was the same for LazyMacrosRef, which is an object in 2.10 and a class in 2.11+.

milessabin: The `Lazy` macros are "special" in all sorts of strange ways.

A closer comparison would be `AnnotationMacros` [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/annotation.scala#L120-L213). If you follow that model I think it's fairly likely you can do this without needed any version specific source.

pgrandjean: I will try to get closer to the original code. But then there is the extraction of `Types.AnnotatedType` which is slightly different between versions:
* 2.10: https://github.com/scala/scala/blob/v2.10.7/src/reflect/scala/reflect/api/Types.scala#L799
* 2.11: https://github.com/scala/scala/blob/v2.11.12/src/reflect/scala/reflect/api/Types.scala#L899

How to go around this?

pgrandjean:
Got rid of the companion object in 2.10 although there is still a tiny bit that is version specific. I don't see how to get around `Types.AnnotatedType`, so that part is now isolated in a trait, AnnotationsHelper, defined in folders scala_2.10 and scala_2.11+.

milessabin:
Can you use something like `case a: AnnotatedType => a.annots` cross versions?

pgrandjean:
Thanks for the hint!

pgrandjean:
Got rid of the version specific implementation. Now everything is in the common bundle.

milessabin:
Sorry for the hold up ...

That's great stuff. Many thanks ... merging :-)

joroKr21:
I don't think this was released

PR #928, opened 2019-09-19
Avoid use of mutable state in macro classloader

retronym:
Recent versions of Scala have an experimental option
(`-Ycache-macro-classloader`) to reuse a macro classloader across
multiple (possibly concurrent) compilations. It is motivated by
performance to reduce overhead of classloading and of the macro
implementation running without time to JIT.

However, this requires macro implenentations to avoid races and leaks
with any classloader-global state.

This commit introduces a generic facility for managing macro
state that should have the same lifecycle as a `Global#Run` and
uses it in the `Cached` macro.

### DIFF
--- a/core/src/main/scala/shapeless/MacroState.scala
+++ b/core/src/main/scala/shapeless/MacroState.scala
@@ -1,0 +1,48 @@
+package shapeless
+
+import scala.reflect.{ClassTag, classTag}
+import scala.reflect.macros.Universe
+
+
+private[shapeless] object MacroState {
+  private class MacroStateAttachment(val state: collection.mutable.HashMap[Class[_], Any])
+
+  /**
+   * Associates some state with a global.Run. Preferable to using static state in the macro
+   * classloader (e.g vars in top level objects) so as to avoid race conditions under `-Ycache-macro-classloader`.
+   *
+   * @tparam T the type of the state. The class tag for this type will be used a a map key, so each user of
+   *           this facility should use a dedicated and distinct class to wrap the state.
+   * @param u The reflection universe, typically obtained from `context.universe` in a macro implementation.
+   * @param factory Factory to create the state.
+   */
+  def apply[T: ClassTag](u: Universe, factory: => T): T = {
+    // Cast needed for access to perRunCaches and convenient access to attachments
+    val g = u.asInstanceOf[scala.reflect.internal.SymbolTable]
+
+    // Sneakily use a symbol attachment on a well-known symbol to hold our map of macro states.
+    // Compiler plugins would typically use `val someState = global.perRunCaches.newMap` in a `Component` that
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/MacroState.scala
```diff
@@ -0,0 +1,48 @@
+package shapeless
+
+import scala.reflect.{ClassTag, classTag}
+import scala.reflect.macros.Universe
+
+
+private[shapeless] object MacroState {
```
retronym: Perhaps this utility could live in macro-compat.

retronym:
Fixed up the CI failures with the latest push.

core/src/main/scala/shapeless/MacroState.scala
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Associates some state with a global.Run. Preferable to using static state in the macro
+   * classloader (e.g vars in top level objects) so as to avoid race conditions under `-Ycache-macro-classloader`.
+   *
+   * @tparam T the type of the state. The class tag for this type will be used a a map key, so each user of
+   *           this facility should use a dedicated and distinct class to wrap the state.
+   * @param u The reflection universe, typically obtained from `context.universe` in a macro implementation.
+   * @param factory Factory to create the state.
+   */
+  def apply[T: ClassTag](u: Universe, factory: => T): T = {
```
dwijnand:
```suggestion
  def getOrElseUpdate[T: ClassTag](u: Universe, factory: => T): T = {
```

... except the key isn't the first param, it's `ClassTag[T].runtimeClass`.

But, still, I think it might help explain intent a bit better.

core/src/main/scala/shapeless/MacroState.scala
```diff
@@ -0,0 +1,48 @@
...
+
+
+private[shapeless] object MacroState {
+  private class MacroStateAttachment(val state: collection.mutable.HashMap[Class[_], Any])
+
+  /**
+   * Associates some state with a global.Run. Preferable to using static state in the macro
+   * classloader (e.g vars in top level objects) so as to avoid race conditions under `-Ycache-macro-classloader`.
+   *
+   * @tparam T the type of the state. The class tag for this type will be used a a map key, so each user of
```
dwijnand:
```suggestion
   * @tparam T the type of the state. The erased `Class[_]` value for this type will be used as the map key, so each user of
```

core/src/main/scala/shapeless/MacroState.scala
milessabin: macro-compat is on the way out, so it's best for it to be here.

milessabin:
Thanks for this! LGTM.

I'd prefer it to live in the same source file as the `Cached` type class, but I'll merge as is and tweak myself.

Thanks again.

PR #937, opened 2019-11-04
Fix #936

### LINKED ISSUE #936: RemoveAll for record gives results in wrong order in some cases
The `ops.record.RemoveAll` type class takes a removal argument that can either be labelled or unlabelled. If the original record contains a type that appears multiple times, and if this type also appears multiple times in an _unlabelled_ removal argument, the results will be in the wrong order.

For example:

```scala
import shapeless._
import ops.record.RemoveAll, record.Record, syntax.singleton._

type R = Record.`'i -> Int, 's -> String, 'c -> Char, 'j -> Int`.T
val r: R = Symbol("i") ->> 10 :: Symbol("s") ->> "foo" :: Symbol("c") ->> 'x' :: Symbol("j") ->> 42 :: HNil

type A = Int :: Int :: HNil
val removeAll = RemoveAll[R, A]
```
And then:
```scala
scala> removeAll(r)
res0: removeAll.Out = (42 :: 10 :: HNil,foo :: x :: HNil)
```

This is a pretty old bug (that I was responsible for) that @abdheshkumar [noticed last week](https://gitter.im/circe/circe?at=5db98cd19c39821509af7245).

travisbrown:
Fixes #936. I've also added some new tests, a couple of which fail without the change.

The fix breaks binary compatibility. It would be possible to fix the issue in a way that doesn't, but I noticed that master already contains a number of changes that aren't compatible with 2.3.3.

### DIFF
--- a/core/src/main/scala/shapeless/ops/records.scala
+++ b/core/src/main/scala/shapeless/ops/records.scala
@@ -549,19 +549,19 @@ package record {
 
     implicit def hconsRemoveAll[L <: HList, H, T <: HList, OutT <: HList, RemovedH, RemainderH <: HList, RemovedT <: HList, RemainderT <: HList]
       (implicit
-        rt: RemoveAll.Aux[L, T, (RemovedT, RemainderT)],
-        rh: Remove.Aux[RemainderT, H, (RemovedH, RemainderH)]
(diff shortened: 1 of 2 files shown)


milessabin:
LGTM. Many thanks! Merging ...

PR #943, opened 2019-12-07
Make Generic1, Split1, IsHCons1 and IsCCons1 covariant

joroKr21:
I.e. covariant in the captured instances, not the main parameter.
Simple-kinded `Generic` doesn't have an alternative,
because it doesn't need to capture any instances.

### DIFF
--- a/core/src/main/scala/shapeless/generic1.scala
+++ b/core/src/main/scala/shapeless/generic1.scala
@@ -20,7 +20,7 @@ import scala.language.experimental.macros
 
 import scala.reflect.macros.whitebox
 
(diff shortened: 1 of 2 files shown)


milessabin:
Can you motivate this for me?

joroKr21:
We can delegate the derivation of supertypes: typelevel/kittens#211

milessabin:
I'm convinced. Add a test or two and I'll merge :-)

joroKr21:
Finally done :smile:

milessabin:

PR #949, opened 2020-01-06
Update sbt to 1.3.6

fthomas:
Normally @scala-steward would have sent a PR for this update but sbt 1.3.5 breaks the `+` command in this repo which prevents @scala-steward from working on shapeless.

Here is what happens to me when trying `+`:
```
sbt:root> + libraryDependencies
[info] Forcing Scala version to 2.10.7 on all projects.
[info] Reapplying settings...
[info] Set current project to root (in build file:/home/frank/data/code/shapeless/)
[error] Expected ';'
[error] project root / libraryDependencies
[error]                                   ^
```
The same command succeeds with 1.3.6.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.5
+sbt.version=1.3.6


milessabin:
Is there an issue for that against SBT 1.3.5? Or is this a quirk of this build?

fthomas:
I found some recent issues and PRs that mention `+` (like https://github.com/sbt/sbt/pull/5265). They don't describe exactly the same error but I guess they are related to what I've observed above.

PR #954, opened 2020-01-23
Disable doc generation

esarbe:
doc generation for dotty doesn't seem to work atm


milessabin:
Thanks!

We should fix doc generation properly, but merging this for now means people can `publishLocal` and experiment.

PR #959, opened 2020-01-27
Support building with scala.js 1.0.0

cquiroz:
This PR supports building shapeless on both Scala.js branches, 0.6.x and 1.0.x. I tried local publication and it seems to work fine but it should be checked when doing a real release

There maybe a better way to organize how the CI build goes but I'm not that familiar with travis

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -14,3 +14,8 @@ addSbtPlugin("org.scoverage"                     % "sbt-scoverage"         % "1.
 addSbtPlugin("org.scala-native"                  % "sbt-scala-native"      % "0.4.0-M2")
 addSbtPlugin("org.portable-scala"                % "sbt-scalajs-crossproject" % "1.0.0")
 addSbtPlugin("org.portable-scala"                % "sbt-scala-native-crossproject" % "1.0.0")
+
+val scalaJSVersion =
+  Option(System.getenv("SCALAJS_VERSION")).getOrElse("1.0.0")
+
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -13,18 +13,20 @@ if [[ "${TRAVIS_PULL_REQUEST}" == "false" &&
 ]]; then
 PUBLISH=coreJVM/publish
 PUBLISHJS=coreJS/publish
+PUBLISHNAT=publish
 else
 PUBLISH=coreJVM/publishLocal
 PUBLISHJS=coreJS/publishLocal
+PUBLISHNAT=publishLocal
 fi
 
 if [[ "$NATIVE" == "true" ]]; then
-    ${SBT} clean validateNative coreNative/${PUBLISH}
+    ${SBT} clean validateNative coreNative/${PUBLISHNAT}
 #elif [[ "$TRAVIS_SCALA_VERSION" == 2.13.* ]]; then
 #    eval $COVERAGE && ${SBT} validate ${PUBLISH} && codecov
 else
   ${SBT} clean validateJVM ${PUBLISH}
-  for sjs in '1.0.0-RC2' '0.6.32'; do
+  for sjs in '1.0.0' '0.6.32'; do
     SCALAJS_VERSION=$sjs ${SBT} clean validateJS ${PUBLISHJS}
   done
 fi
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -11,9 +11,11 @@ if [[ "${TRAVIS_PULL_REQUEST}" == "false" &&
       "${TRAVIS_REPO_SLUG}" == "milessabin/shapeless" &&
       $(cat version.sbt) =~ "-SNAPSHOT"
 ]]; then
-PUBLISH=publish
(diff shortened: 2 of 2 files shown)


milessabin:
Aside from that ... this looks great! Thanks!

build.sbt
```diff
@@ -106,7 +109,8 @@ lazy val commonJsSettings = Seq(
 
 lazy val commonJvmSettings = Seq(
   parallelExecution in Test := false,
-  coverageExcludedPackages := "shapeless.examples.*"
+  coverageExcludedPackages := "shapeless.examples.*",
+  skip.in(publish) := customScalaJSVersion.forall(_.startsWith("0.6"))
```
cquiroz: This setting should prevent publishing the jvm side twice. I think native is called explicitly on the build and not as part of root

milessabin: I missed that. It still builds and tests non-js twice though, doesn't it?

cquiroz: Yes, that is true, I think CI times probably extended due to that change. I wonder if we could at least skip tests in the same way

milessabin: Perhaps it would be better to separate js and non-js builds completely?

cquiroz: Monocle takes that approach with two 'root' projects for JVM and JS

milessabin: Can we not do `coreJVM/publish` vs. `coreJS/publish`?

cquiroz:
I tried to split `coreJVM` and `coreJS` test and publish. Let's see what travis says about it

milessabin:
I think we're nearly there ... thanks!

At the moment it looks like we're only building with 1.0.0 for 2.13.x. I think we should build with 2.11.x and 2.12.x as well.

cquiroz:
I'm being blocked by this
https://travis-ci.org/milessabin/shapeless/jobs/646495468#L2054

I can't really tell what's going on but could it be that a macro generates a repeated symbol? It seems to happen only on 2.11 and 2.12. Any clues?

cquiroz:
The error is triggered on this test and call:
https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/hlist.scala#L3165

```scala
  @Test
  def testFromProductArgs: Unit = {
 ...
    implicit val i2 = 7
    val v3 = Bar.sumImplicitLabelProduct("foo" :: 1 :: HNil)
...
  }
```

It seems `sumImplicitLabelProduct` triggers a macro at
https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/hlists.scala#L330

I have no knowledge of macros but when I print the `product: Expr[L]` param in 2.12 and 2.13 it is different.
`2.13`
```
  HNil.::[Int](1)
}).::[String]("foo")
})
LisExpr[Nothing]({
  shapeless.this.HList.hlistOps[Int :: shapeless.HNil]({
  HNil.::[Int](1)
}).::[String]("foo")
})
```

in `2.12`
```
  HNil.::[Int](1)
}).::[String]("foo")
})
LisExpr[Nothing]({
  shapeless.this.HList.hlistOps[Int :: shapeless.HNil]({
  HNil.::[Int](1)
}).::[String]("foo")
})
Expr[Nothing]({
  <synthetic> <artifact> val x$2: String = "foo";
  shapeless.this.HList.hlistOps[Int :: shapeless.HNil]({
  <synthetic> <artifact> val x$1: Int = 1;
  HNil.::[Int](x$1)t / compileIncremental 12s
}).::[String](x$2)
})
```

I'm thinking the scala-2.12 version adds an extra variable that messes up things

cquiroz:
See https://github.com/scala-js/scala-js/issues/3955

cquiroz:
From the investigation on the scala.js side it looks like the macro is producing an illegal tree that just happens to work on the JVM side but not on the scala.js side
A good question is why is it different on 2.12 and 2.13

cquiroz:
An obvious workaround is to only execute this test on 2.13. It is lame but as much as I try to follow the macro code it can't really find why the expression inside is different (my macro skills are non existent)
@milessabin , what do you think of that option?

joroKr21:
Sorry for the delay, give me some time to take a look and try to find out why the macro produces faulty code. That sounds like a bug in shapeless.

milessabin:
I've had a chance to do a little digging now.

It looks as though `mkParamsImpl` is using the `product` tree it's passed as an argument multiple times (in the nested map over `slices`). On 2.{11,12} `product` contains local `ValDefs`, hence the duplicate definition, whereas on 2.13.x it doesn't. The quickest fix would probably be to create a local `ValDef` in `forwardFromProductImpl` and pass a reference to that through to `mkParamsImpl` instead.

But ... the implementation of `mkParamsImpl` isn't super efficient anyhow: indexing into the `HList` argument using `At` with a sequence of `Nat` arguments isn't exactly optimal, so this method could really do with a complete rework.

milessabin:
I have a fix, so I'm going to pick this PR up with that.

SJS doesn't support 2.10.x, so I think we should probably drop 2.10.x support at this point.

I'll back publish for 2.3.3 for Scala 2.{11, 12, 13}, but going forward SJS will only be fully supported on >= 2.4.x.

cquiroz:
Thanks, this is great

milessabin:
Closed in favour of #967.

PR #964, opened 2020-02-06
Add sliceFields and fieldIndex for records

ceedubs:
These are hopefully described well enough by their syntax methods:

```scala
/**
  * Return the `Nat` index of the field with singleton typed key `k`. Only
  * available if this record has a field with key type equal to the singleton
  * type `k.T`.
  */
def fieldIndex(k: Witness)(implicit fieldIndex: FieldIndex[L, k.T]): fieldIndex.Out = fieldIndex()

/**
  * Slices this record beginning at the field with the singleton typed key
  * `k1`, up through the field with singleton typed key `k2` (inclusive).
  */
def sliceFields(k1: Witness, k2: Witness)(implicit sliceFields: SliceFields[L, k1.T, k2.T]): sliceFields.Out = sliceFields(l)
```

Personally I don't have a use-case for `fieldIndex` itself - the
`FieldIndex` type class was just a means to implementing `sliceFields`.
I went ahead and added the syntax for it since it was available and
there was simmilar syntax for `length` on `HList`.

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -933,7 +933,7 @@ object hlist {
   }
 
   /**
-   * Type class supporting supporting access to the elements in range [a,b[ of this `HList`.
+   * Type class supporting access to the elements in range [a,b[ of this `HList`.
    * Avaialable only if this `HList` contains all elements in range
    *
    * @author Andreas Koestler
--- a/core/src/main/scala/shapeless/ops/records.scala
+++ b/core/src/main/scala/shapeless/ops/records.scala
@@ -873,4 +873,69 @@ package record {
       }
     }
   }
+
+  /**
+   * Type class returning the `Nat` index of a record field.
+   *
+   * @author Cody Allen
+   */
+  trait FieldIndex[L <: HList, K] extends DepFn0 with Serializable {
+    type Out <: Nat
+  }
+
+  object FieldIndex {
+    import shapeless.nat._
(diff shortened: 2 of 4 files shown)


joroKr21:
Hi @ceedubs, thank you for the contribution!

I wonder though, what is the motivation for this functionality.
Do you have an example how you would use it?

ceedubs:
@joroKr21 I was toying around with the idea of using a Shapeless record to represent a data pipeline. For example you could imagine something like:

```scala
val pipeline =
  ("filter" ->> ???) ::
  ("normalize" ->> ???) ::
  ("score" ->> ???) ::
  HNil
```

And then you could use the `slice` method to grab just a subset of the pipeline if say you wanted to look at the results of normalization before data was scored. It's a half-baked idea and I don't know if anything will come of it, but I think that there's some potential.

joroKr21:
@ceedubs I'm mildly 👎 on adding a half-baked API at this point.

Shapeless 2.4 is long overdue and I'm not sure an addition like this would make the cut.
The plan is to have Shapeless 2.5 on 2.13 only and we will probably drop `Witness` because we would be able to use singleton types. Then this PR might become more interesting.

I also think that having those methods in user land is not so bad.

ceedubs:
For what it's worth I wouldn't consider the API to be half-baked. I think that it mirrors the index-based HList API pretty closely. I think that the idea of using it for a data pipeline is half-baked. But I understand that there should be skepticism involved in adding to a library. I'll close this out.

PR #969, opened 2020-02-27
Optimize SelectRange typeclass derivation

Ssstlis:
Hi all!
Once i've find out the SelectRange is useful, but i don't understand why is this instances are so hard to derive (too much unnecessary recursive calling).
So i decide to optimize this derivation and it still works and looks good.

### DIFF
--- a/core/src/main/scala/shapeless/ops/hlists.scala
+++ b/core/src/main/scala/shapeless/ops/hlists.scala
@@ -933,10 +933,10 @@ object hlist {
   }
 
   /**
-   * Type class supporting supporting access to the elements in range [a,b[ of this `HList`.
-   * Avaialable only if this `HList` contains all elements in range
+   * Type class supporting supporting access to the elements in range [a,b] of this `HList`.
+   * Available only if this `HList` contains all elements in range
    *
-   * @author Andreas Koestler
+   * @author Ivan Aristov
    */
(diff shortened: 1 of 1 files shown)


milessabin:
@joroKr21 I'd appreciate your eyeballs on this :-)

core/src/main/scala/shapeless/ops/hlists.scala
```diff
@@ -946,12 +946,22 @@ object hlist {
 
     type Aux[L <: HList, A <: Nat, B <: Nat, Out0 <: HList] = SelectRange[L, A, B] {type Out = Out0}
 
-    implicit def SelectRangeAux[L <: HList, A <: Nat, B <: Nat, Ids <: HList, SelOut <: HList]
-    (implicit range: shapeless.ops.nat.Range.Aux[A, B, Ids], sel: SelectMany.Aux[L, Ids, SelOut]): Aux[L, A, B, SelOut] =
-      new SelectRange[L, A, B] {
-        type Out = SelOut
+    implicit def thenTake[L <: HList, T <: Nat](
```
joroKr21: Not sure about the name of this method. Why not just call it `take`?

joroKr21:
LGTM but I'm not sure about the naming.

core/src/main/scala/shapeless/ops/hlists.scala
Ssstlis: Then i should rename also a param name. Ok, i have an idea.

joroKr21:
@Ssstlis one more thing before merge, would you mind squashing all commits into one?

Ssstlis:
squashed and rebased

PR #976, opened 2020-03-15
Fix new singleton symbol deprecations.

milessabin:
Fixes new issues which showed up in builds against the [scalac nightly](https://travis-ci.org/github/milessabin/shapeless-test-scala-nightly/jobs/662564094#L237-L366).

### DIFF
--- a/core/src/main/scala/shapeless/singletons.scala
+++ b/core/src/main/scala/shapeless/singletons.scala
@@ -191,25 +191,21 @@ trait SingletonTypeUtils extends ReprTypes {
     }
   }
 
-  def isSymbolLiteral(typeStr: String): Boolean =
-    typeStr.startsWith("'") && !typeStr.endsWith("'")
-
   def parseSingletonSymbolType(typeStr: String): Option[Type] =
-    if (isSymbolLiteral(typeStr))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/singletons.scala
```diff
@@ -191,18 +191,25 @@ trait SingletonTypeUtils extends ReprTypes {
...
   def parseLiteralType(typeStr: String): Option[Type] =
-    for {
-      parsed <- Try(c.parse(typeStr)).toOption
-      checked = c.typecheck(parsed, silent = true)
-      if checked.nonEmpty
-      tpe <- SingletonType.unapply(checked)
-    } yield tpe
+    if (isSymbolLiteral(typeStr))
+      Some(SingletonSymbolType(typeStr.tail))
+    else
```
joroKr21: Why not reuse `parseSingletonSymbolType`?

milessabin: Fair comment.

Unfortunately it doesn't seem to be possible to simplify `parseType` to, `parseLiteralType(typeStr) orElse parseStandardType(typeStr)`.

joroKr21: Yep that's exactly what I meant 👍

PR #980, opened 2020-03-18
Poly builders require no import

joroKr21:
The insight is that `build` is now an object (see c9cf7499d87dc55fc5a15c42abdb7a4c981aa5a6).
All we need is a stable path to it which we get if we capture the
preceding part of the definition in a val.

@milessabin in light of this do we want to rename `build` to something else?
I couldn't come up with a better name right away, so opening this PR to discuss.

Addresses #632

### DIFF
--- a/core/src/main/scala/shapeless/poly.scala
+++ b/core/src/main/scala/shapeless/poly.scala
@@ -187,6 +187,7 @@ object PolyDefns extends Cases {
  */
 trait Poly extends PolyApply with Serializable {
   import poly._
+  type λ = this.type
 
   def compose(f: Poly) = new Compose[this.type, f.type](this, f)
 
--- a/core/src/test/scala/shapeless/poly.scala
+++ b/core/src/test/scala/shapeless/poly.scala
@@ -383,8 +383,7 @@ class PolyTests {
 
   @Test
   def testPoly1Builder: Unit = {
-    val myPoly = Poly1.at[Int]( x => x).at[String](_.length).at[Boolean](if(_) 1 else 0).build
-    import myPoly._
+    val myPoly = Poly1.at[Int](x => x).at[String](_.length).at[Boolean](if(_) 1 else 0)
 
     val r1 = myPoly(10)
     assertTypedEquals[Int](10, r1)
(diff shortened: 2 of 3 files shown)


milessabin:
I'm not keen on the change at use-sites from,
```scala
val r1 = myPoly(10)
```
to
```scala
val r1 = myPoly.build(10)
```

joroKr21:
> I'm not keen on the change at use-sites from,

I fixed that particular quirk.

joroKr21:
@milessabin we can even call
```diff
- .foldLeft(0)(myPoly.build)
+ .foldLeft(0)(myPoly)
```

if we do this change:
```diff
-   def foldLeft[R](z : R)(op : Poly)(implicit folder : LeftFolder[L, R, op.type]) : folder.Out = folder(l, z)
+   def foldLeft[R](z : R)(op : PolyApply)(implicit folder : LeftFolder[L, R, op.λ]) : folder.Out = folder(l, z)
```

But that would be a massive API change. WDYT?

milessabin:
I would definitely prefer it if we can preserve the `.foldLeft(0)(myPoly)` syntax.

How does the switch to `op: PolyApply` help with that?

joroKr21:
I added an abstract type in `PolyApply`  that we can use to reference the implicit scope of the  `Case`s.
Conceptually something like this:
```scala
trait PolyApply {
  type λ <: Singleton
}

trait Poly extends PolyApply {
  type λ = this.type
}

trait PolyBuilder extends PolyApply {
  type λ = build.type
  object build extends Poly
}
```

milessabin:
Ahh ... I think I see.

But do `Poly` and `PolyBuilder` actually need to be siblings? Could `PolyBuilder` extend `Poly`?

joroKr21:
> Could PolyBuilder extend Poly?

I guess it could but then we are back to the original problem that `Poly` needs to be an object to have an implicit scope which is why the builder exists in the first place. So in the end we need _some_ object to hold the implicit scope and in the case of builders we can make that a member object.

milessabin:
OK, I think I'm fine with this as-is.

PR #984, opened 2020-03-20
Add missing asInstanceOf call

### LINKED ISSUE #889: Implicit inference for ops.record.Selector doesn't work when Id is passed to a HKT
```
Welcome to Scala 2.12.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_172).
Type in expressions for evaluation. Or try :help.

scala> :settings
-Xprint = List()
-Ybrowse = List()
-Ypartial-unification = true
-Yrepl-sync = true
-Ytyper-debug = false
-d = .
-nowarn = false

scala> case class Hkt[F[_]](foo: F[String])
warning: there was one feature warning; for details, enable `:setting -feature' or `:replay -feature'
defined class Hkt

scala> import shapeless._
import shapeless._

scala> def func[T] = new { def apply[R <: HList](k: Witness)(implicit gen: LabelledGeneric.Aux[T, R], selector: ops.record.Selector[R, k.T]) = 0 }
func: [T]=> AnyRef{def apply[R <: shapeless.HList](k: shapeless.Witness)(implicit gen: shapeless.LabelledGeneric.Aux[T,R],implicit selector: shapeless.ops.record.Selector[R,k.T]): Int}

scala> func[Hkt[Option]].apply('foo)
warning: there was one feature warning; for details, enable `:setting -feature' or `:replay -feature'
res1: Int = 0

scala> func[Hkt[Id]].apply('foo)
<console>:18: error: could not find implicit value for parameter gen: shapeless.LabelledGeneric[Hkt[[+T]T]]{type Repr = R}
       func[Hkt[Id]].apply('foo)
                          ^
```

joroKr21:
fixes #889

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -941,7 +941,7 @@ class GenericMacros(val c: whitebox.Context) extends CaseClassMacros {
     val (p, ts) = ctorDtor.binding
     val to = cq"$p => ${mkHListValue(ts)}.asInstanceOf[$repr]"
     val (rp, rts) = ctorDtor.reprBinding
-    val from = cq"$rp => ${ctorDtor.construct(rts)}"
+    val from = cq"$rp => ${ctorDtor.construct(rts)}.asInstanceOf[$tpe]"
     q"$generic.instance[$tpe, $repr]({ case $to }, { case $from })"
   }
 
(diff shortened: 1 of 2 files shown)


milessabin:
Why is the cast needed there?

joroKr21:
> Why is the cast needed there?

Because Scala won't infer `Id`. But it seems that it doesn't work for all versions 🤔

joroKr21:
Looks like this will only work on Scala 2.13 so let's leave it for 2.5.0

joroKr21:
@milessabin I think we should still merge that because it improves the situation on Scala 2.13

PR #991, opened 2020-03-24
Fix Lazy implicit resolution with type parameters

### LINKED ISSUE #828: Lazy does not instantiate type parameters
From my testing, Lazy does not work if more than a single type parameter is involved. I have only tested two myself.

```scala
import shapeless._

object LazyTest {

  new FindSimpleFooer[Int].find
  new FindLazySimpleFooer[Int].find
  new FindFooer[Int].find
  new FindLazyFooer[Int].find
}

class FindSimpleFooer[Tpe] {
  def find(implicit fooer: SimpleFooer[Tpe]) = fooer
}

class FindLazySimpleFooer[Tpe] {
  def find(implicit fooer: Lazy[SimpleFooer[Tpe]]) = fooer
}

class FindFooer[Tpe] {
  def find[F <: Foo](implicit fooer: Fooer[Tpe, F]) = fooer
}

class FindLazyFooer[Tpe] {
  def find[F <: Foo](implicit fooer: Lazy[Fooer[Tpe, F]]) = fooer
}

trait Foo
case class BarFoo(i: Int) extends Foo

trait SimpleFooer[Tpe] {
  def makeFoo(obj: Tpe): Foo
}
object SimpleFooer {
  implicit val barFooer: SimpleFooer[Int] = (obj: Int) => BarFoo(obj)
}

trait Fooer[Tpe, FooTpe <: Foo] {
  def makeFoo(obj: Tpe): FooTpe
}
object Fooer {
  implicit val barFooer: Fooer[Int, BarFoo] = (obj: Int) => BarFoo(obj)
}
```

joroKr21:
Fallback to replacing type parameters with bounded wildcards.
Might cause performance degradation in case of error so not sure.

Fixes #828

### DIFF
--- a/core/src/main/scala/shapeless/lazy.scala
+++ b/core/src/main/scala/shapeless/lazy.scala
@@ -295,17 +295,21 @@ class LazyMacros(val c: whitebox.Context) extends CaseClassMacros with OpenImpli
 
       private var current = Option.empty[State]
 
+      private def typeParamsToWildcards(tpe: Type): Type = tpe.map { t =>
+        val sym = t.typeSymbol
+        if (sym.isParameter) boundedWildcardType(sym.info.asInstanceOf[TypeBounds]) else t
+      }
+
       def resolveInstance(state: State)(tpe: Type): Option[(State, Tree)] = {
         val former = State.current
         State.current = Some(state)
         val (state0, tree) =
           try {
-            val tree = c.inferImplicitValue(tpe, silent = true)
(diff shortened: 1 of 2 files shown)


joroKr21:
Even internally to the compiler enhancing type parameters with bounds is kinda buggy and side effecting. I tried to extend it to take advantage of the expected type but it didn't work: scala/scala#8582

milessabin:
Yeah, I tried something like that ages ago too.

PR #994, opened 2020-03-27
Add mkAttributedRef specialized for SingleType

joroKr21:
References to singleton types should refer to the getter if possible.
Fixes an issue where we would try to reference a non-existing field.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -299,7 +299,6 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
   val c: blackbox.Context
 
   import c.universe._
-  import internal.constantType
 
   def abort(msg: String) =
     c.abort(c.enclosingPosition, msg)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/generic.scala
```diff
@@ -1017,7 +1021,7 @@ class GenericMacros(val c: whitebox.Context) extends CaseClassMacros {
   }
 
   def mkProductGeneric(tpe: Type): Tree = {
-    val repr = mkHListTypTree(fieldsOf(tpe).map(_._2))
+    val repr = mkHListTpe(fieldsOf(tpe).map(_._2))
```
joroKr21: Now `mkHListTypTree` is not used in shapeless anymore.
But it might be wise to keep it because people might be using it in their own macros.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -1029,18 +1033,16 @@ class GenericMacros(val c: whitebox.Context) extends CaseClassMacros {
...
+      case singleton: SingleType =>
+        cq"p if p eq ${mkAttributedRef(singleton)} => $index"
       case _ =>
         cq"_: $tpe0 => $index"
     }
 
     val coproduct = objectRef[Coproduct.type]
     val ctors = ctorsOf(tpe)
-    val repr = mkCoproductTypTree(ctors)
+    val repr = mkCoproductTpe(ctors)
```
joroKr21: Now `mkCoproductTpe` is not used in shapeless anymore.
But it might be wise to keep it because people might be using it in their own macros.

core/src/main/scala/shapeless/generic.scala
milessabin: I'm fine removing it for 2.4.0.

core/src/main/scala/shapeless/generic.scala
milessabin: Ditto.

core/src/main/scala/shapeless/generic.scala
joroKr21: Ok it turns out there is a lot of dead code already 😄

milessabin:
Latest nightly was [green](https://travis-ci.org/github/milessabin/shapeless-test-scala-nightly/builds/667934458) ... we're good to go :smile:

PR #996, opened 2020-03-28
Setup sbt-ci-release

joroKr21:
@milessabin I think all that is left should be to setup GPG:
https://github.com/olafurpg/sbt-ci-release#gpg


milessabin:
Will need to follow up with a downgrade to openjdk8.

milessabin:
It looks like this is only publishing for 2.11 ...

PR #998, opened 2020-03-29
Make +publishLocal publish full cross-build

olafurpg:
Previously, running `sbt +publishLocal` only published the active platform
(JVM, JS or Native). Now, the command publishes the full cross-build for
both Scala versions (2.11/2.12/2.13) and platforms (JVM, JS, Native).
This change makes it possible for the `sbt ci-release` to release all
modules in one step excluding Scala.js v0.6.x. The Travis script has
been updated to handle Scala.js v0.6.x in a separate sbt command, using
a similar setup as the MUnit build (see https://github.com/scalameta/munit)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,6 @@ env:
     - SCALA_JS_VERSION=1.0.1
     - SCALA_NATIVE=true
   global:
-    - CI_RELEASE=publishSigned
-    - CI_SNAPSHOT_RELEASE=publish
     - secure: eXcMQIb0hHeaJsJGaJYhLS8aONu5YnhK1NOHJZGTBIb6ukSd48uXm6xNgBMozgkOEUBXyrBU9ZnCOxSS+HmJZDbK0Wkr90C4sZH7dHBZYMN3hNfCp5SjPFY2+TgrYZ24G8s/BPei6725AlbXM0vmdXPsvxaOvT36d2Ej7xto8+c=
     - secure: c3IO3G34aTFD+fh/5vqc3sCgwMaClPDziYQDCEUqQBOQFdTS7wbBZFVpqTUUlkMLkeyHZ0BGSLUiDZnu8TGGITtgJBaCUchhLYK5WOavyzCl/NPVOBebviazgtBOzPDKyYDcahjmxCGrMspTiRei86VgCdkNIvldJJm7zK1OAOU=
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -49,7 +47,10 @@ notifications:
 jobs:
   include:
     - stage: release
-      script: sbt ++$TRAVIS_SCALA_VERSION! ci-release
+      script:
+      - |
```
olafurpg: I may have messed up the multiline YAML syntax here, I'm guessing from some examples I found online

build.sbt
```diff
@@ -110,11 +115,8 @@ lazy val commonNativeSettings = Seq(
 
 lazy val coreSettings = commonSettings ++ publishSettings
 
-lazy val root = project.in(file("."))
```
olafurpg: I removed this since it was not aggregating the `coreNative` module. This changes the name of the root module breaking some aliases I saw are defined in build.sbt. You can probably keep the root project if you add `aggregate(coreJVM, coreJS, coreNative)` but make sure to also add `settings(crossScalaVersions := List())`

build.sbt
```diff
@@ -147,9 +149,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform).crossType(
   .jsSettings(commonJsSettings:_*)
   .jvmSettings(commonJvmSettings:_*)
   .jvmSettings(scoverageSettings:_*)
-  .jvmSettings(skip in publish := platform != JVMPlatform)
-  .jsSettings(skip in publish := platform != JSPlatform)
-  .nativeSettings(skip in publish := platform != NativePlatform)
+  .jvmSettings(skip in publish := isCustomScalaJSVersion)
```
olafurpg: This is to prevent releasing JVM/Native artifacts a second time when the JS 0.6 release is happening.

.travis.yml
joroKr21: 🤞

build.sbt
joroKr21: I think we should remove those aliases.

joroKr21: Actually I think that was on purpose. Now opening `sbt` and running `test` fails locally, which is not cool for development.
So I think we should keep the root project as is (the only change I would do is rename it to `shapeless`).

olafurpg:
What is the expected behavior for `coreNative/test`? I currently get compile errors. If the goal is to make `sbt test` succeed without testing on native then you can use the following instead

```scala
.nativeSettings(
  sources in Test := Nil
)
```

The root project should aggregate `coreNative` so that `+publishLocal` publishes the full cross-build. The job on master only published JVM and JS. The root project should also have `crossScalaVersions := List()` in order to avoid resolution errors (I'm not 100% sure why this is the case)

joroKr21:
Oh that is unfortunate. The problem I guess is there is no JUnit for native.

PR #999, opened 2020-03-29
Enable publishing of native module.

olafurpg:
Previously, the coreNative module was not published in the
`+publishLocal` command. Now, the root project aggregates `coreNative`
so that the module is correctly published. To make `sbt test` still
work like before, testing is disabled in Scala Native.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ target/
 .*.swp
 .swo
 .*.swo
+lowered.hnir
--- a/.jvmopts
+++ b/.jvmopts
@@ -2,4 +2,4 @@
 -Dsbt.supershell=true
 -Xms2g
 -Xmx3g
--Xss2m
+-Xss4m
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,16 +14,11 @@ env:
   - {}
   - SCALA_JS_VERSION=0.6.32
   - SCALA_JS_VERSION=1.0.1
-  - SCALA_NATIVE=true
 
 stages:
   - name: test
   - name: release
     if: ((branch = master AND type = push) OR (tag IS present)) AND NOT fork
-
-before_install:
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -30,3 +30,4 @@ target/
 .*.swp
 .swo
 .*.swo
+lowered.hnir
```
olafurpg: This file is generated on `coreNative/test`. I think it no longer does this in the upcoming native version

joroKr21:
@olafurpg thanks a lot for your help 💯 I think this did the trick. In the end I deleted the root project again.

olafurpg:
Awesome! The next step is to try a non snapshot release. They often fail because of random gpg issues

milessabin:
:clap:

Let's cut a 2.4.0-M1 ... @joroKr21 do you want to do the honours?

joroKr21:
Sure! I just need to push a tag right? 🎉

olafurpg:
Yes. I normally push the tag from the command line and write the release notes while waiting for the release to finish

joroKr21:
Cool the release worked 🎉 
The release notes will take a while. It's been so long since the last major release.

olafurpg:
Very cool! Congrats 🎉 Now with releases requiring less ceremony it may be easier to write release notes in the future 😉 

Writing release notes always takes time for large projects. I have found it helpful to use "release drafter" to get a quick summary of PRs that have been merged. You can set it up with GitHub Actions by copying these two files

https://github.com/scalameta/munit/blob/master/.github/workflows/release-drafter.yml
https://github.com/scalameta/munit/blob/master/.github/workflows/release-drafter.yml

It generates release templates like this https://github.com/scalameta/munit/releases/tag/v0.7.1 which you can then enrich with more details.

PR #1005, opened 2020-04-11
Remove deprecated methods which are kept for bincompat

joroKr21:


alexarchambault:
@joroKr21 Don't know if you were planning to remove them in a separate PR, but there's [a few methods / classes](https://github.com/milessabin/shapeless/search?q=stub&unscoped_q=stub) which were added for binary compatibility, and not flagged as deprecated, that you may want to remove too.

joroKr21:
Oh cool thanks, I didn't notice those. I think we should remove them.

PR #1011, opened 2020-04-18
Add DropRange typeclass for HList and Tuple

Ssstlis:
Hi!
I just add two typeclasses for HList and Tuple ops.
My motivation is to create DropRange with it's behavior to use it something like this between two classes:
```scala
//Demo
final case class A(
  a: String,
  b: Boolean,
  c: Int,
  d: DateTime
)

final case class B(
  a: String,
  d: DateTime
)

val a: A = A("", false, 0, DateTime.now())
val b: B =  Generic[B].from(Generic[A].to(a).dropRange(2, 3))
```


Ssstlis:
I'd be appreciate for @joroKr21 review =)

joroKr21:
Sorry, still pondering. I'm afraid of bloating the API.
In general I have no objections to adding type level versions of std lib collection methods.
But this is not one of them so I'm not sure.

@milessabin WDYT?

milessabin:
I'm a bit puzzled by `DropAt` ... how does it differ from `Drop`?

Ssstlis:
Yes, you're right, i missed that typeclass. I'll remove it from PR.

milessabin:
Thanks.

I guess I'm of the same mind as @joroKr21. It would help tip me in favour if you could describe a compelling example of it in use.

milessabin (approved):
LGTM!

joroKr21:
I'm mildly 👎 due to the reasons already stated

milessabin:
Yes, sorry, I approved in error.

I think we should close this now.

PR #1019, opened 2020-05-01
Compiler plugin

joroKr21:
* `Lazy[T]` / `Strict[T]` to `=> T` on Scala 2.13
 * `=> T` to `Lazy[T]` on Scala 2.12
 * drop Scala 2.11


joroKr21:
@milessabin I don't know what I'm doing wrong.
But this rather straightforward compiler plugin just doesn't work...

plugin/src/main/scala/shapeless/ShapelessPlugin.scala
```diff
@@ -0,0 +1,44 @@
...
+          val adaptedRhs = super.transform(rhs)
+          if (adaptedImplicits.eq(implicits) && adaptedRhs.eq(rhs)) method
+          else treeCopy.DefDef(method, mods, name, tparams, paramss :+ adaptedImplicits, tpt, adaptedRhs)
+        case _ =>
+          super.transform(tree)
+      }
+
+      private def adaptImplicit(param: ValDef): ValDef = param.tpt match {
+        case AppliedTypeTree(tq"Lazy" | tq"shapeless.Lazy" | tq"_root_.shapeless.Lazy", targs @ List(_)) =>
+          val mods = param.mods & Flag.BYNAMEPARAM
```
joroKr21, line 34:
```suggestion
          val mods = param.mods | Flag.BYNAMEPARAM
```

joroKr21: face palm

core/src/test/scala/shapeless/orphans.scala
```diff
@@ -261,8 +271,18 @@ class OrphansTest {
     assertEquals("project(product(Eq.eqString, emptyProduct))", implicitly[Eq[Bar]].toString)
     assertEquals("Baz.eqBaz", implicitly[Eq[Baz]].toString)
     assertEquals("Eq.fallback", implicitly[Eq[Quux]].toString)
-    assertEquals("project(coproduct(project(product(project(<loop>), emptyProduct)), coproduct(project(emptyProduct), emptyCoproduct)))", implicitly[Eq[SimpleRec]].toString)
-    assertEquals("project(coproduct(project(product(Eq.eqInt, emptyProduct)), coproduct(project(product(project(<loop>), product(project(<loop>), emptyProduct))), emptyCoproduct)))", implicitly[Eq[Tree[Int]]].toString)
+
+    val simpleRec = implicitly[Eq[SimpleRec]].toString
+    assertTrue(
+      simpleRec == "project(coproduct(project(product(project(<loop>), emptyProduct)), coproduct(project(emptyProduct), emptyCoproduct)))"
+        || simpleRec == "Eq.fallback"
```
joroKr21, line 278: It seems like the `Orphan` macro doesn't work with by-name implicits.
To be honest, I have no motivation to fix it. I don't even know what it's used for.

core/src/test/scala/shapeless/orphans.scala
```diff
@@ -261,8 +271,18 @@ class OrphansTest {
...
+    )
+
+    val tree = implicitly[Eq[Tree[Int]]].toString
+    assertTrue(
+      tree == "project(coproduct(project(product(Eq.eqInt, emptyProduct)), coproduct(project(product(project(<loop>), product(project(<loop>), emptyProduct))), emptyCoproduct)))"
+        || tree == "Eq.fallback"
```
joroKr21, line 284: Same here

core/src/test/scala/shapeless/orphans.scala
```diff
@@ -274,7 +294,17 @@ class OrphansTest {
     assertEquals("project(product(Eq.eqString, emptyProduct))", implicitly[Eq[Bar]].toString)
     assertEquals("Baz.eqBaz", implicitly[Eq[Baz]].toString)
     assertEquals("Eq.fallback", implicitly[Eq[Quux]].toString)
-    assertEquals("project(coproduct(project(product(project(<loop>), emptyProduct)), coproduct(project(emptyProduct), emptyCoproduct)))", implicitly[Eq[SimpleRec]].toString)
-    assertEquals("project(coproduct(project(product(Eq.eqInt, emptyProduct)), coproduct(project(product(project(<loop>), product(project(<loop>), emptyProduct))), emptyCoproduct)))", implicitly[Eq[Tree[Int]]].toString)
+
+    val simpleRec = implicitly[Eq[SimpleRec]].toString
+    assertTrue(
+      simpleRec == "project(coproduct(project(product(project(<loop>), emptyProduct)), coproduct(project(emptyProduct), emptyCoproduct)))"
+        || simpleRec == "Eq.fallback"
```
joroKr21, line 301: `Orphan` not working

core/src/test/scala/shapeless/orphans.scala
```diff
@@ -274,7 +294,17 @@ class OrphansTest {
...
+    )
+
+    val tree = implicitly[Eq[Tree[Int]]].toString
+    assertTrue(
+      tree == "project(coproduct(project(product(Eq.eqInt, emptyProduct)), coproduct(project(product(project(<loop>), product(project(<loop>), emptyProduct))), emptyCoproduct)))"
+        || tree == "Eq.fallback"
```
joroKr21, line 307: `Orphan` not working

core/src/test/scala/shapeless/sybclass.scala
```diff
@@ -486,7 +485,7 @@ class SybClassTests {
     assertEquals(expected4, result4)
   }
 
-  @Test
+  @Ignore
```
joroKr21, line 488: Unfortunately SYB is broken for recursive data types.
I couldn't find a way to fix it @milessabin

joroKr21:
I need some help with the build.
I explicitly set `crossScalaVersions := Seq(Scalal213)` but we get this on 2.11:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/travis/build/milessabin/shapeless/"), "plugin"):
[error]    org.scala-lang.modules:scala-xml _2.12, _2.11
[error]    org.scala-lang.modules:scala-parser-combinators _2.12, _2.11
[error] java.lang.RuntimeException: Conflicting cross-version suffixes in: org.scala-lang.modules:scala-xml, org.scala-lang.modules:scala-parser-combinators
```

milessabin:
I'd be happy to drop the `Orphan` stuff. It was an experiment in the same general area as export-hook. AFAIK it's not widely used and it won't be relevant in Dotty, so I think we should just rip it out.

I'm more concerned by the SYB issues. Can you get it working for 2.13 using only by-name implicits?

joroKr21:
> Can you get it working for 2.13 using only by-name implicits?

No, same issue - incrementing a tree with `everywhere` doesn't work.
Mutual recursion is not supposed to be a problem for by-name implicits right?

milessabin:
> Mutual recursion is not supposed to be a problem for by-name implicits right?

No, that should be fine. I guess the derivation will need reworking for by-name implicits. I put together something different for shapeless-3, but that probably won't backport sensibly to 2.

I think maybe the best thing to do is split it off to a module for now so we can disable the test with a slightly easier conscience. I'll try and find time to take a look at it over the weekend.

joroKr21:
@milessabin did you have a chance to take a look? I tried a few variants but nothing worked.
It looks like the implicit search falls back to the default instance rather than tie the recursive knot...

milessabin:
@joroKr21 sorry, not yet. I'll try and make time for it tomorrow.

milessabin:
I managed to get SYB working with by-name implicits: https://github.com/milessabin/shapeless/commit/774bf0fb5387d6989ff607c83c8d7535fe4dc962.

Originally it was designed so that `everything` and `everywhere` could be curried, but it just made things more complicated so I've dropped it. I'd be surprised if anyone has used that facility.

It's very fragile and the compile times are terrible, so I think it definitely makes sense to move it to a module with a view to eventually retiring it altogether. I have something slightly simpler working for shapeless 3 using the type class derivation infrastructure. In any case I don't think it warrants much more time being spent on it.

joroKr21:
> Originally it was designed so that everything and everywhere could be curried, but it just made things more complicated so I've dropped it. I'd be surprised if anyone has used that facility.

Oh cool, I hadn't tried that 👍 

I will incorporate this change here.
Agree that it doesn't make sense to spend more time on it.

milessabin:
I think this is great, but I still think we should go in the other direction, even though that means abandoning 2.11.x and waiting for 2.12.3.

joroKr21:
I haven't had time for it yet but I think we should use a nightly.
Also I just realised I haven't added the plugin to the test scalac option 🤦

core/src/main/scala_2.13-/shapeless/LazyMacros.scala
```diff
@@ -165,14 +165,7 @@ class LazyMacros(val c: whitebox.Context) extends CaseClassMacros with OpenImpli
           case Right((state, inst)) =>
             val (tree, actualType) = if (root) mkInstances(state)(instTpe0) else (inst.ident, inst.actualTpe)
             current = if (root) None else Some(state)
-            if (root) {
-              val valNme = TermName(c.freshName("inst"))
-              q"""
-              val $valNme: $actualType = $tree
```
joroKr21: @milessabin why was this an eager val? It caused execution order tests to fail:
https://travis-ci.org/github/milessabin/shapeless/jobs/695762315

joroKr21:
@milessabin but there is no way to know where we would need to add `Strict` in 2.12

milessabin:
I think we should just abandon `Strict` ... there won't be any replacement for it going forward. I'm also pretty sure it's not widely used.

milessabin:
I'm afraid I'm :-1: on going in this direction now that we have the option in a released 2.12 to retrofit by-name implicit syntax to a pre-2.12 release.

joroKr21:
> I'm afraid I'm 👎 on going in this direction now that we have the option in a released 2.12 to retrofit by-name implicit syntax to a pre-2.12 release.

That would mean dropping Scala 2.11 and Scala Native.

milessabin:
> That would mean dropping Scala 2.11 and Scala Native.

I think we're just going to have to go that way. People stuck on 2.11 will still have 2.3.3.

I think we really need to get to a core of shapeless 2 which will cross build on 2.12, 2.13 and 3.0. I know of several TL projects which are blocked on this now.

joroKr21:
@milessabin finally got this working 😄

plugin/src/main/scala_2.13-/shapeless/ShapelessPlugin.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import scala.tools.nsc.Global
+import scala.tools.nsc.plugins._
+import scala.tools.nsc.transform._
+
+class ShapelessPlugin(val global: Global) extends Plugin { self =>
+  import global._
+
+  val name = "shapeless-plugin"
+  val description = "Replaces by-name implicits with shapeless.Lazy on Scala 2.13"
```
milessabin, line 11: Shouldn't this be "on Scala < 2.13"?

plugin/src/main/scala_2.13+/shapeless/ShapelessPlugin.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import scala.tools.nsc.Global
+import scala.tools.nsc.plugins._
+import scala.tools.nsc.transform._
+
+class ShapelessPlugin(val global: Global) extends Plugin { self =>
+  import global._
+
+  val name = "shapeless-plugin"
+  val description = "Replaces shapeless.Lazy and shapeless.Strict with by-name implicits on Scala 2.13"
```
milessabin, line 11: It hadn't occurred to me to go in this direction as ... that's awesome!

milessabin (approved):
Aside from the 2.13- plugin description ... :tada: !!! LGTM !!! :tada:

milessabin:
This presumably has to be 2.4.0 due to bincompat?

joroKr21:
> This presumably has to be 2.4.0 due to bincompat?

Yes, for sure 👍

joroKr21:
@milessabin I guess we also want to publish it - is the name `shapeless-plugin` good for that?

joroKr21:
Also the branch protection for the 2.11 build would have to be disabled before merging

milessabin:
:+1: for `shapeless-plugin`.

I'll disable the 2.11 branch checks.

PR #1020, opened 2020-05-04
make MirroredElemTypes type more specific

letalvoj:
The `MirroredElemTypes` is expected to be a `Tuple` in the standard library.

https://github.com/lampepfl/dotty/blob/fe03ee22f107a745dd3a25b4e5bbd35856188206/library/src/scala/deriving.scala#L51-L53

The refined type `Kind` should reflect that as well.

### DIFF
--- a/modules/deriving/src/main/scala/shapeless3/deriving/kinds.scala
+++ b/modules/deriving/src/main/scala/shapeless3/deriving/kinds.scala
@@ -20,7 +20,7 @@ import scala.compiletime._
 import scala.deriving._
 
 object K0 {
-  type Kind[C, O] = C { type Kind = K0.type ; type MirroredType = O ; type MirroredElemTypes }
+  type Kind[C, O] = C { type Kind = K0.type ; type MirroredType = O ; type MirroredElemTypes <: Tuple }
   type Generic[O] = Kind[Mirror, O]
   type ProductGeneric[O] = Kind[Mirror.Product, O]
   type CoproductGeneric[O] = Kind[Mirror.Sum, O]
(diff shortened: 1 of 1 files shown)


letalvoj:
@milessabin ok, it was my code, which required the type param to be `<: Tuple`. I wanned to have an exhaustive match so I specified the type of the first param

```scala
@tailrec inline def mapRec[T <: Tuple, A, B](p:Product,f: A => B)(i:Int): Tuple = inline erasedValue[T] match
    case _: (Stub *: tt) =>  f(productElement[A](p,i)) *: mapRec[tt,A,B](p,f)(i+1)
    case _: (t *: tt) =>     productElement[t](p,i) *: mapRec[tt,A,B](p,f)(i+1)
    case _: Unit =>          ()
```

Nevertheless, should the `Kind` definition be inconsistent with the `Mirror.*Of` definition from core lib?

```scala
    type Of[T]        = Mirror { type MirroredType = T; type MirroredMonoType = T ; type MirroredElemTypes <: Tuple }
    type ProductOf[T] = Mirror.Product { type MirroredType = T; type MirroredMonoType = T ; type MirroredElemTypes <: Tuple }
    type SumOf[T]     = Mirror.Sum { type MirroredType = T; type MirroredMonoType = T; type MirroredElemTypes <: Tuple }
```

milessabin:
Happy to merge this if you add a test which didn't compile before and does compile with this change.

letalvoj:
As I mentioned, it was my code which was failing. I did not have a time to dig around, yet it seems that the proposed change does not have much of an impact other than as documentation - which I still kind of like yet I see why you might not want to merge this.

Hence closing this issue. 👍

PR #1023, opened 2020-05-23
Multiarg infix syntax will be deprecated.

milessabin:
Accommodate upcoming deprecation of multiarg infix syntax which cause breakage as seen [here](https://travis-ci.org/github/milessabin/shapeless-test-scala-nightly/jobs/690235107).


joroKr21:
Did you use a scalafix?

milessabin:
> Did you use a scalafix?

No, just did it by hand.

PR #1028, opened 2020-06-13
add type annotation to public member

kazchimo:


milessabin:
`+validate` will test against all Scala platform/version permutations.

core/src/main/scala/shapeless/hlists.scala
```diff
@@ -62,9 +62,9 @@ object HList extends Dynamic {
   import ops.hlist._
   import syntax.HListOps
 
-  def apply() = HNil
+  def apply(): HList = HNil
```
milessabin, line 65: That needs to be either `HNil` or possibly `HNil.type`.

kazchimo:
@milessabin Executing `validate` in sbt shell cause linking errors.
How can I avoid these errors?

```
[info] Fast optimizing /Users/moriyamakazuki/development/Scala/shapeless/core/js/target/scala-2.13/shapeless-test-fastopt.js
[error] Referring to non-existent class shapeless.HListTests$Fruit
[error]   called from shapeless.HListTests$scalajs$junit$bootstrapper$.newInstance()java.lang.Object
[error]   called from private org.scalajs.junit.JUnitTask.$anonfun$executeTestMethod$2(org.scalajs.junit.Bootstrapper)java.lang.Object
[error]   called from private org.scalajs.junit.JUnitTask.$anonfun$executeTestMethod$1(org.scalajs.junit.Bootstrapper)scala.util.Try
[error]   called from private org.scalajs.junit.JUnitTask.executeTestMethod(org.scalajs.junit.Bootstrapper,org.scalajs.junit.TestMetadata,org.scalajs.junit.Reporter)scala.concurrent.Future
...
[error] There were linking errors
[error] (coreJS / Test / fastOptJS) There were linking errors
```

kazchimo:
To fix this errors I delete `core/js/target` directory and recompile, but it doesn't work.

milessabin:
`+validate` should complete successfully ... what platform are you on?

One thing to try: make sure you've committed your changes and then `git clean -dxf`.

kazchimo:
I develop on the mac os.

> One thing to try: make sure you've committed your changes and then git clean -dxf.

Thank you for the advise, but it didn't work.

In the first place, should I execute `sbt +vlaidate` not `sbt validate`?
If so `sbt +validate` resulted in below, and is it the ideal behaivor?

```
sbt:shapeless> +validate
[warn] Issuing a cross building command, but not all sub projects have the same cross build configuration. This could result in subprojects cross building against Scala versions that they are not compatible with. Try issuing cross building command with tasks instead, since sbt will be able to ensure that cross building is only done using configured project and Scala version combinations that are configured.
[info] Reapplying settings...
[info] Set current project to shapeless (in build file:mypath/shapeless/)
```

The errors I described above are occurred when I execute `sbt valdiate`.
`sbt validate` command tested against jvm, native and js, and was successful in jvm and native.
The errors for the js, do I have to succeed in js on local machine?

joroKr21:
I also get the errors on Mac. I think it has to do with case sensitivity of the file system.

joroKr21:
Maybe it is this bug: scala/bug#11997

joroKr21:
@kazchimo sorry I forgot to ask in your other PR, would you mind adding yourself to the contributors list?

kazchimo:
@joroKr21 yes, i'd love it.

kazchimo:
@milessabin I could resolve above error on #1029 thanks to information @joroKr21 gave. This PR seems to have no more issue.

joroKr21:
Oh I was waiting for you to add yourself to the contributors list in the README.
Could you also squash your commits into one?

kazchimo:
@joroKr21 sorry, I've misunderstood your words.

@milessabin @joroKr21 
I squashed commits and edit contributors info.
Thanks for your help!

PR #1030, opened 2020-06-22
refined if statement

kazchimo:
Firstly, `Symbol.isTerm` is always true if `Symbol.isMethod` is true, so `Symbol.isTerm && Symbol.isMethod` can be rewrriten as `Symbol.isMethod`.
Secondly, `!Symbol.asTerm.isOverloaded` is always true if `Symbol.isMethod` is true, and `asTerm` won't throw an exception because it only throw when `isTerm` is false.


joroKr21:
Thanks, nice simplification 👍

kazchimo:
> TIL
> 
> > Secondly, !Symbol.asTerm.isOverloaded is always true if Symbol.isMethod is true

FYI, the reason why this can be said is as below.
In the doc of `isOverloaded` the relationship of those are described as "If `isOverloaded` is true, `isMethod` is false", then the statement that "if `isMethod` is true, `isOverloaded` is false" can be deduced.

PR #1031, opened 2020-06-26
TypeTrace: use `Context#info` instead of `println`.

fehu:
Reason:
When compiling with Bloop build server, the message emitted by `println` is not shown by default (bloop) reporter.


milessabin:
Before I merge, would you mind adding yourself to the contributors list?

fehu:
Done.
Glad to be of help.

milessabin:
Many thanks! Merging ... :-)

PR #1032, opened 2020-06-27
Add docs

kazchimo:
just add docs walking around the codebase with minor refactorings.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -288,6 +288,14 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
     normalized <:< hlistTpe || normalized <:< coproductTpe
   }
 
+  /**
+   * Lower the order of type kind applying Any type if `tpe` takes type parameters.
+   * Note that it works properly only if `tpe` take only one parameter.
+   *
+   * {{{
+   *   lowerKind(typeOf[List[_]].typeConstructor) -> List[Any]
+   * }}}
+   */
   def lowerKind(tpe: Type): Type =
     if(tpe.takesTypeArgs)
       appliedType(tpe, List(typeOf[Any])).dealias
@@ -328,6 +336,10 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
     nameStr.contains("$anon") || nameStr == "<refinement>"
   }
 
+  /**
+   * @return a List of name and type pairs for the fields of type `tpe`.
+   * @see [[isCaseAccessorLike]] for the definition of what is considered a field.
+   * */
   def fieldsOf(tpe: Type): List[(TermName, Type)] = {
     val clazz = tpe.typeSymbol.asClass
     val isCaseClass = clazz.isCaseClass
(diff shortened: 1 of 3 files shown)


core/src/main/scala/shapeless/generic.scala
```diff
@@ -288,6 +288,13 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
     normalized <:< hlistTpe || normalized <:< coproductTpe
   }
 
+  /**
+   * Lower the order of type kind applying Any type.
```
joroKr21, line 292: Now I notice that it works only for types with one type parameter, so maybe we should document that.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -328,6 +335,10 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
     nameStr.contains("$anon") || nameStr == "<refinement>"
   }
 
+  /**
+   * @return the List of pair of `tpe` case accessor like term's name and final result type of that term.
+   * @see [[isCaseAccessorLike]] for the definition of case accessor "like".
```
joroKr21, line 340: I find that sentence difficult to parse (non-native speaker here). Maybe something like this:
```suggestion
   * @return a List of name and type pairs for the fields of type `tpe`.
   * @see [[isCaseAccessorLike]] for the definition of what is considered a field.
```

core/src/main/scala/shapeless/generic.scala
kazchimo: thanks a lot!

kazchimo: @joroKr21 

```scala
import scala.reflect.runtime.universe._

def lowerKind(tpe: Type): Type =
  if(tpe.takesTypeArgs)
    appliedType(tpe, List(typeOf[Any])).dealias
  else tpe

lowerKind(typeOf[List[_]].typeConstructor)
lowerKind(typeOf[Map[_, _]].typeConstructor)
// val res0: reflect.runtime.universe.Type = List[Any]
// val res1: reflect.runtime.universe.Type = Map[Any]
```

I confirmed lowerKind behaviors locally, and is this result an ideal one?
Map is applied one type parameter, but this looks strange.

joroKr21: > Map is applied one type parameter, but this looks strange.

I have no clue what that means tbh

kazchimo: Sorry, I misunderstood the work of `appliedType`.
I didn't know `appliedType` can be used for any type like below.

```scala
appliedType(typeOf[String], typeOf[Any])
// val res0: Type = String[Any]
```

So, I thought `Map` should take 2 type parameters and Map with one type parameter looked strange to me.
I figure out how `appliedWorks` and the reason of `if(tpe.takesTypeArgs)` exists.

kazchimo: I added the comment about this topic, but feel free to say any point to be improved.

core/src/main/scala/shapeless/generic.scala
```diff
@@ -698,6 +723,8 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
     mkAttributedRef(pre, getter)
   }
 
+  /** Check if `sym` is annotated by [[nonGeneric]]
+   *  including more processes to avoid known issues about annotation and macro. */
```
joroKr21, line 727: Sorry I didn't notice this before. I think the callers of this method don't care about the workarounds for bugs.
```suggestion
  /**
   * Check if `sym` or any of its overrides are annotated by [[nonGeneric]]
   */
```

core/src/main/scala/shapeless/generic.scala
joroKr21: The other option would be to fix it instead:
```scala
def lowerKind(tpe: Type): Type =
    appliedType(tpe, tpe.typeArgs.map(_ => AnyTpe)).dealias
```

PR #1039, opened 2020-09-06
Update For Dotty 0.28.0-bin-20200904-1ab71c1-NIGHTLY

isomarcte:
* `Reporting.error` is now `report.error`
* `TypeOrBounds` was removed from dotty in [this commit](https://github.com/lampepfl/dotty/commit/6872bdbbc9b8028d73f3115b0c325ac8a57ea718).
    * That commit changed `TypeBounds` and `NoPrefix` from `scala.tasty.reflect.Reflection` to be subtypes of `Type`, where before they were not.
    * This has the consequence that pattern matches which were formerly on `TypeOrBounds` which care to discriminate on `NoPrefix` or `TypeBounds` must check these _before_ `Type`, e.g.
        ```scala
          (t: Type) match {
            case _: NoPrefix => ???
            case _: TypeBounds => ???
            case _: Type => ???
          }
        ```
* `AppliedType.apply` in dotty was replaced with an extension method on `Type` named `appliedTo`. Calls in Shapeless have been updated to reflect that.
* Updated the `sbt-dotty` plugin version
* Updated the `sbt` version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ jdk:
 - openjdk8
 
 scala:
-- 0.25.0-bin-20200609-a3b417b-NIGHTLY
+- 0.28.0-bin-20200904-1ab71c1-NIGHTLY
 
 cache:
   directories:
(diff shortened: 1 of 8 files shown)


milessabin:
@isomarcte many thanks for this, but I've actually merged @nicolasstucki's changes from the [dotty-staging fork](https://github.com/dotty-staging/shapeless).

isomarcte:
@milessabin :+1: I didn't realize @nicolasstucki had already got it.

PR #1058, opened 2020-11-18
Add support to extract all variable or type annotations (issue #1051)

### LINKED ISSUE #1051: Add support for type parameters in annotations
Follow up on issue #922 and pull request #925. I would like to add support for type parameters in annotations, both variable and type annotations. For example, define an annotation such as:
```scala
case class MyAnnotation[T](t: T) extends scala.annotation.StaticAnnotation
```

The out type of `Annotations` and `TypeAnnotations` would have the complete "reference" to type `T`.

pgrandjean:
Following issue #1051, adding support to extract all variable or type annotations from case classes.

Closes #1051

### DIFF
--- a/core/src/main/scala/shapeless/annotation.scala
+++ b/core/src/main/scala/shapeless/annotation.scala
@@ -441,7 +441,13 @@ class AnnotationMacros(val c: whitebox.Context) extends CaseClassMacros {
   def extract(tpe: Boolean, s: Symbol): List[c.universe.Annotation] = {
     if (tpe) {
       s.typeSignature match {
-        case a: AnnotatedType => a.annotations.reverse // compiler provides type annotations in reverse
+        case a: AnnotatedType => a.annotations.reverse
+        case c: ClassInfoType =>
+          val parents = c.parents         
+          parents.flatMap {
+            case a: AnnotatedType => a.annotations.reverse
+            case _ => Nil
+          }
         case _ => Nil
       }
     } else {
--- a/core/src/main/scala/shapeless/annotation.scala
+++ b/core/src/main/scala/shapeless/annotation.scala
@@ -172,6 +172,114 @@ object TypeAnnotations {
   implicit def materialize[A, T, Out <: HList]: Aux[A, T, Out] = macro AnnotationMacros.materializeTypeAnnotations[A, T, Out]
 }
 
+/**
+ * Provides all variable annotations for the fields or constructors of case class-like or sum type `T`.
+ *
+ * If type `T` is case class-like, this type class inspects its fields and provides their variable annotations. If
+ * type `T` is a sum type, its constructor types are looked for variable annotations as well.
+ *
+ * Type `Out` is an HList having the same number of elements as `T` (number of fields of `T` if `T` is case
+ * class-like, or number of constructors of `T` if it is a sum type). It is made of `HNil` (no annotations for corresponding
+ * field or constructor) or `HLists` (list of annotations for corresponding field or constructor).
+ *
+ * Method `apply` provides an HList of type `Out` made of `HNil` (corresponding field or constructor not annotated)
+ * or `HList` (corresponding field or constructor has annotations).
+ *
+ * Note that variable annotations must be case class-like for this type class to take them into account.
+ *
+ * Example:
+ * {{{
+ *   case class First(s: String)
+ *   case class Second(i: Int)
+ *
+ *   case class CC(i: Int, @First("a") @Second(0) s: String)
+ *
+ *   val ccFirsts = AllAnnotations[CC]
+ *
+ *   // ccFirsts.Out is  HNil :: (First :: Second :: HNil) :: HNil
+ *   // ccFirsts.apply() is
+ *   //   HNil :: (First("a") :: Second(0) :: HNil) :: HNil
+ *
+ * }}}
+ *
+ * This implementation is based on [[shapeless.Annotations]] by Alexandre Archambault.
+ *
+ * @tparam T: case class-like or sum type, whose fields or constructors are annotated
+ *
+ * @author Patrick Grandjean
+ */
+trait AllAnnotations[T] extends DepFn0 with Serializable {
+  type Out <: HList
+}
+
+object AllAnnotations {
+  def apply[T](implicit annotations: AllAnnotations[T]): Aux[T, annotations.Out] = annotations
+
+  type Aux[T, Out0 <: HList] = AllAnnotations[T] { type Out = Out0 }
+
+  def mkAnnotations[T, Out0 <: HList](annotations: => Out0): Aux[T, Out0] =
+    new AllAnnotations[T] {
+      type Out = Out0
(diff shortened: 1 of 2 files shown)


joroKr21:
That looks binary compatible - it could make it into a 2.3 release

milessabin (approved):
LGTM!

core/src/test/scala/shapeless/annotation.scala
```diff
@@ -175,4 +188,30 @@ class AnnotationTests {
     illTyped(" TypeAnnotations[Second, Dummy] ", "could not find implicit value for parameter annotations: .*")
   }
 
+  @Test
+  def allAnnotations: Unit = {
+    {
+      val all: (First :: HNil) :: shapeless.HNil :: (Second :: Third :: HNil) :: HNil = AllAnnotations[CC3].apply()
```
joroKr21, line 194: You can use `typed` instead of testing this twice

core/src/test/scala/shapeless/annotation.scala
```diff
@@ -175,4 +188,30 @@ class AnnotationTests {
...
+    {
+      val first = AllAnnotations[CC3].apply()
+      assert(first == (First() :: HNil) :: shapeless.HNil :: (Second(2, "b") :: Third('c') :: HNil) :: HNil)
+    }
+  }
+  
+  @Test
+  def allTypeAnnotations: Unit = {
+    {
+      val all: (First :: HNil) :: shapeless.HNil :: (Second :: Third :: HNil) :: HNil = AllTypeAnnotations[CC4].apply()
```
joroKr21, line 207: You can use `typed` instead of testing this twice

joroKr21 (changes requested):
@pgrandjean thank you, the implementation looks great 👍 

I am only missing tests for all annotations of sealed traits

core/src/test/scala/shapeless/annotation.scala
pgrandjean: First time using `typed`. Did some changes but not sure if this is what you have in mind.

core/src/test/scala/shapeless/annotation.scala
pgrandjean: same as previous

pgrandjean:
> @pgrandjean thank you, the implementation looks great 👍
> 
> I am only missing tests for all annotations of sealed traits

Also added tests for sealed trait, although there are no type annotations for sealed traits.

pgrandjean:
branch `issue_1051` rebased on `main`.

joroKr21:
I'm trying to pull that thread but I encountered an annoying issue - `dealias` behaves differently in the compiler and in macros 😕

pgrandjean:
> > Also added tests for sealed trait, although there are no type annotations for sealed traits.
> 
> Hmm it's technically possible:

You're right. I don't know why I was convinced of the opposite. Sorry for that.

`AllTypeAnnotations` fixed for sealed traits. In this case, the type annotations are attached to the parent class (`Base2`), so one needs to go one layer deeper in the AST.

joroKr21 (approved):
@pgrandjean that looks good to me already. Sorry to keep expanding the scope but I'm wondering if we should make something like this work, which seems to me like a more compelling use case for type annotations:
```scala
  final case class validated(id: String) extends StaticAnnotation
  type PosInt = Int @validated("positive") 
  type Email = String @validated("email")
  final case class User(email: Email, age: PosInt)
```

I tried it but it seems that `.dealias` sometimes preserves annotations and sometimes loses them. Anyway I'm happy to merge this as is, just suggesting what we could do next.

pgrandjean:
@joroKr21 I think it is possible to achieve this without `dealias`. Just pushed a change to the `extract` function to take into account type refs. Test added, seems to work. Please let me know if this implementation is OK :)

core/src/main/scala/shapeless/annotation.scala
```diff
@@ -290,10 +384,71 @@ class AnnotationMacros(val c: whitebox.Context) extends CaseClassMacros {
...
       s.typeSignature match {
-        case a: AnnotatedType => a.annotations
+        case a: AnnotatedType => a.annotations.reverse
+        case c: ClassInfoType =>
+          val parents = c.parents         
+          parents.flatMap {
+            case a: AnnotatedType => a.annotations.reverse
+            case _ => Nil
+          }
+        case t: TypeRef => extract(tpe, t.sym)
```
joroKr21, line 451: We should only follow aliases. And I suggest a bit of cleanup with a helper:

```scala
  def extract(tpe: Boolean, s: Symbol): List[c.universe.Annotation] = {
    def fromType(t: Type): List[c.universe.Annotation] = t match {
      case AnnotatedType(annotations, _) => annotations.reverse
      case ClassInfoType(parents, _, _) => parents.flatMap(fromType)
      case TypeRef(_, sym, _) if sym.asType.isAliasType => extract(tpe, sym)
      case _ => Nil
    }

    if (tpe) fromType(s.typeSignature)
    else s.annotations
  }
```

core/src/test/scala/shapeless/annotation.scala
```diff
@@ -172,7 +193,31 @@ class AnnotationTests {
...
+    val st = AllTypeAnnotations[Base2].apply() // sealed trait
+    typed[(First :: HNil) :: (Second :: Third :: HNil) :: HNil](st)
+
+    val cc = AllTypeAnnotations[CC4].apply() // case class
+    typed[(First :: HNil) :: HNil :: (Second :: Third :: HNil) :: HNil](cc)
+    assert(cc == (First() :: HNil) :: HNil :: (Second(2, "b") :: Third('c') :: HNil) :: HNil)
+    
+    val user = AllTypeAnnotations[User].apply() // type refs
+    typed[(First :: HNil) :: (Third :: HNil) :: HNil](user)
+    assert(user == (First() :: HNil) :: (Third('c') :: HNil) :: HNil)
```
joroKr21, line 220: That's awesome!

joroKr21 (changes requested):
One last thing 🙏

core/src/main/scala/shapeless/annotation.scala
pgrandjean: sure. done.

core/src/test/scala/shapeless/annotation.scala
pgrandjean: Agreed!

joroKr21 (approved):

joroKr21:
Thank you @pgrandjean for your work and patience and congratulations on your first merge to shapeless! 🥳

pgrandjean:
Thank you @joroKr21 !

joroKr21:
@pgrandjean wow I completely forgot about #925, sorry!

GitHub lied to me and told me you're a first time contributor - it even asked me to approve your CI run 🙈 
But having a bad memory is my own fault 😄 so I guess congratulations on your second PR 👍 

I will try to backport them to the 2.3 branch soon, I think there will be no issues 🤞

PR #1062, opened 2020-12-01
Treat expected string as literal pattern

user501254:
This makes it easier to type-check against error strings that contain regex sensitive characters.
Without this, someone would have to manually escape such characters.

### DIFF
--- a/core/src/main/scala/shapeless/test/typechecking.scala
+++ b/core/src/main/scala/shapeless/test/typechecking.scala
@@ -43,7 +43,7 @@ class IllTypedMacros(val c: blackbox.Context) {
     val (expPat, expMsg) = expected match {
       case null => (null, "Expected some error.")
       case Literal(Constant(s: String)) =>
-        (Pattern.compile(s, Pattern.CASE_INSENSITIVE | Pattern.DOTALL), "Expected error matching: "+s)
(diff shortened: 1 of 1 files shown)


user501254:
If such matching was intentional, would you consider adding another flag to disable such behavior?

joroKr21:
The regex pattern matching is intentional 👍 
I'm not sure what you mean by a separate flag.

user501254:
By adding another `apply`:

```
object illTyped {
  def apply(code: String): Unit = macro IllTypedMacros.applyImplNoExp
  def apply(code: String, expected: String): Unit = macro IllTypedMacros.applyRegex
  def apply(code: String, expected: String, useRegex: Boolean): Unit = macro IllTypedMacros.applyImpl
}

class IllTypedMacros(val c: blackbox.Context) {...}
```

milessabin:
Sorry about this ... it appears that renaming `master` to `main` has resulting in this PR being closed automatically. That's not supposed to happen.

But I think this might have ground to a halt anyway. If that's not the case, please reopen and target at `main`.

joroKr21:
FWIW what munit does is provide a macro that gives you the error message as a string and you can assert on it however you wish: https://scalameta.org/munit/docs/assertions.html#compileerrors

PR #1073, opened 2021-01-29
scala-native 0.4.0

xuwei-k:


joroKr21:
This doesn't quite work for master, because the matrix needs to be update to include all Scala versions, not just 2.11.
I'm also trying to restore a backwards compatible state of shapeless on the [series/2.3 branch](https://github.com/milessabin/shapeless/tree/series/2.3)

joroKr21:
v2.3.4 with Scala Native support is out 🥳 https://github.com/milessabin/shapeless/releases/tag/v2.3.4

PR #1075, opened 2021-01-29
Workaround lampepfl/dotty#11250

OlivierBlanvillain:
Workaround https://github.com/lampepfl/dotty/issues/11250


milessabin:
Interesting! And bizarre!

Why do the equivalent types that I define behave properly wrt separate compilation, but the stdlib ones don't?

milessabin:
Actually, even with these changes the tests don't compile separately.

milessabin:
Right, it seems that there's no difference between my definitions and the stdlib ones in this respect, ie. mine will work iff the stdlib ones do. That makes more sense.

However that means that this PR isn't actually a workaround for the issue.

OlivierBlanvillain:
I thought sbt was compiling /main and /test separately, isn't that the case? How/where did you notice the issue after this change?

milessabin:
> I thought sbt was compiling /main and /test separately, isn't that the case?

It is the case.

> How/where did you notice the issue after this change?

With your branch: you left the tests in main and the tests commented out in test. As soon as I removed the tests from main and uncommented the ones in test the same failure recurred.

I've just pushed that change to this branch ... you can see it failing as before.

OlivierBlanvillain:
Makes sense. Is there a dotty version where this pattern compiles? It would helpful to able to git bisect.

milessabin:
> Is there a dotty version where this pattern compiles? It would helpful to able to git bisect.

An earlier version worked prior to https://github.com/lampepfl/dotty/pull/7389 but it wouldn't be a like for like comparison because most of these match type had to be reworked after than PR.

milessabin:
Can you explain what's going on in these changes?

OlivierBlanvillain:
Yes, I was waiting on the CI output to make that it passes. The last commit is an actual workaround (unlike my previous attempt were I forgot to uncomment the problematic tests). The issue on the dotty side seems to be related to an infinite loop in find members when looking for `pairs.to`. Inlining the definition of these types to avoid going thought inheritance seems to be sufficient to doge the issue.

After looking into it on the dotty side I don't think I will be able to fix the issue in a reasonable time frame, and given that we should aim at removing the temporary `-Y` flag for the RC, I propose to fix the issue on the shapeless side where is seems relatively easy.

julienrf:
@milessabin This change seems necessary to get Shapeless 3 work with Scala 3.0.0-RC1. Do you intend to merge it or do you think we should look for a better alternative?

OlivierBlanvillain:
It's not needed anymore, lampepfl/dotty#11250 has been fixed and shapeless3 compiles as is (I forgot to close the PR, sorry about that)

PR #1077, opened 2021-02-10
Update Scala Native to 0.4.0

lolgab:
- Update Scala 2.13 to 2.13.4
- Update Scala 2.12 to 2.12.13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 os: linux
-dist: trusty
+dist: bionic
 language: scala
 
 jdk:
-  - oraclejdk8
+  - openjdk8
 
 scala:
   - 2.11.12
-  - 2.12.12
-  - 2.13.3
+  - 2.12.13
+  - 2.13.4
 
 env:
   - {}
@@ -39,12 +39,9 @@ jobs:
     - stage: test
       env:
         - SCALA_NATIVE=true
-      before_install:
-        - curl https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/travis_setup.sh | bash -x
     - stage: release
       before_install:
         - git fetch --tags
-        - curl https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/travis_setup.sh | bash -x
       script:
         - sbt ci-release
         - SCALA_JS_VERSION=0.6.33 sbt clean sonatypeBundleClean ci-release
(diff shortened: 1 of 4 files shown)


joroKr21:
I think we need a separate branch for 2.3 compatible changes. The current state of master is not binary compatible and it's both not something we can release yet and also holding back backwards compatible changes. I will look into setting up a separate branch soon.

xuwei-k:
🤔 👀 https://github.com/milessabin/shapeless/pull/1073

lolgab:
@xuwei-k Sorry when I opened the PR I didn't notice your branch.
By the way, you forgot to add Scala 2.12 and 2.13 to Scala Native's `crossScalaVersions`.

lolgab:
@joroKr21 Can I do something to help? Shapeless is a blocker for the typelevel ecosystem to be ported to Scala Native.

joroKr21:
I got started but hit some unexpected problems. The idea is simple - start with the 2.3.3 tag and on top:
 * copy the build
 * copy the Travis CI config
 * fix any errors
 * enable MiMa to ensure binary compatibility 
 * add Scala native
 * (optional) drop Scala 2.11 and Scala.js 0.6

Unfortunately the 2.3.3 tag doesn't cross compile cleanly. Maybe @milessabin remembers why?

So I'm still at the step trying to fix errors. I will push whatever I have to a branch later today.

joroKr21:
Check the [series/2.3.3](https://github.com/milessabin/shapeless/tree/series/2.3.3) branch

milessabin:
Unfortunately the problem is the 2.13.x collections changes. I was able to make a source compatible 2.3.3 for 2.13.x which was fine for back publishing because there wasn't a previous 2.13.x artefact. But although that cross-compiled for 2.10/11/12 it did so in a binary incompatible way.

This is awkward now, because if we wanted a 2.3.4 which is binary compatible with 2.3.3 we'd to rework how things are done for 2.11/12 to maintain binary compatibility, and that means either more variant source or two series/2.3 branches. I'm not super-keen on either.

A 2.3.4 for 2.13.x would be reasonably straightforward, as would a non-binary compatible 2.4.0 for 2.11/12/13. What are peoples preferences?

joroKr21:
I would be in favour of more variant sources on a single series/2.3 branch so that we have the freedom to backport other binary compatible changes.

milessabin:
I've made some progress on [series/2.3](https://github.com/milessabin/shapeless/tree/series/2.3). There are still some bincompat issues to resolve due to the 2.13.x collections changes.

joroKr21:
v2.3.4 with Scala Native support is out 🥳 https://github.com/milessabin/shapeless/releases/tag/v2.3.4

PR #1090, opened 2021-03-12
Update Scala to 3.0.0-RC1

julienrf:
I think we should not let the implementations in `dotty-staging` and this one diverge much.

I cherry-picked their commits, and I added a couple of more commits to clean things up.


modules/data/src/main/scala/shapeless3/data/monoidal.scala
```diff
@@ -18,11 +18,9 @@ package shapeless3.data
 
 import scala.compiletime._
 
-import Monoidal._
-
 trait Monoidal {
-  type to[_]
-  type from[_]
+  type to[_] <: Tuple
+  type from[_ <: Tuple]
```
joroKr21, line 23: Does it only work with `Tuple` now? 🤔

julienrf: I’m not sure, this change was applied in commit 5c4777fbe0f0ee29f1f048acaac300f41df315bf, maybe @milessabin knows more?

milessabin:
Thanks for making a start on this ... I've got some time to move things along next week.

modules/data/src/main/scala/shapeless3/data/monoidal.scala
milessabin: Tuples are now used as the canonical representation ... there doesn't seem to be any obvious reason to invent another type to play that role.

joroKr21: Ok for tuples probably not, but for `Either` the jury is still out I think 🤔

milessabin:
This looks great ... I'm going to merge as-is, then follow up with an additional tweak from @smarter, dependency updates and switch to github actions.

modules/data/src/main/scala/shapeless3/data/monoidal.scala
milessabin: No, this is fine for `Either` as well ... see [here](https://github.com/milessabin/shapeless/blob/shapeless-3/modules/data/src/main/scala/shapeless3/data/monoidal.scala#L141-L164).

PR #1095, opened 2021-03-19
Publish snapshots

milessabin:


joroKr21 (approved):

build.sbt
```diff
@@ -40,7 +40,7 @@ ThisBuild / githubWorkflowBuild := Seq(
 )
 
 ThisBuild / githubWorkflowTargetTags ++= Seq("v*")
-ThisBuild / githubWorkflowPublishTargetBranches :=
+ThisBuild / githubWorkflowPublishTargetBranches +=
```
joroKr21, line 43: 😄 
```suggestion
ThisBuild / githubWorkflowPublishTargetBranches ++=
```

milessabin: Yeah, I borked that completely :roll_eyes:

milessabin:
Unclear why that didn't work. I haven't yet set up the secrets, so I was expecting this to try to publish and fail, but instead the job seems to have simply been skipped. I wonder if the job is implicitly conditional on the secrets being present?

joroKr21:
> Unclear why that didn't work. I haven't yet set up the secrets, so I was expecting this to try to publish and fail, but instead the job seems to have simply been skipped. I wonder if the job is implicitly conditional on the secrets being present?

Ahh, it's a `master` vs `main` thing - we are still using `master` and `sbt-github-actions` generated `main`.

milessabin:
Riiight ... that's also on my TODO list, but I wanted to do one thing at a time.

PR #1098, opened 2021-03-19
Switch to github actions

milessabin:


joroKr21 (approved):

joroKr21:
AFAIU artifact upload is to speed up publishing, is that right?

joroKr21 (approved):

milessabin:
> AFAIU artifact upload is to speed up publishing, is that right?

Yup. It seems to be very fragile right now.

PR #970, opened 2020-02-28
Replace anonymous classes created by macros with instance helpers

joroKr21:
This reduces the size of the generated bytecode and
the number of emitted classfiles (on Scala 2.12+).

Thanks to @neko-kai for pointing out this inefficiency.

 - [x] `Generic`
 - [x] `DefaultSymbolicLabelling`
 - [x] `Witness`
 - [x] `WitnessWith`
 - [x] `NatWith`
 - [x] `SingletonOps`
 - [x] `Generic1`
 - [x] `Split1`
 - [x] `IsHCons1`
 - [x] `IsCCons1`

`Annotations` and `Default` already have helpers.

PR #1107, opened 2021-04-13
Replace anonymous classes created by macros with instance helpers

joroKr21:
This reduces the size of the generated bytecode and
the number of emitted classfiles (on Scala 2.12+).
Also simplidies the macro implementations.

Backports #970

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,11 +65,11 @@ addCommandAlias("core", ";project coreJVM")
 addCommandAlias("scratch", ";project scratchJVM")
 addCommandAlias("examples", ";project examplesJVM")
 
-addCommandAlias("validate", ";root;validateJVM;validateJS;validateNative")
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -315,7 +316,7 @@ lazy val mimaSettings = Seq(
   mimaPreviousArtifacts := Set(organization.value %% moduleName.value % "2.3.3"),
   mimaBinaryIssueFilters := {
     // Macro internals - ignore
-    val macroFilters = List("CaseClassMacros", "IsCons1Macros", "LazyMacros")
+    val macroFilters = List("CaseClassMacros", "IsHCons1Macros", "IsCCons1Macros", "IsCons1Macros", "LazyMacros", "SingletonTypeMacros")
```
joroKr21, line 319: I wonder if it's worth it to try and preserve binary compatibility here, I didn't make any effort

milessabin (approved):
LGTM!

PR #1108, opened 2021-04-15
Coproduct instances expose ordinal

TimWSpence:
Expose ordinals in `CoproductInstances#fold2` for the sake of eg deriving `Ord`

### DIFF
--- a/modules/deriving/src/test/scala/shapeless3/deriving/deriving.scala
+++ b/modules/deriving/src/test/scala/shapeless3/deriving/deriving.scala
@@ -93,7 +93,7 @@ class DerivationTests {
     assert(v2.compare(SomeInt(0), SomeInt(0)) == 0)
     assert(v2.compare(SomeInt(0), SomeInt(1)) == -1)
     assert(v2.compare(SomeInt(1), SomeInt(0)) == 1)
-    assert(v2.compare(SomeInt(0), NoneInt) == -1)
+    assert(v2.compare(SomeInt(0), NoneInt) == 1)
 
     val v3 = Ord[NoneInt.type]
     assert(v3.compare(NoneInt, NoneInt) == 0)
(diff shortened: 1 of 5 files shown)


modules/deriving/src/test/scala/shapeless3/deriving/deriving.scala
```diff
@@ -78,6 +78,48 @@ class DerivationTests {
...
+    assert(v1.compare(Box(1), Box(1)) == 0)
+    assert(v1.compare(Box(1), Box(0)) == 1)
+    assert(v1.compare(Box(0), Box(1)) == -1)
+
+    val v2 = Ord[OptionInt]
+    assert(v2.compare(NoneInt, NoneInt) == 0)
+    assert(v2.compare(SomeInt(0), SomeInt(0)) == 0)
+    assert(v2.compare(SomeInt(0), SomeInt(1)) == -1)
+    assert(v2.compare(SomeInt(1), SomeInt(0)) == 1)
+    assert(v2.compare(SomeInt(0), NoneInt) == -1)
```
milessabin, line 96: I guess I find this ordering a bit strange.

milessabin (approved):
I find the ordinal based ordering a bit arbitrary ... is that really how it's done in Haskell? `Just(1)` < `Nothing`?

Aside from that this all looks great :-)

TimWSpence:
> I find the ordinal based ordering a bit arbitrary ... is that really how it's done in Haskell? `Just(1)` < `Nothing`?
> 
> Aside from that this all looks great :-)

Yeah, they declare the data constructors in the opposite order so you get the more natural ordering. I've done the same here now

modules/deriving/src/test/scala/shapeless3/deriving/deriving.scala
joroKr21: It's definitely better than alphabetic ordering 😜

milessabin: :roll_eyes:

PR #1111, opened 2021-04-16
Remove unnecessary macrocompat annotations

joroKr21:


milessabin (approved):
LGTM!

milessabin:
Oh! Actually, [this](https://github.com/milessabin/shapeless/blob/series/2.3/core/src/main/scala_2.13-/shapeless/versionspecifics.scala#L47-L49) can go too ...

joroKr21:
> Oh! Actually, this can go too ...

MiMa complains about that

milessabin:
It's `private[shapeless]` though ... surely that's safe to remove?

joroKr21:
> It's private[shapeless] though ... surely that's safe to remove?

We would have to add filters then. I'd rather keep it. WDYT?

milessabin:
Why keep it? It's not accessible externally and you've removed all the internal uses ... surely it's just dead weight now?

joroKr21:
> Why keep it? It's not accessible externally and you've removed all the internal uses ... surely it's just dead weight now?

It's less code / effort to keep the class than to add the filters. But ok you convinced me that it's better to remove it.

PR #1115, opened 2021-04-21
Removed allocation in object Tag

joroKr21:
Backport #705


joroKr21:
I'm gonna fast track some of these backports to save on time. We can always revert or fixup later.

PR #1131, opened 2021-04-24
Expose CaseClassMacros as blackbox

joroKr21:
`blackbox.Context` is a superclass of `whitebox.Context` so it's ok.

Backports #977

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -295,7 +295,7 @@ trait ReprTypes {
 }
 
 trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
-  val c: whitebox.Context
+  val c: blackbox.Context
 
   import c.universe._
   import internal.constantType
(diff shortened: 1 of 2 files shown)


core/src/main/scala/shapeless/generic.scala
```diff
@@ -295,7 +295,7 @@ trait ReprTypes {
 }
 
 trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
-  val c: whitebox.Context
+  val c: blackbox.Context
```
xuwei-k, line 298: 😢 

https://github.com/limansky/beanpuree/issues/11
https://github.com/limansky/beanpuree/blob/e86098b9674805966b3b75d62351a61ad1a50b61/src/main/scala/me/limansky/beanpuree/BeanUtils.scala#L21-L22

joroKr21: @xuwei-k we've broken binary compatibility for macros in many ways. Unfortunately it's too difficult to maintain. At least the error you get is a compile time error so as soon as the other library which is using macros updates it will work again.

PR #1144, opened 2021-04-28
Derive traverse

TimWSpence:


modules/deriving/src/main/scala/shapeless3/deriving/erased.scala
```diff
@@ -124,6 +129,21 @@ final class ErasedProductInstancesN[K, FT](val mirror: Mirror.Product, is: Array
...
+    if (n == 0) x0
+    else {
+      val x = toProduct(x0)
+      var acc = pure(EmptyTuple)
+      var i = 0
+      while(i < n) {
+        acc = ap(map(acc, (x: Tuple) => (y: Any) => x ++ Tuple1(y)), f(is(i), x.productElement(i)))
+        i = i+1
+      }
+      ap(pure(x => mirror.fromProduct(x)), acc)
```
joroKr21, line 143: @TimWSpence shouldn't we be using `ArrayProduct` here?

PR #1153, opened 2021-05-07
Restore macro binary compatibility with 2.3.3

joroKr21:
~I thought this would be too hard, but actually it was not.~ CC @xuwei-k 
Well, it is quite difficult in fact - see the CI failures.


milessabin (approved):
LGTM!

PR #1158, opened 2021-05-08
Reduce 2.12 / 2.13 Jar size by using instance constructors 

joroKr21:
Generate less anonymous classes for implicit instances,
esp. for tuple and function instances which are code generated.

Note: this slightly increases the Jar size on 2.11 because it doesn't translate to `LambdaMetaFactory`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,4 @@
 import com.typesafe.sbt.SbtGit.GitKeys._
-import com.typesafe.tools.mima.core.ProblemFilters._
 import com.typesafe.tools.mima.core._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 import sbtcrossproject.CrossProject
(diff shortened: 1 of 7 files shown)


milessabin (commented):
Retrospectively ... LGTM!

catostrophe:
@joroKr21 this probably breaks bincompat.

Moving `Aux` from the `*Instances` trait to the object led to:
```
[error] Symbol 'type shapeless.ops.FnFromProductInstances.Aux' is missing from the classpath.
[error] This symbol is required by 'value io.circe.generic.decoding.IncompleteDerivedDecoders.ffp'.
```

joroKr21:
@catostrophe that is surprising to me - I thought type aliases are erased and moving them around doesn't break binary compatibility. Also MiMa didn't complain. So maybe something more going on. Will have to investigate ...

joroKr21:
I opened lightbend/mima#627
The fix is obvious - we will have to move the type aliases back to the traits

catostrophe:
Oh no. The whole `circe` family has been released with shapeless 2.3.6. It just breaks in runtime with 2.3.7.

catostrophe:
@joroKr21 could you add a comment to 2.3.6 release notes that it shouldn't be used?

PR #1161, opened 2021-05-10
Reduce Jar size by using instance constructors

joroKr21:
Generate less anonymous classes for implicit instances,
esp. for tuple and function instances which are code generated.

Forward port #1158


milessabin (approved):
LGTM!

PR #1162, opened 2021-05-11
Fix subtle pickle compatibility problem with type aliases

joroKr21:
Move `Aux` type aliases back to the parent traits of
`Tupler`, `FnFromProduct` and `FnToProduct`.
Also bump MiMa version to 0.9.0.


milessabin (commented):
LGTM!

PR #1167, opened 2021-05-15
Revert source breaking change in TupleOps

joroKr21:
Users might be using `TupleOps` in a generic context where they have `<: Product` but not `: IsTuple`.
Propagating the bound might force downstream libraries to break binary compatibility which is not good.

Reverts #1132


milessabin (approved):
LGTM!

PR #1172, opened 2021-05-17
Optimize erasedTraverse to use ArrayProduct

joroKr21:


joroKr21:
Hmm looks like we need to update the repo settings for 3.0.0

TimWSpence:
Apparently I'm much too casual about allocations 😂 Thanks Miles! 👍

TimWSpence (approved):
LGTM!

PR #1174, opened 2021-05-20
The shapeless-3 README includes an incorrect build.sbt dependency

adamnfish:
The README docs for the sbt dependency is missing the Scala version. Changing `%` to `%%` corrects this problem.

Before:

```
[warn] 	Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last update for the full output
[error] (update) sbt.librarymanagement.ResolveException: Error downloading org.typelevel:shapeless3-deriving:3.0.0
[error]   Not found
[error]   Not found
[error]   not found: /Users/username/.ivy2/localorg.typelevel/shapeless3-deriving/3.0.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/typelevel/shapeless3-deriving/3.0.0/shapeless3-deriving-3.0.0.pom
```

After
```
[info] Fetching artifacts of
https://repo1.maven.org/maven2/org/typelevel/shapeless3-deriving_3/3.0.0/shapeless3-deriving_3-3.0.0.jar
  100.0% [##########] 115.9 KiB (283.3 KiB / s)
```

PR #1175, opened 2021-05-20
Fix dependency example in Readme

ybasket:
The `%` instead of `%%` lets dependency resolution fail.


joroKr21:
Thanks, but fixed with #1174

PR #1176, opened 2021-05-21
Update the Monoid example in the readme

cb372:
* Replace `@alpha` with `@targetName`
* The syntax for implementing a trait with extension methods has changed
* You have to repeat the `@targetName` annotation for each implementation, which is a bit annoying

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ Using shapeless 3 the derivation of a monoid for a Scala ADT is as simple as,
 
 ```scala
 import shapeless3.deriving.*
-import scala.annotation.targetName
 
 // Type class definition, eg. from Cats
 trait Monoid[A]:
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ import scala.annotation.targetName
 trait Monoid[A]:
   def empty: A
   extension (x: A)
-    @targetName("combine") def |+| (y: A): A
+    def combine(y: A): A
 
 object Monoid:
   inline def apply[A](using ma: Monoid[A]): Monoid[A] = ma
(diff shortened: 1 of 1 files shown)


milessabin:
This is great ... thanks!

milessabin:
Though actually, I think this change adds a lot of noise to the example. I think I'd rather just drop the symbolic notation and have plain `def combine`.

milessabin (changes requested):
Though on reflection, I think this change adds a lot of noise to the example. I think I'd rather just drop the symbolic notation and have plain `def combine`.

joroKr21:
> You have to repeat the @targetName annotation for each implementation, which is a bit annoying

That is unfortunate. Is it by design or should we open a ticket for it?

cb372:
>Is it by design or should we open a ticket for it?

It appears to be by design, as it results in a pretty clear error message:

```
[error] 18 |      def |+|(y: Unit): Unit = ()
[error]    |          ^
[error]    |error overriding method |+| in trait Monoid of type (x: Unit)(y: Unit): Unit;
[error]    |  method |+| of type (x: Unit)(y: Unit): Unit misses a target name annotation @targetName(combine)
```

cb372:
>drop the symbolic notation

I thought you might say that 😄 

Updated.

milessabin:
Thanks for that, but honestly, I think we should go back to what I had originally, which didn't use extension methods at all. The example is to illustrate the derivation, not the new type class syntax.

It would be lovely if you had time for that, but if not I'll do it myself over the weekend.

milessabin:
Actually, leave it with me ... thanks again for flagging up the problem.

milessabin:
@cb372 continued here: https://github.com/milessabin/shapeless/pull/1177

Do you approve?

PR #1177, opened 2021-05-21
Fixed README Monoid example

milessabin:
Slightly less noisy version of https://github.com/milessabin/shapeless/pull/1176.


cb372 (approved):
👏

PR #1182, opened 2021-05-31
Optimize sized.ToHList and ToSizedHList

### LINKED ISSUE #1181: performance issue with combined `Length` and `ToSizedHList` implicit derivation
There seems to be a performance problem with the combined implicit derivation of `Length` and `ToSizedHList`. When `n` grows, the compilation time becomes quickly unpractical. I reproduced the problem with shapeless 2.3.7 and scala 2.12.14/2.13.6 . 

```scala
  val n = Nat(100)
  val f = Fill[n.N, String]

  def test[N <: Nat](implicit l: Length.Aux[f.Out, N], v: ToSizedHList[List, String, N]): Unit = {}

  // perf issue 
  test

  // no perf issue
  test[n.N]
  val l = Length[f.Out]
  ToSizedHList[List, String, l.Out]
  ```

joroKr21:
* Use type parameters and `Aux` instead of path-dependent types
 * Optimize runtime by replacing direct recursion with `foldRight`

Fixes #1181

### DIFF
--- a/core/src/main/scala/shapeless/ops/sizeds.scala
+++ b/core/src/main/scala/shapeless/ops/sizeds.scala
@@ -17,6 +17,8 @@
 package shapeless
 package ops
 
+import shapeless.ops.hlist.Fill
+
 object sized {
   /**
    * Type class supporting conversion of this `Sized` to an `HList` whose elements have the same type as in `Repr`.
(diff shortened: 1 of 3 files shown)


milessabin (approved):
LGTM!

PR #1184, opened 2021-06-01
Optimize sized.ToHList and ToSizedHList

joroKr21:
* Use type parameters and `Aux` instead of path-dependent types
 * Optimize runtime by replacing direct recursion with `foldRight`

Forward port #1182


milessabin (approved):
LGTM!

PR #1191, opened 2021-07-03
Fix Generic instance for case classes with implicit arguments

### LINKED ISSUE #1189: Generic in shapeless 2.3.5+ not working for classes with context bounds if some implicit value is present
Hi, I am using shapeless in my project and met some strange behavior. I am not sure if it is a bug.

The following code is a minimal example. It compiles with shapeless 2.3.5+ only if the instance is commented out. It works well with shapeless 2.3.3/2.3.4.
```
import shapeless._

trait ATypeClass[T]

case class AnUnrelatedCaseClass[T](v: T)

case class ACaseClassWithContextBound[T: ATypeClass]()

object Main extends App {
  implicit def instance[T: ATypeClass]: ATypeClass[AnUnrelatedCaseClass[T]] = new ATypeClass[AnUnrelatedCaseClass[T]] {}

  /*
   * With shapeless 2.3.5+, the following function compiles only if instance is commented out
   * With shapeless 2.3.3/2.3.4, the following function always compiles
   */
  def compilesOnlyIfInstanceIsCommented[T: ATypeClass]
  : Generic.Aux[ACaseClassWithContextBound[T], HNil] = Generic[ACaseClassWithContextBound[T]]
}
```

lsrcz:
The pull request fixed #1189 and added the test case referred in that issue.

Root cause of the bug:
The materialized `Generic` instances calls the CompanionObject's `apply` method without type parameters in the constructor. If the scala compiler cannot infer some type parameter, it cannot collect the correct implicit arguments.

Fix:
Let the constructor call the `apply` method with type parameters.

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -874,7 +874,7 @@ trait CaseClassMacros extends ReprTypes with CaseClassMacrosVersionSpecifics {
           }
         val nonCaseParamLists: List[List[Tree]] = List.fill(numNonCaseParamLists(tpe))(Nil)
         new CtorDtor {
-          def construct(args: List[Tree]): Tree = q"${companionRef(tpe)}(...${args :: nonCaseParamLists})"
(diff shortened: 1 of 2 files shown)


joroKr21 (approved):
Very nice fix, thank you!

PR #1197, opened 2021-08-01
Remove unnecessary asInstanceOf call

### LINKED ISSUE #1194: Generic.Aux compiles but errors with ClassCastException at runtime
Scala: 2.13.5
Shapeless: 2.3.7

I'm pretty positive that the following code used to not compile in Shapeless (note the private constructor and different return type of the companion apply):

```
case class Foo private(value: String)
object Foo {
    def apply(v: String): Either[String, Foo] = Left("No ways")
}
val gen: Generic.Aux[Foo,String :: HNil] = implicitly[Generic.Aux[Foo, String :: HNil]]
gen.from("test" :: HNil)
```

However, with the aforementioned versions, this does compile, but the last line will blow up at runtime with:

```"java.lang.ClassCastException: class scala.util.Left cannot be cast to class repl.MdocSession$App$Foo (scala.util.Left is in unnamed module of loader java.net.URLClassLoader @22fec1e1; repl.MdocSession$App$Foo is in unnamed module of loader 
scala.reflect.internal.util.AbstractFileClassLoader @43317cc1)
  at repl.MdocSession$App$$anonfun$2.apply(json.worksheet.sc:22)
  at repl.MdocSession$App$$anonfun$2.apply(json.worksheet.sc:22)
  at shapeless.Generic$$anon$1.from(generic.scala:164)
```

Note the above code will work if you make constructor public and remove the companion apply.

joroKr21:
Looks like it was necessary for some older version of Scala 2.13

Fixes #1194

### DIFF
--- a/core/src/main/scala/shapeless/generic.scala
+++ b/core/src/main/scala/shapeless/generic.scala
@@ -1066,7 +1066,7 @@ class GenericMacros(val c: whitebox.Context) extends CaseClassMacros {
     val (p, ts) = ctorDtor.binding
     val to = cq"$p => ${mkHListValue(ts)}.asInstanceOf[$repr]"
     val (rp, rts) = ctorDtor.reprBinding
-    val from = cq"$rp => ${ctorDtor.construct(rts)}.asInstanceOf[$tpe]"
+    val from = cq"$rp => ${ctorDtor.construct(rts)}"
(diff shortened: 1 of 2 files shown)


milessabin (approved):
LGTM!

PR #1198, opened 2021-08-01
Remove unnecessary asInstanceOf call

joroKr21:
Looks like it was necessary for some older version of Scala 2.13

Forward port #1197


milessabin (approved):
LGTM!

PR #1239, opened 2022-02-11
Merge pull request #1238 from joroKr21/default-access

### LINKED ISSUE #1099: `Default.AsRecord` causes `StackOverflow` in a path-dependent type's companion object
I'm not sure if this is expected behavior or not, but it caused a bug deep within my code base that wasn't intuitive to me. The following code throws a `StackOverflowError`.

```
import shapeless._

class DefaultsTest {
  final case class A(a: Int = 9)

  object A {
    val d = Default.AsRecord[A].apply()
  }
}

object DefaultsTest {
  def main(args: Array[String]): Unit = {
    val c = new DefaultsTest
    println(c.A.d)
  }
}
```

It works the way I expect if:
- `d` is declared as a `def` instead of a `val`
- `A` is not path-dependent
- `A.a` does not have a default value
- `d` is declared outside of `A`'s companion object

I'm using shapeless 2.3.3 on Scala 2.12.12.

joroKr21:
Access nested defaults through this to avoid stack overflow

Forward ports #1238
Fixes #1099

### DIFF
--- a/core/src/main/scala/shapeless/default.scala
+++ b/core/src/main/scala/shapeless/default.scala
@@ -240,15 +240,18 @@ class DefaultMacros(val c: whitebox.Context) extends CaseClassMacros {
         alt => alt.isMethod && hasDefaultParams(alt.asMethod)
       }
 
-    def defaultsFor(fields: List[(TermName, Type)]) = for {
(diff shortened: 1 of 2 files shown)


milessabin (approved):
LGTM!

PR #1240, opened 2022-02-11
Remove refinement from Witness.value

joroKr21:


milessabin:
Does this still preserve the singleton type?

joroKr21:
> Does this still preserve the singleton type?

Yes - we have tests for that. It's the type on the RHS of `T` that is the singleton. So this trick is not necessary.

joroKr21:
How about we just release it and see if someone complains? It's binary compatible either way so easy to revert.

milessabin:
> How about we just release it and see if someone complains? It's binary compatible either way so easy to revert.

Go for it ...

joroKr21:
> Go for it ...

I can't merge it without approval 😅

milessabin:
> I can't merge it without approval :sweat_smile:

Whoops! Approved now ...

milessabin (approved):
LGTM!

PR #1242, opened 2022-02-11
Remove refinement from Witness.value

joroKr21:
Backports #1240


milessabin:
Same question as on #1240 ... does this preserve the singleton type?

Issue #1248, opened 2022-03-19
Possible derivation regression since shapeless 2.3.8 


pomadchin:
The shapeless 2.3.8 dependency update caused frameless test compilation failure, you can find the PR and the CI failure here: https://github.com/typelevel/frameless/pull/612

I decided to go investigate it a little and to prepare a scoped example of the code that causes compilation errors, it looks like it may be related to the `type alias` resolution and `Generic` derivation:

#### Code example, function fooPrintT compiles with Shapeless <= 2.3.7 but not with 2.3.8

```scala
import shapeless.{Generic, HList, HNil, ::}

trait TC[A] { def foo: A }
object TC {
  def apply[A](implicit ev: TC[A]): TC[A] = ev

  implicit val booleanTC = new TC[Boolean] { def foo: Boolean = true }
  implicit val intTC = new TC[Int] { def foo: Int = 42 }
  implicit val longTC = new TC[Long] { def foo: Long = 420L }

  implicit def genericTC[T, L <: HList](implicit gen: Generic.Aux[T, L], tc: TC[L]): TC[T] = new TC[T] { def foo: T = gen.from(tc.foo) }
  implicit def tcHNil: TC[HNil] = new TC[HNil] { def foo: HNil = HNil }
  implicit def tcHCons[H, T <: HList](implicit dh: TC[H], dt: TC[T]): TC[H :: T] = new TC[H :: T] { def foo: H :: T = dh.foo :: dt.foo }
}

case class TX2[A, B](a: A, b: B)
case class TX3[A, B, C](a: A, b: B, c: C)

def fooPrint[A: TC, B: TC](data: TX2[A, B]): Unit = println(TC[TX2[A, B]].foo)

fooPrint(TX2[Int, Boolean](1, true))

type TTX3[A, B] = TX3[Long, Int, TX2[A, B]]

// compiles with shapeless <= 2.3.7
def fooPrintT[A: TC, B: TC](data: TTX3[A, B]): Unit = println(TC[TTX3[A, B]].foo)

fooPrintT(TX3(1L, 1, TX2(1, true)))
```

#### build.sbt

```scala
libraryDependencies += "com.chuusai" %% "shapeless" % "2.3.8"

scalaVersion := "2.13.8"
```

#### Compilation error

```scala
could not find implicit value for parameter ev: TC[TTX3[A,B]]
[...]

PR #1249, opened 2022-03-20
Fix Generic materialization for type aliases (series/2.3)

pomadchin:
This PR resolves https://github.com/milessabin/shapeless/issues/1248 with the help of this comment https://github.com/milessabin/shapeless/issues/1248#issuecomment-1073112655

PR #1250, opened 2022-03-20
Fix Generic materialization for type aliases

pomadchin:
The same as https://github.com/milessabin/shapeless/pull/1249 but against the main branch.

Closes https://github.com/milessabin/shapeless/issues/1248


joroKr21:
The issue is here: https://github.com/milessabin/shapeless/blob/main/core/src/main/scala/shapeless/generic.scala#L880 - `tpe.typeArgs` is different for the type alias and doesn't match `apply` - this fix does the job. If we remove the `typeArgs` another test fails. Unfortunately that also means that custom `apply` methods with a different number of type arguments won't work. I will think a bit more about it but this fix is probably the best we can do.

joroKr21 (approved):
I can't think of a better solution. Thanks for the fix 👍

PR #1255, opened 2022-03-25
When case class has custom unapply, reference getters instead

### LINKED ISSUE #1059: Could not find Generic instance if object has overload "unapply" method since Scala 2.13.4
## A.scala

```scala
package example

case class A(x: Int, y: String)

object A {
  def unapply(a: A): Option[(Int, String, String)] = ???
}

object Main {
  shapeless.Generic[A]
}
```

## build.sbt

```scala
libraryDependencies += "com.chuusai" %% "shapeless" % "2.3.3"

scalaVersion := "2.13.4"
```

## log

```
A.scala:10:20: could not find implicit value for parameter gen: shapeless.Generic[example.A]
[error]   shapeless.Generic[A]
[error]                    ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

joroKr21:
Fixes #1059


milessabin (approved):
LGTM!

PR #1286, opened 2022-11-13
Add support for `Generic` materialized in macro-generated companion object of nested case class

DmytroMitin:
Possibly fixes https://github.com/milessabin/shapeless/issues/1287


joroKr21:
Nice thanks. At some point I think I tried to make the constructor default for case classes but it didn't work. I don't remember why exactly. You could try and see which tests fail. I also wonder if we should provide the generic macro annotation as functionality.

joroKr21 (approved):

PR #1300, opened 2023-02-13
remedy build failure with Scala 2.13.10

fordm:


joroKr21:
Hey @fordm, thank you for contributing!

Could you say a few words about the issue you're having with Scala 2.13.10? I'm a bit confused by this change.

fordm:
> Hey @fordm, thank you for contributing!
> 
> Could you say a few words about the issue you're having with Scala 2.13.10? I'm a bit confused by this change.



> Hey @fordm, thank you for contributing!
> 
> Could you say a few words about the issue you're having with Scala 2.13.10? I'm a bit confused by this change.



> Hey @fordm, thank you for contributing!
> 
> Could you say a few words about the issue you're having with Scala 2.13.10? I'm a bit confused by this change.

Hi.  This issue was this failing build: https://github.com/milessabin/shapeless/pull/1281  So the main branch is stuck at Scala 2.13.7 by the looks of it.  As Scala 2.13.7 is affected by https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-36944 it would be good to get the pipeline running again so Shapeless 2.13 can used in places where this may be an issue.

joroKr21:
Oh right, I completely forgot to look into it. Thanks 👍

joroKr21:
Oh hmm I forgot but it needs an update on the required checks.
@milessabin I don't have access rights to do that.

milessabin:
@joroKr21 I've updated the rules to move to 2.13.10.

I don't see anywhere to give you any more permissions than you already have, I'm afraid. Do we need to create a team to support that?

joroKr21:
> I don't see anywhere to give you any more permissions than you already have, I'm afraid. Do we need to create a team to support that?

I have no idea. I'm missing the "Settings" menu like in other repos where I have admin access.

Screenshot 2023-02-17 at 10 03 32

joroKr21 (approved):

PR #1303, opened 2023-03-25
Update `try-shapeless.sh` script

xuwei-k:


joroKr21 (approved):
Thanks a lot, looks like we neglected this script for a long time.

PR #1324, opened 2023-09-06
Concrete return type for `MkFieldLens.mkFieldLens`

### LINKED ISSUE #1323: MkFieldLens.mkFieldLens returns a derived type instead of type parameter
`MkFieldLens.mkFieldLens[A, N <: Nat, R <: HList, B]` returns an `Aux[A, K, mkLens.Elem]` instead of using `MkRecordSelectLens.Aux[R, K, B]` and return a `Aux[A, K, B]`

While this compiles just fine, it makes the job harder for static code analysis and intellij specifically is getting confused by it

liorregev:
Fixes #1323


joroKr21 (approved):
Thanks! Will have to backport this to `series/2.3` branch to actually release

PR #1373, opened 2024-09-22
Use primary constructor to extract field annotations

### LINKED ISSUE #1372: Auxiliary constructors with configured codecs
When using circe to decode a class, having an auxiliary constructor produces an error.  The circe folks traced this to a shapeless error: https://github.com/circe/circe/issues/1129.

scala> case class Qux()
defined class Qux

scala> case class Foo(i: Int)
defined class Foo

scala> case class Bar(i: Int) { def this() = this(0) }
defined class Bar

scala> shapeless.Annotations[Qux, Foo]
res0: shapeless.Annotations[Qux,Foo]{type Out = None.type :: shapeless.HNil} = shapeless.Annotations$$anon$1@577d35bc

scala> shapeless.Annotations[Qux, Bar]
<console>:16: error: could not find implicit value for parameter annotations: shapeless.Annotations[Qux,Bar]
       shapeless.Annotations[Qux, Bar]
                            ^

joroKr21:
Fixes #1372

### DIFF
--- a/core/shared/src/main/scala/shapeless/annotation.scala
+++ b/core/shared/src/main/scala/shapeless/annotation.scala
@@ -378,11 +378,10 @@ class AnnotationMacros(val c: whitebox.Context) extends CaseClassMacros {
 
   def getAnnotationTreeOptions(tpe: Type, typeAnnotation: Boolean): List[List[(Type, Tree)]] = {
     if (isProduct(tpe)) {
-      val constructorSyms = tpe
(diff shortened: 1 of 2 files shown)


milessabin (approved):
LGTM!

joroKr21:
@milessabin publishing is broken again 🤔 
https://github.com/milessabin/shapeless/actions/runs/10981901535/job/30489610732

PR #1375, opened 2024-09-25
Generate `Automatic-Module-Name` attribute

hubertp:
As it stands, Java 9 Platform Module System cannot infer module name from shapeless' jars due to cross compilation (underscores in `_2.xx`).
This simple change adds an `Automatic-Module-Name` attribute to MANIFEST, making it possible to require `shapeless` modules, rather than going through various repackaaging hacks.
A proper modularization would include definition of `module-info.java` but that's outside the scope.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,6 +91,8 @@ lazy val commonSettings = crossVersionSharedSources ++ Seq(
   )
 )
 
+lazy val javaModuleName = settingKey[String]("Java module name")
+
 def configureJUnit(crossProject: CrossProject) = crossProject
   .jvmSettings(libraryDependencies += "com.github.sbt" % "junit-interface" % "0.13.3" % "test")
   .jsConfigure(_.enablePlugins(ScalaJSJUnitPlugin))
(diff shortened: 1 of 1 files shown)


hubertp:
Any problem with this PR? This should unblock anyone who wants to use `shapeless` as a Java module, with a minimal effort.

build.sbt
```diff
@@ -134,7 +137,10 @@ lazy val coreTestMacros = crossProject(JSPlatform, JVMPlatform, NativePlatform)
 lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .crossType(CrossType.Full)
   .configureCross(configureJUnit)
-  .settings(moduleName := "shapeless")
+  .settings(
+    moduleName := "shapeless",
+    javaModuleName := "shapeless.core"
```
joroKr21, line 142: Perhaps we should use just `shapeless` for this one? Shapeless 2 is not really modular, the plugin and scratch are not published artifacts.

joroKr21 (approved):

PR #1390, opened 2025-01-09
Fix sourcemap path for ScalaJs

kalejami:
Sourcemaps for ScalaJs do not work because they refer to local files. This PR fixes the issue by using "-P:scalajs:mapSourceURI" to point the path to the source files on github

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,6 +91,18 @@ lazy val commonSettings = crossVersionSharedSources ++ Seq(
   )
 )
 
+lazy val jsSourceMapSettings = Def.settings(
+  scalacOptions += {
+    val tagOrHash =
+      if (!isSnapshot.value) s"v${version.value}"
+      else git.gitHeadCommit.value.getOrElse("main")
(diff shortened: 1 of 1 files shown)


joroKr21 (approved):
For reference, `sbt-typelevel` does the same:
https://github.com/typelevel/sbt-typelevel/blob/main/github/src/main/scala/org/typelevel/sbt/TypelevelScalaJSGitHubPlugin.scala

kalejami:
So, any chance getting that released any time soon? Would greatly reduce warning logs with stuff like vite :)

Thanks in advance.

joroKr21:
It looks like releases are currently broken. CC @milessabin 
https://github.com/milessabin/shapeless/actions/runs/12963426396/job/36161090572

milessabin:
@joroKr21 this should be fixed now ... it was stale Sonatype credentials.

joroKr21:
We also need the `setup-sbt` action, I guess should update `sbt-github-actions` to the latest

PR #1397, opened 2025-02-27
Bumped Scala and SBT versions

milessabin:
Supercedes #1389 and #1392.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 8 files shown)


som-snytt:
They should make `supersedes` a hard keyword. It's not a hobby horse, it's just that some word-of-the-day said why it's spelled that way, I never went back. It's not like defense vs defence, Heaven forfend.

Actually I take it back. `m-w` leaves it in the [dictionary](https://www.merriam-webster.com/dictionary/supercede), but I think they should call it the [Scots spelling](https://www.merriam-webster.com/dictionary/supersede). The drinking game would be, spell it with a `c`, you have a Scotch.

PR #1399, opened 2025-02-28
fix plugin project setting

xuwei-k:
this is a compiler-plugin project. not sbt-plugin


joroKr21:
> I can't help wondering how this ever worked 🤪

It's unpublished and we only use it inside shapeless itself

